PR #109, opened 2013-03-01
Add provisions to initialize clock tree

aspalmer:
This commit adds functionality that would allow user to
define/initialize the clock tree configuration as a structure, which
allows for simpler and much more flexible clock initialization.


esden:
I think we already have RCC setup functions. So this PR seems to be superseded? Do I see it right?

Marking as stale

PR #194, opened 2013-08-18
Fix handling of control transfers where size is a multiple of bMaxPacket...

nagasakihacky:
...Size0

End of transfer is indicated by a "short" packet where packet size is less than MaxPacketSize. If the transfer has a length that is a multiple of MaxPacketSize, a zero length packet must be sent to indicate the end of the transfer.

By calling the "normal transmission" section rather than the "end of transmission" section for the last packet of data, the state is not prematurely set to LAST_DATA_IN, so usb_control_send_chunk() gets called again and sends a ZLP

### DIFF
--- a/lib/usb/usb_control.c
+++ b/lib/usb/usb_control.c
@@ -73,7 +73,7 @@ int usbd_register_control_callback(usbd_device *usbd_dev, uint8_t type,
 
 static void usb_control_send_chunk(usbd_device *usbd_dev)
 {
-	if (usbd_dev->desc->bMaxPacketSize0 <
+	if (usbd_dev->desc->bMaxPacketSize0 <=
 			usbd_dev->control_state.ctrl_len) {
 		/* Data stage, normal transmission */
 		usbd_ep_write_packet(usbd_dev, 0,


kuldeepdhaka:
case 1: device return the equal amount of data that the host is expecting.
 sub case 1: the length of data is multiple of bMaxEndpointSize0
        you want to send a ZLP in the end 
        (afaif host right after getting the expected data, send a STATUS-OUT)
 sub case 2: the length of data is not multiple of bMaxEndpointSize0
       no need to send ZLP

case 2: device return less data than the host is expecting
 sub case 1: the length of data is multiple of bMaxEndpointSize0
       you want to send a ZLP in the end 
      (afaik need to send ZLP to tell host that i have no more data)
 sub case 2: the length of data is not multiple of bMaxEndpointSize0
      no need to send a ZLP
 sub case 3: wLength > 0, but device return 0 bytes, 
      (afaik ZLP need to be send)

case 3: wLength = 0, no data stage

kuldeepdhaka:
current code:
1.1. since their is no ZLP sending anywhere, doing in general way. no problem
1.2. no problem
2.1. [https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_control.c#L156] problem!
2.2. no problem
2.3 [https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_control.c#L156] problem!

you changes:
1.1. [https://github.com/nagasakihacky/libopencm3/blob/patch-1/lib/usb/usb_control.c#L76] problem! because your changes force to send a ZLP (ie stack is expecting to send a ZLP, host send a STATUS-OUT boom!) [**you broke this**]
1.2. no problem
2.1. [https://github.com/nagasakihacky/libopencm3/blob/patch-1/lib/usb/usb_control.c#L156] since your change force to send a ZLP, **you are solving this**
2.2. no problem
2.3 [https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_control.c#L156] problem!

disclaimer: just my analysis. i could be completly wrong!

karlp:
If someone wants to "fix" this, it should be pretty easy to provide test device firmware and some test host code.  @nagasakihacky do you have that available at all?

nagasakihacky:
After finding this and a few other quite fundamental bugs along the way I ended up reverting to ST's libraries. At some point I may see if I can switch back to this library, but I didn't have the luxury of time and so had to abandon

karlp:
no problem, thanks for the report at least.

kuldeepdhaka:
please try out https://github.com/kuldeepdhaka/libopencm3-examples/tree/usb-stack-tester
it contain a example to test the above analysis on usb stack

kuldeepdhaka:
as per my finding, this commit will break the usb stack.

karlp:
@kuldeepdhaka So, this commit as is is no good, but the issue it attempts to fix is valid, is that correct?  I was hoping to look at your stack tester control transfer sizing tests, that's the problem here right?

kuldeepdhaka:
@karlp yes, the issue is valid but commit has problems.

karlp:
fixed via merge of 3ed12b6fd944c4b3c921b8f73a6912f856edb68a

PR #262, opened 2014-02-07
Update the modules for the examples tree

BuFran:
Update the modules for compatible build of the examples tree.


flixr:
@BuFran any updates here?

BuFran:
I'm back, so I'll strip out the size report outside this PR. I had problems with newest gcc-arm toolchains (the style of output was changed, values are different and not corresponds to real FLASH/RAM usage, so some investigation is needed)

karlp:
examples have been updated multiiple times since this was last filed. not sure why this is still here.

PR #274, opened 2014-02-14
Usb works

UweBonnes:
For the STM32F103, the SOF interrupt is not enabled at all, so any user SOF callback function is only called when another interrupt is triggered. However having the SOF interrupt enabled without any SOF user callback function given results in a 1Kz of useless interrupts.  The second patch of this series disables SOF interrupts as long as no user SOF callback function is given, at the expense of some commands more in the interrupt routine.

Please discuss.

The first patch is cosmetical...


karlp:
Closed via the inclusion of a tweaked version of #388

PR #282, opened 2014-03-03
STM32F072 USB support in clean way

BuFran:
Grabbed #278 and #265 from kuldeepdhaka's great pioneer work, code rewritten to be compatible with current coding style and inclusion tree of libopencm3.

Timers tested, USB needs some more testing as one definition changed (USB_DADDR_ENABLE -> USB_DADDR_EF according to the datasheet name. )

The DAC header fix from #278 is bad way, it has been discussed in the past on irc. It needs another solution (more globally stable).


karlp:
Phew, f0 usb support, including code found here has finally landed in master! thanks for the tolerance!

PR #284, opened 2014-03-04
[stm32][usart] API functions to get various status flags

flixr:
- enables usage of same code on F1,2,3,4

alternative to #283


esden:
OK as it seems the consensus is that we don't need those functions any more I am closing this pull request. :smile:

PR #285, opened 2014-03-04
[stm32] adc_read_injected returns signed int

flixr:
Since reading an injected channel can apparently return negative values, the return type should probably be a signed int...


jsphuebner:
the code in #967 is tested

PR #286, opened 2014-03-06
[stm32] start implementing timer_get_frequency

flixr:
Get the timer clock frequency in order to more conveniently calculate the prescaler.
- untested and only valid for F1 and STM32F405xx/07xx and STM32F415xx/17xx
- does not handle/check TIMPRE in RCC_CFGR found in F42xxx and F43xxx

Ran out of time, so in case this sounds useful and someone wants to test it and/or extend it for more chips, here we go....


esden:
There is probably still a good reason to add this if someone wants to pick this up. But considering how old this is I will mark it as stale until someone feels up for picking this up again.

Stale means I will close this in a month unless someone picks this up.

PR #305, opened 2014-04-18
Add provisions to support multiple altsettings

kuldeepdhaka:
This commit implements the support for one interface to have multiple
altsettings. It also adds hook that user can use to perform actions
when the alsetting switch is performed by host.

Changes:
* For backward compatibility, placed a pointer instead of allocating memory for whole interface struct.
* Always execute callback (even if the current interface alternate-settings matches).
* Multiple configuration support.

Signed-off-by: Kuldeep Singh Dhaka <kuldeepdhaka9@gmail.com>

Conflicts:
	lib/usb/usb_standard.c

### DIFF
--- a/include/libopencm3/usb/usbstd.h
+++ b/include/libopencm3/usb/usbstd.h
@@ -163,7 +163,8 @@ struct usb_config_descriptor {
 
 	/* Descriptor ends here.  The following are used internally: */
 	const struct usb_interface {
-		int num_altsetting;
+		uint8_t *cur_altsetting;
+		uint8_t num_altsetting;
 		const struct usb_iface_assoc_descriptor *iface_assoc;
 		const struct usb_interface_descriptor *altsetting;
 	} *interface;
--- a/lib/usb/usb_standard.c
+++ b/lib/usb/usb_standard.c
@@ -46,6 +46,13 @@ void usbd_register_set_config_callback(usbd_device *usbd_dev,
 	usbd_dev->user_callback_set_config = callback;
 }
 
+void usbd_register_set_altsetting_callback(usbd_device *usbd_dev,
+					   void (*callback)(usbd_device *usbd_dev,
+							    uint16_t wIndex, uint16_t wValue))
+{
+	usbd_dev->user_callback_set_altsetting = callback;
+}
+
 static uint16_t build_config_descriptor(usbd_device *usbd_dev,
 				   uint8_t index, uint8_t *buf, uint16_t len)
 {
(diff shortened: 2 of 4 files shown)


include/libopencm3/usb/usbstd.h
```diff
@@ -163,7 +163,8 @@ struct usb_config_descriptor {
 
 	/* Descriptor ends here.  The following are used internally: */
 	const struct usb_interface {
-		int num_altsetting;
+		uint8_t * cur_altsetting;
+		uint8_t num_altsetting;
```
karlp: are these uint8_t to save ram? or are they actually limited in some way?  just thinking out loud, but ```int_least8_t``` might be enough here?

karlp:
I'd like to see more {} around if/else cases, but otherwise I guess this looks good and well overdue :)

include/libopencm3/usb/usbstd.h
kuldeepdhaka: as per usb specs, bInterfaceNumber ,bAlternateSetting, bNumInterfaces are 8bit wide and unsigned (as seen in descriptor).
any benifit of increasing the width (or change sign)? afaif, No.
also, added benefit ram saving :p

kuldeepdhaka:
@karlp added {} around if/else cases.

lib/usb/usb_standard.c
```diff
@@ -268,32 +275,53 @@ static int usb_standard_set_interface(usbd_device *usbd_dev,
...
+		return USBD_REQ_NOTSUPP;
 	}
+
+	iface = &cfx->interface[req->wIndex];
+	
+	if (req->wValue >= iface->num_altsetting) {
+		return USBD_REQ_NOTSUPP;
+	}
+
+	if(iface->cur_altsetting) {
```
karlp: space after the if and before the ( here please (same below in a couple of places)

kuldeepdhaka:
@karlp if , ( space done.

karlp:
Merged via rebase, after some minor whitespace cleanup.  thanks for the patience!

superjyce:
On Windows, USB enumeration requests bDescriptorType:0x21 that this patch does not add.
So, here my patch :
`diff --git a/lib/usb/usb_standard.c b/lib/usb/usb_standard.c
index 885f077..284107c 100644
--- a/lib/usb/usb_standard.c
+++ b/lib/usb/usb_standard.c
@@ -157,6 +165,11 @@ static int usb_standard_get_descriptor(usbd_device *usbd_dev,
                *buf = usbd_dev->ctrl_buf;
                *len = build_config_descriptor(usbd_dev, descr_idx, *buf, *len);
                return USBD_REQ_HANDLED;
+       case 0x21: // added to support functional descriptors
+               cur_altsetting = *(usbd_dev->config->interface[descr_idx].cur_altsetting);
+               *buf = (void *)usbd_dev->config->interface[descr_idx].altsetting[cur_altsetting].extra;
+               *len = usbd_dev->config->interface[descr_idx].altsetting[cur_altsetting].extralen;
+               return USBD_REQ_HANDLED;
        case USB_DT_STRING:
                sd = (struct usb_string_descriptor *)usbd_dev->ctrl_buf;
`

kuldeepdhaka:
not related to this commit, should be handled in application code.
see usbd_register_control_callback

PR #315, opened 2014-05-18
generic spi_init_master cleanup. 

rewolff:
The "spi_init_master" function required me, the user, to know that certain parameters were intended for certain control registers (mostly CR1). It's fine to directly use the function parameter in the implementation, but the coder using the function does not want to know what registers these things go into. So some additional defines were introduced that are control-register agnostic. 

Apparently my previous patch regarding some flash programming comments is also included. And maybe a DAC comment as well. I'm not good enough with GIT yet to separate those. Sorry. Feel free to educate me, especially if you want to keep them as separate commits in the master branch or something like that.

### DIFF
--- a/include/libopencm3/cm3/common.h
+++ b/include/libopencm3/cm3/common.h
@@ -54,10 +54,10 @@
 
 /* Generic bit-band I/O accessor functions */
 #define BBIO_SRAM(addr, bit) \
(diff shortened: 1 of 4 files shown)


esden:
I am marking this as stale, unless someone decides to pick this up again I will close this pr in a month.

PR #322, opened 2014-05-29
Prevent inadvertently clearing USB_EP_RX_CTR and USB_EP_TX_CTR

gmccollister:
USB_EP_RX_CTR and USB_EP_TX_CTR must be set when USB_EP_REG(n) is modified to prevent inadvertently clearing either bit. Writing a 1 for these bits causes no change to the register.

USB_EP_RX_CTR and USB_EP_TX_CTR should probably be removed from USB_EP_NTOGGLE_MSK but care will need to be taken to also OR in USB_EP_RX_CTR and USB_EP_TX_CTR anywhere the register is written.


karlp:
merged via rebase

PR #323, opened 2014-05-30
cm3/cortex.h: specify headers for non-standard types explicitly

joe-skb7:
If user includes cortex.h header without previously included stdbool.h
and stdint.h a build errors like this occur:

include/libopencm3/cm3/cortex.h:84:1: error: unknown type name 'bool'
 static inline bool cm_is_masked_interrupts(void)
 ^

Signed-off-by: Sam Protsenko <joe.skb7@gmail.com>


karlp:
The original problems with the includes has been fixed.  I've filed https://github.com/libopencm3/libopencm3/issues/475 to track the remaining quirks with bool vs uint32_t in the code itself.

PR #351, opened 2014-10-05
stm32: SPI_I2S common definitions (LSB)

amirhammad:
Definitions for I2S in _include/libopencm3/stm32/common/spi_common_all.h_ have the following style:

> #define SPI_I2SCFGR_I2SCFG_LSB 8
> #define SPI_I2SCFGR_I2SCFG_SLAVE_TRANSMIT	0x0
> #define SPI_I2SCFGR_I2SCFG_SLAVE_RECEIVE	0x1
> #define SPI_I2SCFGR_I2SCFG_MASTER_TRANSMIT	0x2
> #define SPI_I2SCFGR_I2SCFG_MASTER_RECEIVE	0x3

So when one want to use one of those defines, he must also shift this definition left by the value of SPI_I2SCFGR_*_LSB.

I see workaround by including this LSB into bitmask definitions

> #define SPI_I2SCFGR_I2SCFG_SLAVE_TRANSMIT	(0x0 << 8)
> #define SPI_I2SCFGR_I2SCFG_SLAVE_RECEIVE	(0x1 << 8)
> #define SPI_I2SCFGR_I2SCFG_MASTER_TRANSMIT	(0x2 << 8)
> #define SPI_I2SCFGR_I2SCFG_MASTER_RECEIVE	(0x3 << 8)

I don't see any reason to have *_LSB separated.

### DIFF
--- a/include/libopencm3/stm32/common/spi_common_all.h
+++ b/include/libopencm3/stm32/common/spi_common_all.h
@@ -295,11 +295,10 @@ specific memorymap.h header before including this header file.*/
 #define SPI_I2SCFGR_I2SE			(1 << 10)
 
 /* I2SCFG[9:8]: I2S configuration mode */
-#define SPI_I2SCFGR_I2SCFG_LSB			8
-#define SPI_I2SCFGR_I2SCFG_SLAVE_TRANSMIT	0x0
-#define SPI_I2SCFGR_I2SCFG_SLAVE_RECEIVE	0x1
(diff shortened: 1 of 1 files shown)


karlp:
This does appear to make I2S more like SPI for this case.  But the library in general has a bit of a mix.  If we're going to do this, then we need to _declare_ a final, once and for all style for this, write it in HACKING.txt and make _all_ the code handle this.  So while this seems somewhat valid, I'm going to close this for now as incomplete.

PR #353, opened 2014-10-14
Doxygen: Generic _commmon_ EXCLUDE_PATTERNS for stm32 family

ejoerns:
When I scrolled through the Doxygen docs, I discovered that some STM32 platforms reference files and functions that do not belong to them.

E.g. STM32F1 GPIO module includes `gpio_common_f0234.c`, STM32L1 DMA module includes `dma_common_f24.c`.

Obvioiusly, this is caused by insufficient exclude patterns in the corresponding `Doxyfile`s. This PR should fix that for now and for the future. :)

---

**Commit message**

Previously, most exclude patterns were written to match exactly one type
of common file, e.g. _common_f02.h files.
The drawback of this method is that newly added files are not covered
by the existing exclude pattern and thus pollute the generated
documentation of controllers they do not belong to.

These more generic exclude patterns instead handle all kind of possible
combinations of existing stm32 platforms (L1,F[0-4]) (that follow the
current naming convention).

### DIFF
--- a/doc/stm32f0/Doxyfile
+++ b/doc/stm32f0/Doxyfile
@@ -25,9 +25,10 @@ INPUT                  += ../../lib/stm32/f0 \
 EXCLUDE                 = ../../include/libopencm3/stm32/f0/usb.h \
                          ../../include/libopencm3/stm32/f0/usb_desc.h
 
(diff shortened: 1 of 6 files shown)


karlp:
I actually finally killed this issue by making the doxygen source lists be automatically generated from the build files. it should be "always" perfect and self contained now.

PR #359, opened 2014-11-10
Arm m0 rcc addition

gtoonstra:
I'm doing a project with an stm32 M0 chip (STM32F042C6), which runs on 48MHz max. This chip also has a CAN bus. Some defines had to be added to make the build work. I'm not 100% sure about the "rcc_clock_setup_in_hse_8mhz_out_48mhz" function. It was copied and adapted from the F1 series, but the device seems to run ok with it. 

The datasheet for this M0 chip can be found here: http://www.st.com/web/en/catalog/mmc/FM141/SC1169/SS1574/LN1823/PF259561


karlp:
Everything in his PR has landed via others at this point, sorry we couldn't get this further at the time.

PR #375, opened 2014-12-05
[stm32f/usb] fix for minimum fifo size=16

H2OBrain:
fix for fifo size minimum (INEPTXFD,..)

### DIFF
--- a/lib/usb/usb_fx07_common.c
+++ b/lib/usb/usb_fx07_common.c
@@ -46,8 +46,17 @@ void stm32fx07_ep_setup(usbd_device *usbd_dev, uint8_t addr, uint8_t type,
 	 * Configure endpoint address and type. Allocate FIFO memory for
 	 * endpoint. Install callback funciton.
 	 */
-	uint8_t dir = addr & 0x80;
+	uint16_t fifo_word_len;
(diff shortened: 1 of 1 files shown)


karlp:
instead of fixing the fix, can you squish this into a single commit please?

Further, can you properly describe what the problem is here, I don't follow it immediately.

H2OBrain:
Hi Karl

This fix changes makes sure, that the minimum tx-fifo depth configured is 16 (words) (see reference manual DM00031020). I'm not sure if this is correct for the other supported chips though.
Also the "real fix" between c411a0f and [here](https://github.com/H2OBrain/libopencm3/commit/23a5a0fe355f925e58711dcea8c09310bffe80f8) "enables" the usb fifo, so it's possible to transmit more than 1 packet at a time (which improves USB transmission rate ~10x to 8mbit/s, near the USB2.0 full-speed limit).
There are other fixes to be made in libopencm3 like cable disconnect detection, fifo-flushing, configuration changes.. I fixed those in my simplified "codebase" as a hack.
I still want to make a working usb-example, but I still have no time to do it.

You can close the pull request, this fix isn't very important. Without it max_size should always be 64.

Cheers,
Oli

PR #376, opened 2014-12-05
[stm32f4-ltdc] initial commit

H2OBrain:
Can you tell me what i should do with all those helper macros, please?


esden:
This is a good point @kbob I think using similar system to the F4 clock system is the way to go. I have merged this pull request into master as is. Because it is better to have in there than nothing. We should continue polishing and improving the convenience functions for this through additional pull requests.

This pull request was merged through 7fbd49b3fc68e43097b653fddc3a39f176d62f58 so I am closing it.

PR #377, opened 2014-12-05
expanding on ethernet

brabo:
expanded ethernet mac/phy, more configurability. replaced ethernet.o by the ethernet/ mac/phy files in f1/Makefile.


karlp:
The comments on this PR make me suspect that there were objections to it at the time?  IS this still stuff that should be added?  I'll try and rebase it if it's still important, but if anyone has other opinions please let me know, otherwise I'll close it in a ~month or so.

esden:
As Karl essentially marked it as stale a while ago I will close this PR. It would be nice to get Ethernet support though but alas.

PR #379, opened 2014-12-07
First cut at an HSI configuration API for rcc on the STMF4 series.

ChuckM:
This is a first cut at a simplified setup function for the HSI clock on the STM32F series. The function is called simply `rcc_set_hsi_clock(uint32_t speed)` and it takes 32 bit constant which has encoded the PLL values to achieve that speed.  It is possible to generalize it to something even more generic, one could have `rcc_set_pll_clock(uint32_t pll_constants, uint32_t base_frequency);` where the  `pll constants` would have been set for an HSE or HSI input, and `base_frequency` would be the starting crystal frequency. But that is prone to error is you used HSI PLL flags and an HSE crystal for example. 

Two things I do like is that it automatically computes the best APB2 and APB1 clocks. This avoids problems where the constants don't match the setup and things like the UARTs don't get the right baud rate constants. It also automatically sets the FLASH wait states based on the data sheet. I did consider making the FLASH wait state value a parameter so that it is completely general purpose at all voltages for HSI.


include/libopencm3/stm32/f0/rcc.h
```diff
@@ -374,6 +386,36 @@ Control</b>
...
+ *
+ * This macro packs away bits that are used when configuring the PLL. The
+ * "gotcha" is that PREDIV[3:0] are in CFGR2 rather than CFGR so we pack
+ * those in the bottom 4 bits, knowing we will strip them off and use them
+ * in CFGR2.
+ */
+#define RCC_PLLCFGR_BITS(mul, src, prediv) (\
+		((mul & RCC_CFGR_PLLMUL_MASK) << RCC_CFGR_PLLMUL_SHIFT) |\
+		((src & RCC_CFGR_PLLSRC_MASK) << RCC_CFGR_PLLSRC_SHIFT) |\
+		((src & RCC_CFGR2_PREDIV_MASK) << RCC_CFGR2_PREDIV_SHIFT))
```
ChuckM: They are used in the `rcc_pll_clock_setup` in `lib/stm32/f0/rcc.c` because the PLL has two configuration registers in that family. But the API stays the same `rcc_pll_clock_setup(configuration_bits, base_frequency);`

include/libopencm3/stm32/l0/rcc.h
```diff
@@ -457,6 +477,15 @@
...
 
+/* --- Helper Defines ----------------------------------------------------- */
+/* MSI Base frequency is 4194304 Hz can be divided by 2, 4, 8, 16, 32, 64 */
+#define RCC_MSI_4MHZ				((0x6 << 24) | 4194304)
+#define RCC_MSI_2MHZ				((0x5 << 24) | 2097152)
+#define RCC_MSI_1MHZ				((0x4 << 24) | 1048576)
+#define RCC_MSI_524KHZ				((0x3 << 24) | 524288)
+#define RCC_MSI_262KHZ				((0x2 << 24) | 262144)
+#define RCC_MSI_131KHZ				((0x1 << 24) | 131072)
+#define RCC_MSI_65KHZ				((0x0 << 24) | 65536)
```
ChuckM: No, just the available frequencies from the data sheet, I store the config bits in the upper 8 bits of the `#define` and the actual frequency in the lower 24 bits. This lets it provide both the configuration data and the frequency that will be used (which is unlike the HSI which is fixed frequency).

karlp: heh, ok, as long as you feel it's meaningful to go down to Hz level, when ST says they're just nominal :)

ChuckM:
Here is the latest updated, it adds the 'super EZ clock setup' function which, if we play our cards right could be the only API any beginner/casual user needs to set up the clock domain. It depends on three functions: rcc_hsi_clock_setup(), rcc_hse_clock_setup(), and rcc_pll_clock_setup(). And we add the new function rcc_compute_pll_bits() which implements in code the logic you have to go through to figure out what to put in all the PLL registers for the F4.

lib/stm32/l0/rcc.c
```diff
@@ -290,5 +309,183 @@ void rcc_wait_for_osc_ready(enum rcc_osc osc)
...
+	case 0:
+		return MSI;
+	case 1:
+		return (RCC_CR & RCC_CR_HSI16DIVF) ? HSI4 : HSI16;
+	case 2:
+		return HSE;
+	case 3:
+		return PLL;
+	}
+	return MSI;	/* not reached */
```
ChuckM: Fixed by combining the case 0: case into a default: case which the compiler recognizes will never get past the switch statement.

lib/stm32/f4/rcc.c
```diff
@@ -446,60 +458,78 @@ void rcc_osc_bypass_disable(enum rcc_osc osc)
...
 
 void rcc_set_ppre2(uint32_t ppre2)
 {
-	uint32_t reg32;
-
-	reg32 = RCC_CFGR;
-	reg32 &= ~((1 << 13) | (1 << 14) | (1 << 15));
-	RCC_CFGR = (reg32 | (ppre2 << 13));
+	RCC_CFGR = (RCC_CFGR & (~(RCC_CFGR_PPRE_MASK << RCC_CFGR_PPRE2_SHIFT))) |
+		   ((ppre2 & RCC_CFGR_PPRE_MASK) << RCC_CFGR_PPRE2_SHIFT);
```
ChuckM: Actually I think all of these register setting functions are candidates to be made into static inline functions inside the rcc.h file. Really they are just bit manipulation in a register.

lib/stm32/f4/rcc.c
```diff
@@ -446,60 +458,78 @@ void rcc_osc_bypass_disable(enum rcc_osc osc)
...
 
 void rcc_set_pll_source(uint32_t pllsrc)
 {
-	uint32_t reg32;
-
-	reg32 = RCC_PLLCFGR;
-	reg32 &= ~(1 << 22);
-	RCC_PLLCFGR = (reg32 | (pllsrc << 22));
+	RCC_PLLCFGR = (RCC_PLLCFGR & (~RCC_PLLCFGR_PLLSRC)) |
+			  ((pllsrc & 1) * RCC_PLLCFGR_PLLSRC);
```
ChuckM: Made it look like the others using PLLSRC_MASK and PLLSRC_SHIFT

lib/stm32/f4/rcc.c
```diff
@@ -446,60 +458,78 @@ void rcc_osc_bypass_disable(enum rcc_osc osc)
...
+			   (clk_bits & RCC_CFGR_SW_MASK);
+	/* wait for the switch */
+	while (((RCC_CFGR >> RCC_CFGR_SWS_SHIFT) & RCC_CFGR_SW_MASK) != clk_bits) ;
+}
 
+enum rcc_osc rcc_get_sysclk(void) {
+	uint32_t clk_bits = (RCC_CFGR >> RCC_CFGR_SWS_SHIFT) & RCC_CFGR_SW_MASK;
+	switch (clk_bits) {
+		default:
+			return HSI;
```
ChuckM: it was combined with the case 0: case so it ended up on top. Moved to the bottom.

lib/stm32/f4/rcc.c
```diff
@@ -13,6 +13,18 @@
  * This library supports the Reset and Clock Control System in the STM32 series
  * of ARM Cortex Microcontrollers by ST Microelectronics.
  *
+ * @date 10 Dec 2014
+ *
+ * Added additional helper routine rcc_set_pll_clock and some constants to make
+ * it easy to pick different clock speeds. Its also a bit more space efficient
+ * as clock parameters are #defines rather than static structures.
```
ChuckM: deleted.

ChuckM:
Its been a bit over 5 months on this PR, I've pretty much put in everything you asked for but I get it that it isn't compatible with what you want to do with clocks going forward. I'm not going to bother to rebase it unless you want to make progress here.

PR #382, opened 2014-12-14
Add PHY register descriptions for some of useful phys

BuFran:
This is first cleaned PR to extend ethernet functions to other chips.

All definitions here specified are not tested, only the "generic" part and ksz8051 part.  It is only rewrite from scratch from the datasheet.

Register definitions BCR BSR ID1 ID2 ANTX ANRX ANEXP ANNPTX ANNPRX are standardized and common for all PHYs I learned. Currently they are defined in each PHY family but it can be changed.

The problem in phy register definitions is that they have not standardized names, the names that I made are shorten version of the register name sentences found in datasheet. The names can be changed when some shorts are not self-descriptive.

This PR doesn't touch anything, when users look at them and agreed that it is good to go, it should be in master soon.


karlp:
I tried to merge this, but it contains duplicate definitions of  KSZ80X1_CR1_POLARITY,  KSZ8051_MIICTRL_100MRESTORE, KSZ80X1_CR1_MDIX 

I've fixed the compiler errors from the old PR, but I don't know what should be done abotu the different family type things.  I guess you autogenerated this?  I suggest suffixes for the family variants, but I won't do that myself.  I've pulled the non-kzx*** commtis into staging.

karlp:
merged last pieces in 23fc65d4 and 20b7956d

BuFran:
Thanks, I was long time gone. I am trying now to jump back at horse :-)

PR #386, opened 2014-12-17
stm32: f4: flash: support extended sector ranges

karlp:
F42xx and F43xx have extended sector ranges.

Reported-by: H2OBrain@irc

Alternative version of https://github.com/libopencm3/libopencm3/pull/385

PR #387, opened 2014-12-17
Drafts/flash 24 erase2

karlp:
Second version of the flash program size branch


flixr:
Is this tested, or #386 ?
Looks like the difference is basically only in flash_set_program_size `FLASH_CR |= psize;`

karlp:
closed in favour of #386

PR #389, opened 2014-12-29
stm32: f373: support the 373xc devices

jepler:
These devices have 32kiB RAM and 256kiB flash and are otherwise
register compatible with the f373x8 and f373xc devices.

I have personally tested the generated linkscript on a board with the stm32f373cc.

Signed-off-by: Jeff Epler <jepler@unpythonic.net>


esden:
Thank you! Merged with ef97de242e372efb2f40c39eb2f8a5c9a62c1af3

PR #390, opened 2014-12-31
L0 f0 "usart_get_clock()" routines, for autodetermining baudrates

ChuckM:
This is a work in progress on generating common USART support for the f0/l0 series. Thought I would put it out there if anyone else is looking to work on this part of the system.


include/libopencm3/stm32/common/usart_common_fl_0.h
```diff
@@ -0,0 +1,302 @@
...
+
+/* USART_RTOR Values --------------------------------------------------------*/
+
+#define USART_RTOR_BLEN_SHIFT		24
+#define USART_RTOR_BLEN			(0xFF << USART_RTOR_BLEN_SHIFT)
+#define USART_RTOR_BLEN_VAL(x)		((x) << USART_RTOR_BLEN_SHIFT)
+
+#define USART_RTOR_RTO_SHIFT		0
+#define USART_RTOR_RTO			(0xFF << USART_RTOR_RTO_SHIFT)
+#define USART_RTOR_RTO_VAL(x)		((x) << USART_RTOR_RTO_SHIFT)
```
ChuckM: They were not actually used so I've converted them to the more common 'shift/mask' variant.

include/libopencm3/stm32/common/usart_common_fl_0.h
```diff
@@ -0,0 +1,302 @@
...
+/*****************************************************************************/
+
+BEGIN_DECLS
+
+void usart_set_baudrate(uint32_t usart, uint32_t baud);
+void usart_set_databits(uint32_t usart, uint32_t bits);
+void usart_set_stopbits(uint32_t usart, uint32_t stopbits);
+void usart_set_parity(uint32_t usart, uint32_t parity);
+void usart_set_mode(uint32_t usart, uint32_t mode);
+void usart_set_flow_control(uint32_t usart, uint32_t flowcontrol);
```
ChuckM: That is correct, except that both are defined to be the same thing. The original created what were presumably easy to remember constants which have an identical name/function with the usart_common_all.c code. That way the call to usart_set_flow_control works the same on both the f0/l0 as well as the other chips from a function call level, but it uses a different register in the f0/l0 case. The `#define` is set up as the common constant.

include/libopencm3/stm32/l0/rcc.h
```diff
@@ -610,6 +640,17 @@ void rcc_osc_ready_int_enable(enum rcc_osc osc);
...
+void rcc_set_sysclk(enum rcc_osc clk);
+void rcc_set_ppre1(uint32_t div);
+void rcc_set_ppre2(uint32_t div);
+void rcc_set_hpre(uint32_t div);
+void rcc_hsi4_clock_setup(void);
+void rcc_hsi16_clock_setup(void);
+void rcc_msi_clock_setup(uint32_t msi_freq);
+void rcc_hse_clock_setup(uint32_t hse_freq);
+void rcc_pll_clock_setup(uint32_t pll_freq, uint32_t src_freq);
+
```
ChuckM: The l0 rcc code is incomplete in HEAD so this code is a cherry pick out of the RCC PR to get the necessary function of l0. Once one of these goes to HEAD the other will adapt to use what was decided to go in. Feel free to ignore this code as it will change depending on whether or not the RCC PR is integrated.

lib/stm32/common/usart_common_l0f0.c
```diff
@@ -0,0 +1,460 @@
...
+ *
+ * @brief <b>libopencm3 STM32L0xx USART</b>
+ *
+ * @version 1.0.0
+ *
+ * @date 7 Jul 2013
+ *
+ * LGPL License Terms @ref lgpl_license
+ * TODO: 
+ *  This, and the F0 USART code (from which it is currently
```
ChuckM: noted and updated.

lib/stm32/common/usart_common_l0f0.c
```diff
@@ -0,0 +1,460 @@
...
+ * usart_reg_base
+ * @param[in] bits unsigned 32 bit. Word length in bits 8 or 9.
+ */
+
+void usart_set_databits(uint32_t usart, uint32_t bits)
+{
+	if (bits == 8) {
+		USART_CR1(usart) &= ~USART_CR1_M; /* 8 data bits */
+	} else {
+		USART_CR1(usart) |= USART_CR1_M;  /* 9 data bits */
```
ChuckM: Actually yes, and it should support 7 bit mode (added).

lib/stm32/f0/rcc.c
```diff
@@ -43,6 +43,7 @@
 /* Set the default clock frequencies */
 uint32_t rcc_ahb_frequency = 8000000; /* 8MHz after reset */
 uint32_t rcc_apb1_frequency = 8000000; /* 8MHz after reset */
+uint32_t rcc_lse_frequency = 32768;		/* most common default */
```
ChuckM: I've gone back and forth on that. Do I put it into the lse_enable() ? Does anyone use an LSE that **isn't** 32768Hz ? It is needed if we are computing the baud rate than the baud rate clock is connected to the LSE, but I didn't feel comfortable just _assuming_ it was going to be 32768.

lib/stm32/l0/Makefile
```diff
@@ -34,9 +34,10 @@ CFLAGS		= -Os -g \
 
 ARFLAGS		= rcs
 
-OBJS		= gpio.o rcc.o
+OBJS		= gpio.o rcc.o usart.o
 
-OBJS            += gpio_common_all.o gpio_common_f0234.o rcc_common_all.o
+OBJS            += gpio_common_all.o gpio_common_f0234.o rcc_common_all.o \
+					usart_common_l0f0.o
```
ChuckM: Noted

ChuckM:
Thanks for the comments Karlp, I've updated the files and moved some things. The rcc.c file will be "fixed" when one of these goes to HEAD. Without it, it isn't really buildable.

lib/stm32/f0/rcc.c
karlp: I'd kinda just simply not support automatically setting up baud rates if you've configured the source to be LSE.  It's an edge case I think, and requiring this sort of variable to exist?

esden:
So is this done now? Are we happy with this?

include/libopencm3/stm32/common/usart_common_fl_0.h
```diff
@@ -0,0 +1,296 @@
...
+#define USART_RQR(usart_base)		MMIO32(usart_base + 0x18)
+
+#define USART_ISR(usart_base)		MMIO32(usart_base + 0x1c)
+
+#define USART_ICR(usart_base)		MMIO32(usart_base + 0x20)
+
+#define USART_RDR(usart_base)		MMIO8(usart_base + 0x24)
+
+#define USART_TDR(usart_base)		MMIO8(usart_base + 0x28)
+
```
karlp: Don't really see any need for all this whitespace here myself.

include/libopencm3/stm32/common/usart_common_fl_0.h
```diff
@@ -0,0 +1,296 @@
...
+#define USART_RQR(usart_base)		MMIO32(usart_base + 0x18)
+
+#define USART_ISR(usart_base)		MMIO32(usart_base + 0x1c)
+
+#define USART_ICR(usart_base)		MMIO32(usart_base + 0x20)
+
+#define USART_RDR(usart_base)		MMIO8(usart_base + 0x24)
+
+#define USART_TDR(usart_base)		MMIO8(usart_base + 0x28)
```
karlp: These are wrong.  They have 0..8, (9 bits) not 8.  You'll annoy people doing parity sort of stuff :)

include/libopencm3/stm32/common/usart_common_fl_0.h
```diff
@@ -0,0 +1,296 @@
...
+bool usart_get_interrupt_source(uint32_t usart, uint32_t flag);
+uint32_t usart_get_clock(uint32_t usart);
+void usart_set_autobaudrate(uint32_t usart, int mode);
+
+END_DECLS
+
+#endif
+/** @cond */
+#else
+#warning "usart_common_fl_0.h should not be included directly, only via usart.h"
```
karlp: this should just be an #error, as we did in https://github.com/libopencm3/libopencm3/commit/721f3d74eb20bbbbe7438c98adece47adc7e7a0b

lib/stm32/f0/usart.c
```diff
@@ -31,399 +31,43 @@
...
-	if ((flag >= USART_ISR_IDLE) && (flag <= USART_ISR_TXE)) {
-		return ((flag_set & USART_CR1(usart)) != 0);
-	/* Overrun error */
-	} else if (flag == USART_ISR_ORE) {
-		return flag_set && (USART_CR3(usart) & USART_CR3_CTSIE);
-	}
-
-	return false;
-}
-
```
ChuckM: This "diff" does not give a good representation of what happened to this file. It was all pulled out and put into the common module. The only thing in usart.c now is the get_clock() function which is unique to the F1 vs the L1 or L0.

include/libopencm3/stm32/common/usart_common_fl_0.h
```diff
@@ -0,0 +1,286 @@
...
+#define USART_CR2(usart_base)		MMIO32(usart_base + 0x04)
+#define USART_CR3(usart_base)		MMIO32(usart_base + 0x08)
+#define USART_BRR(usart_base)		MMIO32(usart_base + 0x0c)
+#define USART_GTPR(usart_base)		MMIO32(usart_base + 0x10)
+#define USART_RTOR(usart_base)		MMIO32(usart_base + 0x14)
+#define USART_RQR(usart_base)		MMIO32(usart_base + 0x18)
+#define USART_ISR(usart_base)		MMIO32(usart_base + 0x1c)
+#define USART_ICR(usart_base)		MMIO32(usart_base + 0x20)
+#define USART_RDR(usart_base)		MMIO16(usart_base + 0x24)
+#define USART_TDR(usart_base)		MMIO16(usart_base + 0x28)
```
karlp: is 16 actually right? I know 8 was wrong, but is 16 actually correct?  This doesn't appear to be like the spi peripherals or dma where the size of the operation matters.

include/libopencm3/stm32/common/usart_common_fl_0.h
karlp: I see what you mean more clearly on this one now, it's similar to the stop bits defines here:
https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/stm32/common/usart_common_all.h#L82-L85

karlp:
The outstanding code in this PR is the "usart_get_clock()" routines, which are used for autodetermining how to set the baudrate.  The rest of it has more or less landed via other unification efforts

karlp:
#1191  will provide the same thing, which has become even more required with other families with even more clock source choices for peripherals.  Leaving it open for now, but as that's _likely_ to merge to get f7/h7/mp1/l5/l4/g4 support moving forwards, and it's more "fresh" it will likely obsolete this one. (it also covers more periphs)

esden:
Considering that #1191 landed this is probably outdated and superseded. If I am wrong please feel free to reopen.

Issue #391, opened 2015-01-05
SPI code is broken on STM32F4


ChuckM:
This simple [SPI example code](https://gist.github.com/ChuckM/df360ba3df235f482561) leaves the SPI port wedged with a MODE Fault. You cannot send any data over the SPI port. If I manually set the CR1 bits, check for mode fault, if set reset the CR1 bits. I can get it to work but the library code is broken for this version of the chip.

ChuckM:
Ok, with a **lot** of investigation, this is what I am finding. On the F4, if you want to run in _master_ mode, you must choose one of the following modes:

  * Multi-master - SSOE in CR2 set to `0`, the NSS pin is an "input" and your SPI port can switch from MASTER to SLAVE
    mode based on this pin's state. (which you can control in software if you choose)00
  * Always Master - SSOE in CR2 set to `1`, the SPI port is always the MASTER it can use NSS as an
    output which will change state when it is sending.
  * "TI" mode - FRF in CR2 set to `1`, which puts the chip in a mode where it ignores other settings
    and sends out CRCs and stuff. (didn't explore too deeply). 

The correct answer most of the time seems to be the second one.

ChuckM:
Ok, so to add to the discussion from IRC. 

When you setup a SPI port in Master mode, you **usually** only want it in master mode. The STM32 series lets you dynamically switch into and out of master mode using the _`NSS`_ pin. My assertion is that in the _usual_ case, you will want to disable mode switching via the `NSS` pin, which you can do _either_ by setting `SSOE` in `SPI_CR2` (as I've done in my PR), **or** you set both `SSM` and `SSI` in `SPI_CR1` as is done in one of the examples by calling `spi_enable_software_slave_management` followed by `spi_set_nss_high`. And it then redundantly fixes the problem by calling `spi_enable_ss_output`.

The bug is that `spi_init_master` doesn't use either of these techniques when it is called so it leaves the SPI port incompletely set up as a master port.

PR #393, opened 2015-01-06
Fix common case SPI handling

ChuckM:
The common case for SPI ports in master mode is that they are not
also running as Slaves some times. For these chips the SSOE bit must
be set (or NSS tied high). Since it is common for people to use a separate
GPIO to select remote slaves and they expect the master to always be the master
this sets that up by default.

See issue #391

### DIFF
--- a/lib/stm32/common/spi_common_all.c
+++ b/lib/stm32/common/spi_common_all.c
@@ -392,7 +392,7 @@ void spi_set_receive_only_mode(uint32_t spi)
 }
 
 /*---------------------------------------------------------------------------*/
-/** @brief SPI Enable Slave Management by Hardware
+/** @brief SPI Disable Slave Management by Hardware
 
 In slave mode the NSS hardware input is used as a select enable for the slave.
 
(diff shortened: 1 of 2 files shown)


flixr:
So what happens if you have currently working code that calls `spi_enable_software_slave_management` and `spi_set_nss_high` after `spi_init_master` and your pull request is merged?

esden:
Do we have an update on this? I think @flixr question is valid. Will this silently break code that already is working?

ChuckM:
I've updated it per @flixr suggestion. The fix is to undo SSOE if you call 'spi_enable_software_slave_management` after you call `spi_init_master`.

ChuckM:
Some additional reasoning here which might help (would love @flixr 's feedback too). So when you set spi master mode you **must** do one of three things, either drive SS with SSOE, or set software slave management and set NSS high, or assign NSS to an IO pin and then tie that pin high. Not doing one of those things leads to a ton of frustration as the chip continually "mode faults" on SPI. There are cases in the while where you see how people dealt with this, I manually set the flags and didn't bother with spi_init_master, @karlp changed both (forced nss high and set ssoe), others have just ended up with software slave management and nss set high and a comment "why is this required?". The current fix will set SSOE by default when you call `spi_init_master`. If you call `spi_enable_software_slave_management` right after that then that function now unsets SSOE as it knows you're going to do the NSS management thing in your software. All existing code works, even code that is multimaster.

flixr:
Looks good to me :+1:  (without having had the chance to actually test it)...

ChuckM:
Can we merge this ?

karlp:
I'm looking at this over the weekend. should be good to go.

karlp:
Merged via rebase.

I tested some existing code, that still worked, and new code also still worked, and I got rid of a line in my test code.

Issue #369, opened 2014-11-27
stm32f107 connectivity line primary remap issue


brabo:
because the stm32f107 connectivity line uses in AFIO_MAPR some remap bits that are reserved in other lines (SPI3 remap for example).
this line 171 of lib/stm32/f1/gpio.c prevents these from being set:

AFIO_MAPR |= (swjdisable & AFIO_MAPR_SWJ_MASK) | (maps & 0x1FFFFF);

as a workaround i have changed this line to:

AFIO_MAPR |= (swjdisable & AFIO_MAPR_SWJ_MASK) | maps;

which is probably just fine for the f1 connectivity line, but may not be suitable for other lines.

karlp:
Fixed via the merge of https://github.com/libopencm3/libopencm3/commit/eebef017186cfbd2a4663c3dda1eb98dad6f1725 but #503  will extend this further.

PR #394, opened 2015-01-08
DRAFT: stm32: f1: afio remapping fixes for connectivity line devices

karlp:
Draft only.  I think the "VERY_SAFE" option is correct, but if you only
use the defines, then the second form is all that's needed.

For thought: This function has only ever |= remaps, there's _no_
function to clear a remap?!

Draft fix for https://github.com/libopencm3/libopencm3/issues/369

### DIFF
--- a/lib/stm32/f1/gpio.c
+++ b/lib/stm32/f1/gpio.c
@@ -159,7 +159,7 @@ available in place of some of the SWJ signals. Full SWJ capability is obtained
 by setting this to zero. The value of this must be specified for every call to
 this function as its current value cannot be ascertained from the hardware.
 
-@param[in] swjdisable Unsigned int8. Disable parts of the SWJ capability @ref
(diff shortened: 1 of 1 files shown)


ChuckM:
Ok, I'm confused by what you are trying to achieve here. could you not just have the one liner:
```
AFIO_MAPR |= (swjdisable & AFIO_MAPR_SWJ_MASK) | (maps & ~AFIO_MAPR_SWJ_MASK);
```
It seems the 'not safe' version you propose simply removes two & operations so it saves 2 instructions

karlp:
Sure, I just put it on two lines to make it easier to read.  The only volatile is the final RMW to AFIO_MAPR so it shouldn't make any code difference.  I don't really mind either way, I'll repush final code.

of more interest to me is that we provide no way of clearing any of these remaps!

karlp:
superseded by https://github.com/libopencm3/libopencm3/pull/502

PR #396, opened 2015-01-13
stm32: f0: Fixes setting of PLL multiplication factor

dmsc:
Really use RCC_CFGR_PLLMUL negated as a mask.

Signed-off-by: Daniel Serpell <daniel.serpell@gmail.com>


ChuckM:
The function goes away when we unify the rcc handling.

dmsc:
Hi, any ETA for the unified rcc handling?

Current code in head does not work reliably with HSE+PLL because of this.

karlp:
Far enough out that I'll merge this right now.  But it's very true, some RCC unification would have both a) avoided this in the first place and b) will remove this code.

PR #397, opened 2015-01-13
Regexp special characters need to be escaped inside regexp

Getty:
at least without this fix it was not working with my awk on debian stable

### DIFF
--- a/scripts/genlink.awk
+++ b/scripts/genlink.awk
@@ -31,9 +31,9 @@ BEGIN {
 	gsub(/\r$/,"");
 
 	tmp = "^"$1"$";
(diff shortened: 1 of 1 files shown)


karlp:
what are the symptoms of this failing?  how are you running this?

karlp:
@Getty any follow up?  other people on debian have not been having problems, how exactly was it failing for you?

Getty:
I must reconstruct what I was doing but this script is not part of the regular build process, so it is clear that not everybody reports it.

karlp:
merged in 0cb1db0 thanks for filing this! we should have noticed earlier that it was just mawk/gawk!

PR #400, opened 2015-01-16
[stm32] desig: high nibble to hex char first

flixr:
When converting the uniqute id to a hex string, it seems more sensible to put the higher nibble in the string before the lower nibble.


flixr:
@karlp we had already discussed this on IRC a while ago... any reason not to merge this now?

karlp:
Seems reasonable to me, only reason not to is for anyone who already has serial numbers out there using this.  We're not claiming any sort of API stability anyway though?

flixr:
considering how many other things change, no... we are not claiming API stability...

PR #403, opened 2015-01-30
[cortex.h] Fix syntax error.

mour:
Fixed syntax error. Removed extra underscore.


esden:
Thank you very much! Merged with 377f782

PR #346, opened 2014-09-10
stm32: usb otg_fs and otg_hs register definitions

amirhammad:
Most of the usb-host-related definitions were missing, so were added.

File otg_fs.h were not in the same style of definitions in otg_hs.h. This change should not affect previous functionality.
Example: 
Changed from this style:
#######
#define MMIO32(USB_OTG_FS_BASE + 0x000)
#######
to this style:
#######
#define OTG_GOTGCTL 0x000
...
#define OTG_FS_GOTGCTL MMIO32(USB_OTG_FS_BASE + OTG_GOTGCTL)
#######

In OTG_HS were bug at line 28: 
"OTG_GOTGIN" instead of "OTG_GOTGINT". 
It is then referenced (not really) in line 90.
#define OTG_HS_GOTGINT MMIO32(USB_OTG_HS_BASE + OTG_GOTGINT)

UPDATE 1: 
Created new file otg_common.h that contains common register definitions from files:
otg_hs.h and otg_fs.h

PR #405, opened 2015-02-03
stm32/usb: otg_fs and otg_hs register definitions

amirhammad:
As discussed in [pull request 346](https://github.com/libopencm3/libopencm3/pull/346), 
this commit fixes the following things:   

* USB host register definitions added.
* Extracted common register and bitfield definitions
    	from 'otg_fs.h' and 'otg_hs.h'
    	into new file 'otg_common.h'.
    	Modified usb low-level drivers to adopt to new style of bitfields.
* Fixed typo OTG_GOTGIN -> OTG_GOTGINT (according to the datasheet)

### DIFF
--- a/include/libopencm3/stm32/otg_common.h
+++ b/include/libopencm3/stm32/otg_common.h
@@ -1,0 +1,489 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2010 Gareth McMullin <gareth@blacksphere.co.nz>
+ * 
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+ 
+/*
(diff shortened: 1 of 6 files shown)


karlp:
note to self, this one looked good, need to chase it up a bit

kuldeepdhaka:
oh, i was not aware of the this PR while working on #510 
this could be another way of commonizing otg stuff.
my recommendation is 50% 50% over acceptance of #510 and this pr.

kuldeepdhaka:
please seperate the three points into 3 seperate commits.
afaif, it would be easier to review, (and maybe reuse).

though @karlp might be ok with single (3 in 1) commit.

amirhammad:
I could not separate modifying low-level drivers in separate commit since the library could not be compiled with modified headers. Although, adding usb host-related stuff, may be split into two commits. And also OTG_GOTGIN -> OTG_GOTGINT fix could be separated. 
I will wait for the opinion of @karlp .

amirhammad:
I pushed rebased commit into usb-otg-register-defs-rebased...

kuldeepdhaka:
since the otg(or device only or host only) perip is in other uC (not specific to stm32) [licenced from Synopsys Inc]. these def can be moved to directory "include/usb/perip" where they can be generalized. and commonized source to "lib/usb/periph"

kuldeepdhaka:
based on @amirhammad's "usb-otg-register-defs-rebased" 
here is a quick branch for above suggestion 
https://github.com/kuldeepdhaka/libopencm3/tree/usb-dwc-otg-register-defs

karlp:
I have thought about when the right time to try pulling out the dwc code as a separate file, rather than calling it a stm32 version would be right.  I had felt that this separation work was worth doing first, but you can make a case to do it one go.  I felt that getting all of these definitions out now would make it easier to keep moving forward later.

karlp:
merged via rebase in 6357630, after a couple of updates to handle recently merged code.

I agree that eventually we'll out the dwc_otg core into an even more shared standalone driver, but this still reduces a lot of duplication and should make it easier to do that project later.

PR #406, opened 2015-02-06
Stm32f4 ltdc

esden:
Cleaned up the newly added LTDC code. Are there any objections to using static inline instead of function like defines?

From my personal expenience the static inline functions are significantly easier to debug if something goes wrong and will function the same way in the cases used in this patch set.

Alternative would be to move them to the c file as all other convenience functions currently are, independent of the simplicity.


BuFran:
:+1: for me for static inline functions

ChuckM:
Looks good to me :+1:

esden:
Thank you guys! Merged with 5658a4ebcff368d28bc5e5bdfa574ba3651c346a

PR #408, opened 2015-02-19
Ltdc dma

kbob:
Additional register and bitfield definitions to make the lcd-dma demo compile.


esden:
Thank you very much! Merged with fd141a81313876e3dc9253d682f7314fa2a59718

PR #410, opened 2015-02-27
RCC request for comment

esden:
I am looking into the way RCC is being initialized. I do like the way it is done on F4 as it makes the code more readable and should make the code reasonably flexible without adding copy pasta functions as it is the case for F1. (the idea for F1 was that it is saving memory by being a single function and we use it only once in the code, the compiler will optimize out the other implementations)

This pull request is meant to start conversation on how we can retrieve the different resulting clocks after the RCC is set up.

Also I am adding some prefix changes in this patch to decrease the name space pollution by the library. That we could fold in into this upgrade as we will be breaking the current API anyways.

The proposed solution adds the resulting clock values to the clock scale struct. There are two alternatives to this solution that would introduce functions returning the currently set clocks.

This solution would be the basis for implementing similar approach for all the other chips supported by the library. This patch is meant only as an example.

Here is the list of possible solutions:
1. Store the clocks in the struct. (can be error prone and has copypasta issues and whoever writes the struct has to make sure they put the right values in)
2. Implement functions that return the current clock setting based on the last used scale struct. It will return wrong value if a developer decides to set the clock "by hand" instead of using the setup function and struct.
3. Implement functions that return the current clock setting based on the relevant register values set. I am not sure if we can access all of them without disturbing the system.
4. Code generation. Have a script in libopencm3 that generates static RCC functions and const variables containing everything. Might be the smallest memory footprint solution.

The solution 2 and 3 will likely cost more flash memory as the functions will likely be bigger than storing 4 additional variables in the scale struct. Also the functions have to be very careful that they take all factors into account and will likely have to be implemented differently for every chip supported by the library.

I am curious what your guys take on this is and if you can thing of even better ways of doing this that will use as little memory as possible, be accurate and introduce as little possible implementation bugs and so on.

This pull request is a continuation on the proposed solution in #286 and will impact #379

### DIFF
--- a/lib/stm32/common/usart_common_all.c
+++ b/lib/stm32/common/usart_common_all.c
@@ -48,9 +48,18 @@ usart_reg_base
 
 void usart_set_baudrate(uint32_t usart, uint32_t baud)
 {
+
+#if defined STM32F4
+	uint32_t clock = rcc_clock_scale_used->apb1_frequency;
+
+	if ((usart == USART1) ||
+	    (usart == USART6)) {
+		clock = rcc_clock_scale_used->apb2_frequency;
+	}
+#else
 	uint32_t clock = rcc_apb1_frequency;
 
-#if defined STM32F2 || defined STM32F4
+#if defined STM32F2
 	if ((usart == USART1) ||
 	    (usart == USART6)) {
 		clock = rcc_apb2_frequency;
@@ -59,6 +68,7 @@ void usart_set_baudrate(uint32_t usart, uint32_t baud)
 	if (usart == USART1) {
 		clock = rcc_apb2_frequency;
 	}
+#endif
 #endif
 
 	/*
--- a/include/libopencm3/stm32/f4/pwr.h
+++ b/include/libopencm3/stm32/f4/pwr.h
@@ -72,14 +72,14 @@ LGPL License Terms @ref lgpl_license
 
 /* --- Function prototypes ------------------------------------------------- */
 
-typedef enum {
-	SCALE1,
-	SCALE2,
-} vos_scale_t;
+enum pwr_vos_scale {
+	PWR_SCALE1,
+	PWR_SCALE2,
+};
 
 BEGIN_DECLS
 
-void pwr_set_vos_scale(vos_scale_t scale);
+void pwr_set_vos_scale(enum pwr_vos_scale scale);
 
 END_DECLS
 
--- a/include/libopencm3/stm32/f4/rcc.h
+++ b/include/libopencm3/stm32/f4/rcc.h
@@ -538,11 +538,6 @@ static inline void rcc_ltdc_set_clock_divr(uint8_t pllsaidivr)
 }
 
 /* --- Variable definitions ------------------------------------------------ */
-extern uint32_t rcc_ahb_frequency;
-extern uint32_t rcc_apb1_frequency;
-extern uint32_t rcc_apb2_frequency;
-
-/* --- Function prototypes ------------------------------------------------- */
 
 enum rcc_clock_3v3 {
 	RCC_CLOCK_3V3_48MHZ,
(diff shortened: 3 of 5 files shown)


ChuckM:
Have you seen this ? https://gist.github.com/ChuckM/0429f13af59415682e7e and the #379 PR ?

karlp:
I believe options 1) and 2) are the only sane paths.  Inspection can never determine what crystals are attached externally, and generation is something I think should be considered by individual projects looking to do marginal optimizatons specific to their needs.

karlp:
I like the prefix cleanup, and I generally like how contained the code is to just have a "used" struct.  It's never going to be fully reliable in the face of people modifying things by themselves anyway,   I like the direction this is going in more than #379.

lib/stm32/common/usart_common_all.c
```diff
@@ -48,9 +48,18 @@ usart_reg_base
...
 {
+
+#if defined STM32F4
+	uint32_t clock = rcc_clock_scale_used->apb1_frequency;
+
+	if ((usart == USART1) ||
+	    (usart == USART6)) {
+		clock = rcc_clock_scale_used->apb2_frequency;
+	}
+#else
```
ChuckM: I have suggested in (l0/f0 USART PR #390 ) that we had a chips specific `usart32_t usart_get_clock(usart)` which would be called by the common code. (you can see how it works in the PR) which avoids this sort of ifdef in the code and is a bit cleaner. In the L0/F0 case they can pick from the AHB clock, the LSE clock, or the HSI clock for their baudclock source.

ChuckM:
In general I am not a big fan of building structures like this. Fortunately clock initialization is done infrequently so its not a huge pain. Perhaps a simpler implementation would layer a set of very simple RCC register manipulating functions, a slightly higher layer of specific clock functions, and a single "easy" API which is simple `uint32_t clock_init(int mhz)` where the user passes in the speed they want and the function returns the best it can do. 

That the clock scaling with power and vcc levels seeps through the API in various places also bothers me, mostly because I feel like anyone who is coding to a minimum power foot print is probably already hand coding the registers, and second because many of the assumptions break if you are doing dynamic power use.

lib/stm32/common/usart_common_all.c
esden: Yes a convenience function for getting the uart clock is a good idea. I was more concerned with the RCC part first.

esden:
I am not sure it will be that easy to come up with an algorithm that does all of this automagically. I have the feeling there will be too many corner cases. Also how does that function know what crystal is connected to the system, unless you make that only available if you use the HSI.

I think especially in the RCC there are a few common combinations of input and output clocks. But for specific situations you will have to end up setting up your own dividers depending on your needs and how much power you want to save. Are you sure you know based on the required system clock bus what peripheral busses the user will want? Maybe they want to overclock or overclock a specific timer bus to save power or get some specific application work?

I feel it is more in spirit of the library to actually expose all the knobs instead of hiding them behind an automagical function that makes a ton of assumptions for you.

ChuckM:
I agree the structs are complicated that is why my rewrite just uses a `#define` value :smile: I'm still unclear what the 'copypasta' argument is against just having `rcc_pll_clock_setup(constant, hse_frequency);` (where `hse_frequency` is ignored if the constant is for the HSI version) And when you name your constant `RCC_8MHZ_HSE_168MHZ_PLL` its pretty easy to remember too.

ChuckM:
Oh, and you can calculate clock speeds if you know whether or not the system is running on HSE and the value of that HSE crystal. But it seems like over kill to recompute that every time when you can just cache a copy of it in a variable somewhere. My original rewrite had a `get_clock_speed({AHB, APB1, APB2})` function but it did not provide much utility over having an variable I could reference with the value in it.

karlp:
I think the typedef/enum cleanup should be merged as is.  I'll do the work to update examples.  My only concern about the second commit is that it starts the f4 deviating even further, but I think it's a good way of doing it.  We already had the globals, keeping that single struct is fine too.  _perhaps_ what's needed to go with this, is a way to let a user update that value with their own, (rather than only saving it from one of the rcc_ helper functions?) so that things like uart_setbaud_rate can still work if they've fiddled the clock later?

karlp:
Coming back to this, I still think we should merge the typedef/enum cleanup right now.

I also really like the second commit, it opens us up to having a few paths forward, and should be easier to simplify a lot of the current rcc goop.  Only real question I have is whether we want to have a single clock_used structure, that covers all the platforms? or try and have differnet ones?

I think a single one would be nice, even though it would have "lots" of redundant bits for some platforms?

karlp:
Commit 2 looks nice even more now, that I want to make some of the rcc setup routines more robustly do things in the right order!

esden:
Considering this is an ancient and likely outdated by now I will close this.

PR #412, opened 2015-03-15
Higher Level GPIO example

ChuckM:
This is a look at how you might provide a higher level GPIO API for
programs that use libopencm3, it is a demo only and not seriously
expeced to be committed.

### DIFF
--- a/include/libopencm3/stm32/pins.h
+++ b/include/libopencm3/stm32/pins.h
@@ -1,0 +1,76 @@
+/* pins.h 
+ *
+ * Example of a high level GPIO function for LIBOPENCM3 based
+ * programs.
+ *
+ * Copyright (c) 2015, Charles McManis, all rights reserved.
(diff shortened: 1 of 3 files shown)


ChuckM:
So if you had this API the blink demo would be (using the default clock)
```
include <libopencm3/stm32/pins.h>

int main(void) {
    int i;
    pin_output(PA5, PXX);
    while (1) {
        pin_toggle(PA5, PXX);asdasdasdasdasdasdf
        fori (i = 0; i < 100000; i++) {
            __asm__("nop");
        }
    }
}
```

esden:
Converted to draft considering this is just an RFC

ChuckM:
Would you be open to me adding the rest of it? (I currently have this code in my adjacent library I named libloc3util which is a collection some of these secondary but user friendly APIs)

PR #415, opened 2015-03-16
[SPI] Standard mode selection

kuldeepdhaka:
just for ease of use.


ChuckM:
This would be great with the master fix in #315 :+1:

karlp:
I've tweaked this a bit here: https://github.com/karlp/libopencm3/commit/8e2293208243684ebc25157f8664815a4b525e47

* replaced "std" with "standard"  (All other methods in this file use full words)
* split the bracketed read modify write into two lines, as is done in most other parts of the file,  (see the baud rate setting for instance)
* Tweaked the doxygen to generate a proper table:
http://imgur.com/3vZwwXv

Let me know if that's still ok with you, and I can merge it

karlp:
Merged via rebase with tweaks from above in c28d7bc18782c21429ddab9fc985cfe0f6db375e

PR #416, opened 2015-03-16
fix #303

### LINKED ISSUE #303: endpoint size 0 is hardwire though its is to fetched from usb device descriptor
usb/usb.c
in function: void _usbd_reset(usbd_device *usbd_dev)
usbd_ep_setup(usbd_dev, 0, USB_ENDPOINT_ATTR_CONTROL, 64, NULL);
instead of 64 , it should be dev->desc->bMaxPacketSize0

kuldeepdhaka:
#303


karlp:
Closed via rebase in 7dbb93c78411b37bec64b5ca5be55076b0ab1b15

Kuldeep, Note how the commit message itself is descriptive, not just "fix #303"

PR #418, opened 2015-03-25
[stm32f4-rng] Basic rng functions with documentation.

Sulter:


Sulter:
A few simple functions for the random number generator (RNG), for the STM32F4xx family devices. The current version is implemented by doing polling (no interrupts).

lib/stm32/f4/rng.c
```diff
@@ -0,0 +1,72 @@
...
+
+/*---------------------------------------------------------------------------*/
+/** @brief Returns a random number.
+When a number is randomly generated, and there was no errors, the function will return it.
+@returns data Unsigned int32.
+*/
+uint32_t rng_get_random(void)
+{
+    //Waits while the number is not ready, and/or there was an error.
+    while(((RNG_SR & RNG_SR_DRDY) != 1 || (RNG_SR & (RNG_SR_CEIS & RNG_SR_SEIS)) != 0));
```
ChuckM: Generally prefer parenthesis around expressions on either side of || or && operators. What does this do if there is an error? Is the error automatically reset or does it hang forever?

Sulter:
Thank you for your feedback ChuckM.
Those issues should now be fixed.

ChuckM:
:+1: from me.

brabo:
Tested on STM32F429i-disco, works very nice :+1:

lib/stm32/f4/rng.c
```diff
@@ -0,0 +1,93 @@
...
+
+/*---------------------------------------------------------------------------*/
+/** @brief Randomizes a number.
+@param[in] pointer to a uint32_t that will be randomized.
+@returns uint8_t, 0 on error, 1 on success.
+*/
+uint8_t rng_get_random(uint32_t *rand_nr)
+{
+    //wait for data to get ready
+    while ((RNG_SR & RNG_SR_DRDY) != 1);
```
karlp: why not just while (!(RNG_SR & RNG_SR_RDY)) ?

lib/stm32/f4/rng.c
```diff
@@ -0,0 +1,93 @@
...
+#include <libopencm3/stm32/rng.h>
+
+/**@{*/
+
+/*---------------------------------------------------------------------------*/
+/** @brief Randomizes a number.
+@param[in] pointer to a uint32_t that will be randomized.
+@returns uint8_t, 0 on error, 1 on success.
+*/
+uint8_t rng_get_random(uint32_t *rand_nr)
```
karlp: why explicitly an 8bit int?  this will require masking the return value, either just "int" or "bool" would be fine here.

karlp:
This is pretty good, but it has a lot of style issues, please run "make stylecheck" in the root directory. Particularly, the positions of the {}, spaces vs tabs, and all the // comments.  They should probably all just be dropped, they don't really provide any extra value.  Finally, I'd consider squishing this into a single commit, no reason to introduce an API then change it immediately.

karlp:
This has now landed thanks!

PR #420, opened 2015-03-29
[USB] Added support for changing callback for endpoint.

kuldeepdhaka:
Earlier, callback could only be assigned at the time of endpoint setup.

### DIFF
--- a/lib/usb/usb.c
+++ b/lib/usb/usb.c
@@ -79,11 +79,11 @@ usbd_device *usbd_init(const usbd_driver *driver,
 	usbd_dev->ctrl_buf = control_buffer;
 	usbd_dev->ctrl_buf_len = control_buffer_size;
 
-	usbd_dev->user_callback_ctr[0][USB_TRANSACTION_SETUP] =
+	usbd_dev->user_endpoint_callback[0][USB_TRANSACTION_SETUP] =
 	    _usbd_control_setup;
-	usbd_dev->user_callback_ctr[0][USB_TRANSACTION_OUT] =
+	usbd_dev->user_endpoint_callback[0][USB_TRANSACTION_OUT] =
 	    _usbd_control_out;
-	usbd_dev->user_callback_ctr[0][USB_TRANSACTION_IN] =
+	usbd_dev->user_endpoint_callback[0][USB_TRANSACTION_IN] =
 	    _usbd_control_in;
 
 	int i;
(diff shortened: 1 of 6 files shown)


lib/usb/usb.c
```diff
@@ -176,5 +176,13 @@ void usbd_ep_nak_set(usbd_device *usbd_dev, uint8_t addr, uint8_t nak)
 	usbd_dev->driver->ep_nak_set(usbd_dev, addr, nak);
 }
 
+void usbd_ep_callback_set(usbd_device *usbd_dev, uint8_t addr, void (*callback)(usbd_device *usbd_dev, uint8_t ep))
+{
+	enum _usbd_transaction dir = (addr & 0x80) ? USB_TRANSACTION_IN : USB_TRANSACTION_OUT;
```
karlp: this doesn't really appear to be a _usbd_transaction?

karlp:
Or, more fully, you're not allowing changes for the callback for SETUP transactions?  is that intended?

kuldeepdhaka:
@karlp most of the time, application want to set in/out endpoint callback only.
* non-zero control endpoint are very very rare.
* non-zero endpoint are actually useful to application (always need to be implemented as control, and ep0 callback is used by libopencm3-usb).
* control endpoint only need setup callback.

kuldeepdhaka:
"Or, more fully, you're not allowing changes for the callback for SETUP transactions? is that intended?"
@karlp if that is a problem then neither "usbd_ep_setup()" (and https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_private.h#L146) give that facility.
we need to fix them too. usbd_ep_setup() does allow creation of CONTROL endpoint but do not accept setup callback.

kuldeepdhaka:
update to use "usbd_dev->user_endpoint_callback"
preq: #483

karlp:
Ok, two things, firstly, what's the real use case for this anyway?  and secondly, I still don't get why you can't just call usbd_ep_setup() again?

kuldeepdhaka:
1) application code only _just_ want to change callback after endpoint has been created.
  my usecase: i had multiple alternate settings. when a alternate-setting from host is recv'd, i required to change callback to another part of the code. (note: size of the endpoint is same)
2) if usbd_ep_setup() is called twice, it will allocate memory twice for the endpoint. (ex: https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_f103.c#L131)

kuldeepdhaka:
note: (for checker and me) "usbd_ep_callback_set" should be "usbd_set_ep_callback".

karlp:
Ok, reason number 2 is good enough :)

PR #421, opened 2015-03-30
fix #302, validate control transfer SET_CONFIGURATION and SET_INTERFACE.

### LINKED ISSUE #302: Multiple configuration will NULL the control endpoint ( + many other related minors)
usb/usb_standard.c  usb_standard_set_configuration()
~~1) in this function , when their are multiple configuration, the loop inside the if will clear all callback (including ep0 callback) leading *doom* (device unresponsive).~~

2) their is no checking for invalid/bogus values wValue(bConfigurationNumber). as per doc usb2.0 p257 if any config. is invalid, request error should be issued. also, if bConfiguration=0 then device should be place in address state.
(their's more info on the page)

3)if (req->wValue == usbd_dev->current_config) comparision is not required since their is nothing mentioned related given configuration is already set.

kuldeepdhaka:
device will accept only vaild SET_CONFIGURATIOn and SET_INTERFACE control control transfer.
transaction will fail if host try to pass invalid configuration, interface number or interface alternate setting.

### DIFF
--- a/lib/usb/usb_standard.c
+++ b/lib/usb/usb_standard.c
@@ -248,17 +248,28 @@ static int usb_standard_set_configuration(usbd_device *usbd_dev,
 					  uint8_t **buf, uint16_t *len)
 {
 	int i;
+	const struct usb_config_descriptor *cfg;
 
 	(void)req;
 	(void)buf;
 	(void)len;
 
-	/* Is this correct, or should we reset alternate settings. */
-	if (req->wValue == usbd_dev->current_config) {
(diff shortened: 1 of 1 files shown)


karlp:
The commit comment here is not useful. Please summarize why/what in the commit comment itself, not just "fix #302"

karlp:
No, I mean in the commit itself, not just the description of the pull request. :)

kuldeepdhaka:
@karlp already done

PR #339, opened 2014-07-17
include/usbd.h: add endpoints address definition

fjullien:
This just makes code reading easier.

Signed-off-by: Franck Jullien <franck.jullien@gmail.com>

PR #422, opened 2015-03-30
usb: add endpoint address calculation macros

karlp:
Some people find it easier to read without the direction bit included.

This is intended as a replacement for: https://github.com/libopencm3/libopencm3/pull/339

Debate points are the length of the macro, but it keeps it standard with the rest of the usb defines.


karlp:
closed via rebase in 4f5863e37260e3729df366d90daad5bad8cc6051

PR #423, opened 2015-04-09
SAM3 peripheral ID, PMC, PIO

felixheld:
add the peripheral IDs, support for the power management controller and the GPIOs for all SAM3 subfamilies

### DIFF
--- a/include/libopencm3/sam/3a/pmc.h
+++ b/include/libopencm3/sam/3a/pmc.h
@@ -1,0 +1,64 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2012 Gareth McMullin <gareth@blacksphere.co.nz>
+ * Copyright (C) 2015 Felix Held <felix-libopencm3@felixheld.de>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef LIBOPENCM3_PMC_H
+#define LIBOPENCM3_PMC_H
+
+#include <libopencm3/sam/memorymap.h>
+#include <libopencm3/sam/common/pmc_common_all.h>
+#include <libopencm3/sam/common/pmc_common_3a3s3x.h>
+#include <libopencm3/sam/common/pmc_common_3a3u3x.h>
+
+/* --- Power Management Controller (PMC) registers ------------------------- */
+
+/* Peripheral Control Register */
+#define PMC_PCR				MMIO32(PMC_BASE + 0x010C)
+
+
+/* --- Register contents --------------------------------------------------- */
+
+
+/* --- PMC Master Clock Register (PMC_MCKR) -------------------------------- */
+
+/* PLLA Divide by 2 */
+#define PMC_MCKR_PLLADIV2		(0x01 << 12)
+
+
+/* --- PMC Peripheral Control Register (PMC_PCR) --------------------------- */
+
+/* Enable */
+#define PMC_PCR_EN			(0x01 << 28)
+
+/* Divisor Value */
+#define PMC_PCR_DIV_SHIFT		16
+#define PMC_PCR_DIV_MASK		(0x03 << PMC_PCR_DIV_SHIFT)
+#define PMC_PCR_DIV_PERIPH_DIV_MCK	(0x00 << PMC_PCR_DIV_SHIFT)
+#define PMC_PCR_DIV_PERIPH_DIV2_MCK	(0x01 << PMC_PCR_DIV_SHIFT)
+#define PMC_PCR_DIV_PERIPH_DIV4_MCK	(0x02 << PMC_PCR_DIV_SHIFT)
+
+/* Command */
+#define PMC_PCR_CMD			(0x01 << 12)
+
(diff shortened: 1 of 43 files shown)


include/libopencm3/sam/3n/pmc.h
```diff
@@ -0,0 +1,139 @@
...
+
+/* --- PMC Oscillator Calibration Register (PMC_OCR) ----------------------- */
+
+/* Selection of RC Oscillator Calibration bits for 12 Mhz */
+#define PMC_OCR_SEL12			(0x01 << 23)
+
+/* RC Oscillator Calibration bits for 12 Mhz */
+#define PMC_OCR_CAL12_SHIFT		16
+#define PMC_OCR_CAL12_MASK		(0x7F << PMC_OCR_CAL12_SHIFT)
+
```
ChuckM: On going discussion about masks and shifts. If you leave the mask unshifted (just the bit pattern) then code like this works
`#define get_value(reg, shift, mask) (reg) >> (shift) & (mask)`

and
`#define set_value(reg, shift, mask, val) (reg() &= ~((mask)<<(shift)); (reg) |= ((val) & (mask)) << (shift)`

include/libopencm3/sam/pio.h
```diff
@@ -17,80 +19,16 @@
...
-#define PIO_REHLSR(port)		MMIO32((port) + 0x00D4)
-#define PIO_FRLHSR(port)		MMIO32((port) + 0x00D8)
-/* 0x00DC - Reserved */
-#define PIO_LOCKSR(port)		MMIO32((port) + 0x00E0)
-#define PIO_WPMR(port)			MMIO32((port) + 0x00E4)
-#define PIO_WPSR(port)			MMIO32((port) + 0x00E8)
-/* 0x00EC-0x0144 - Reserved */
-
+#if defined(SAM3A)
+#       include <libopencm3/sam/3a/pio.h>
```
aethaniel: would be better to have sam/sam3a or sam/atsam3a for people doing web search later

felixheld:
stm32f series also uses path names like stm32/f1

include/libopencm3/sam/3a/pmc.h
```diff
@@ -0,0 +1,64 @@
...
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef LIBOPENCM3_PMC_H
+#define LIBOPENCM3_PMC_H
+
+#include <libopencm3/sam/memorymap.h>
+#include <libopencm3/sam/common/pmc_common_all.h>
+#include <libopencm3/sam/common/pmc_common_3a3s3x.h>
+#include <libopencm3/sam/common/pmc_common_3a3u3x.h>
```
karlp: How does this work to just include all of them? if they don't need to be included separately, why do they need to be in separate files at all?

include/libopencm3/sam/3n/pmc.h
```diff
@@ -0,0 +1,139 @@
...
+/* --- PMC Programmable Clock Register 2 (PMC_PCK2) ------------------------ */
+
+/* Master Clock Source Selection */
+#define PMC_PCK2_CSS_PLL_CLK		(2 << PMC_PCK2_CSS_SHIFT)
+
+
+/* --- PMC Interrupt Enable Register (PMC_IER) ----------------------------- */
+
+/* PLL Lock Interrupt Enable */
+#define PMC_IER_LOCK			PMC_IER_LOCKA
```
karlp: Is this because some other devices in this family have LOCKB/LOCKC and so on?  (Similar to ADC vs ADC1 on stm32?

include/libopencm3/sam/3a/pmc.h
felixheld: all of those files need to be included here. they need to be separate files because they contain defines for stuff that isn't present on all subfamilies and those subfamilies aren't subsets.

include/libopencm3/sam/3n/pmc.h
felixheld: The PLLA is called PLL on the devices with only one PLL. I wanted to keep stuff consistent with the datasheet here.

lib/sam/common/gpio_common_3a3u3x.c
```diff
@@ -19,16 +32,20 @@
...
+/** @brief Initialize GPIO pins
+ *
+ * @param[in] port uint32_t: GPIO Port base address
+ * @param[in] pin uint32_t
+ * @param[in] flags enum gpio_flags
+ */
 void gpio_init(uint32_t port, uint32_t pins, enum gpio_flags flags)
 {
-	switch (flags & 3) {
+	switch (flags & 0x7) {
```
karlp: Was there a define missing for this?  It's used again as a raw value lower down

lib/sam/common/pmc.c
```diff
@@ -44,20 +44,29 @@ void pmc_plla_config(uint8_t mul, uint8_t div)
 
 void pmc_peripheral_clock_enable(uint8_t pid)
 {
+#if defined(SAM3N) || defined(SAM3U)
+	//SAM3N and SAM3U only have one Peripheral Clock Enable Register
+	PMC_PCER = 1 << pid;
+#else
```
karlp: Is this actually necessary?  I mean, sure, you save a tiny tiny amount of flash, but pid will always be lower than 32 on 3n and 3u anyway right?  (Or are the defines missing?  If so, it might be better to check for #if defined(PMC_PCER1) rather than each family

lib/sam/common/gpio_common_3a3u3x.c
felixheld: thats just the mask for extracting the pin direction or connected peripheral from the enum. it changed, because on some devices are more peripheriy blocks than PERIPHA and PERIPHB, so we need 3 bits here

felixheld:
that part was already there; i only modified it so it works for all subfamilies

lib/sam/common/pmc.c
felixheld: the same register has a different name on SAM3N and SAM3U and the register for the enable bits 32 to 63 isn't there. checking for the presence PMC_PCER2 should also do the trick

felixheld:
i'll fix that

karlp:
Closed via rebase, with a minor fixup on the file modes.  Thanks!

PR #424, opened 2015-04-21
stm32f0: check correct registers for ready flags

karlp:
F0 should check the oscillator ready bits in the regular registers, just
like the docs claim, and just like every other stm32, rather than trying
to check for the interrupt flags.

Reported-by: n1b on irc


karlp:
Closed via rebase in f473d40038ad83748557834fb1e9ef7326e1142d

PR #395, opened 2015-01-12
stm32: l1: Add support for MCO setting.

iqyx:

PR #425, opened 2015-04-23
stm32: common: Extract MCO source setting

karlp:
This adds MCO source selection to some targets, and removes and standardizes
the mask/shift usage for all targets.  For devices that support MCO2, this
supports only MCO1.  No attempt has been made to extract MCO prescaler, which
is not available on all F1 and F3.

This is a replacement for https://github.com/libopencm3/libopencm3/pull/395


karlp:
merged via rebase in 5746fd4

PR #426, opened 2015-04-28
stm32: f1: Backup registers are only 16bit.

karlp:
They are only defined as 16bit, and should be accessed as such.


karlp:
RM0008r15 section 6.4 says they can be accessed as 16 or 32, so let's just leave it as 32bit wide accesses unless someone wants more

PR #427, opened 2015-04-28
[cortex.h] Added missing include

mour:
The include adds missing definitions of bool and uint32_t.

If included in code, this header will produce 'unknown type' compilation errors unless stdbool.h and stdint.h are included before it. In order for the header file to be self-sufficient, stdbool.h and stdint.h need to be included. Since most headers here seem to include common.h when they need them, I did that too.

### DIFF
--- a/include/libopencm3/cm3/cortex.h
+++ b/include/libopencm3/cm3/cortex.h
@@ -31,6 +31,8 @@
 #ifndef LIBOPENCM3_CORTEX_H
 #define LIBOPENCM3_CORTEX_H
 
+#include <libopencm3/cm3/common.h>
+
 /**@{*/
 
 /*---------------------------------------------------------------------------*/


karlp:
There's another issue for this, and, like that issue, I'm not super interested in merging partial cleanups.  If it's not using anything from common, the "right" way really is to add the stdbool/stdint includes directly, and any other file doing this should do the same thing.  We shouldn't rely on including some higher level for this sort of thing.

mour:
Ah ok, sorry, I should have checked the open issues. Just to clarify, you don't want to merge small fixes because they would clutter up the commit history? Anyway, should I close the pull request or is there a possibility that this will one day get merged? (I can change the common.h include to stdint.h & stdbool.h)

karlp:
Thanks for filing it again actually, pushed me to go and at least make improvements. :)

PR #428, opened 2015-04-30
make: use shared DEBUG_FLAGS (-gdb3) by default

karlp:
-ggdb3 make slightly bigger .elf files, but allows gdb to understand
macros, which libopenocm3 uses somewhat extensively.  Make this the
default, and pull it up to the common base makefile, so it can be easily
substituted.

--

This is just a simple cut, there's a _lot_ of common code that can be pulled up a level or two to reduce the copypasta.

### DIFF
--- a/lib/Makefile.include
+++ b/lib/Makefile.include
@@ -26,6 +26,9 @@ endif
 # common objects
 OBJS += vector.o systick.o scb.o nvic.o assert.o sync.o dwt.o
 
+# Slightly bigger .elf files but gains the ability to decode macros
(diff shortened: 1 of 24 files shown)


sgissi:
Unless someone uses a debugger other than GDB, I'm Ok to merge

flixr:
:+1:

karlp:
Well, if they use lldb or similar, they now only have to change a single make variable to tweak it to something else, instead of in a pile of different files :)

karlp:
merged via rebase in 6d658c2

PR #430, opened 2015-05-08
Atmel SAM: UART, USB and misc fixes

oskirby:
I created a fork of libopencm3 a few months ago for a project using the AT91SAM3S8xx SoCs and I wound up adding a few drivers and some miscellaneous fixes. I would like to submit my work back to the official libopencm3 project. Any feedback would be appreciated.

### DIFF
--- a/include/libopencm3/sam/eefc.h
+++ b/include/libopencm3/sam/eefc.h
@@ -79,5 +79,10 @@ static inline void eefc_set_latency(uint8_t wait)
 #endif
 }
 
+#define EEFC_UNIQUEID_SIZE      16
+
+/* Reads the 128-bit flash unique ID. */
+void eefc_read_uniqueid(uint32_t port, void *uid);
+
 #endif
 
--- a/include/libopencm3/sam/gpio.h
+++ b/include/libopencm3/sam/gpio.h
@@ -53,6 +53,11 @@ static inline void gpio_clear(uint32_t gpioport, uint32_t gpios)
 	PIO_CODR(gpioport) = gpios;
 }
 
+static inline uint32_t gpio_value(uint32_t gpioport, uint32_t gpios)
+{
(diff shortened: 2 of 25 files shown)


felixheld:
I made some bigger changes to the PIO and PMC peripherals of the whole sam3 family in pull request #423, so you probably need to rebase and modify your patches on those units a bit, so they can be applied. Also maybe call the USB peripheral something like usb_fs, since some of the chips also have a high speed USB peripheral

karlp:
I second what Felix says, about rebasing this on current code if at all possible.  However, for some of the simpler things like fixing typos, if you can submit them as standalone pulls, it's substantially easier to review and merge them quickly.

oskirby:
The DMA controllers also appear to be identical between all SAM3 families.

oskirby:
As for the weak ISR declarations, I am using the gcc-arm-embedded PPA from Terry Guo (https://launchpad.net/~terry.guo/+archive/ubuntu/gcc-arm-embedded) to build my firmware (GCC 4.9.3 for the arm-none-eabi target). What I was finding is that if a source file implementing an ISR included libopencm3/sam/3s/nvic.h, GCC would apply the weak attribute to the implementation of that function.

This in turn, would mean that both the implementation of the ISR in my sources, as well as the aliases in lib/sam/3s/vector_nvic.c were both weak. When GCC encounters two symbols that are both weak, it will not generate a warning, but it selects the first implementation that it finds for linking into the final image. For most people, I expect that they probably never notice this, since libopencm3 probably gets linked after their code. I could have worked around it by changing the order in which my libraries were linked so that vector_nvic.o came after my ISRs, but that seemed like a fragile solution.

oskirby:
I will rebase my work, and submit them as separate pull requests.

esden:
I would like to see this added to the codebase but it needs more work.

I am marking this as stale. If it is not picked up again in a month I will close this issue.

PR #433, opened 2015-05-13
sam: add UART helper functions.

oskirby:
This patch adds some helper functions to access the UART on the Atmel SAM chips. The UART appears to be common across all of the SAM3 variants. The API was chosen to be as close to the USART as possible.


karlp:
I don't see any real reason for this to be a new file with duplicated code.  Have a look at how the stm32 family handles usart vs uart.  They're both identical except for extra functions for usarts, but it's all in one file, as the registers and definitions are all the same.

PR #434, opened 2015-05-13
sam: Add register definitions for timers, reset and DMA controllers

oskirby:
This patch adds some register definitions for the reset and peripheral DMA controllers, which appear to be common.

Also included are more detail to the timer register, however, I did notice a few differences between some SAM3 variants. Most notably the SAM3U is lacking the stepper motor (TC_SMMR), fault mode (TC_FMR), and write protect (TC_WPMR) registers, and all other variants have more than one timer block.

### DIFF
--- a/include/libopencm3/sam/pdc.h
+++ b/include/libopencm3/sam/pdc.h
@@ -1,0 +1,61 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2015 Owen Kirby <oskirby@gmail.com>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef SAM3X_PDC_H
+#define SAM3X_PDC_H
+
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/sam/memorymap.h>
+
+
+/* --- PDC registers ----------------------------------------------------- */
+
+#define PDC_RPR(x)			MMIO32((x) + 0x100)
+#define PDC_RCR(x)			MMIO32((x) + 0x104)
+#define PDC_TPR(x)			MMIO32((x) + 0x108)
+#define PDC_TCR(x)			MMIO32((x) + 0x10C)
+#define PDC_RNPR(x)			MMIO32((x) + 0x110)
+#define PDC_RNCR(x)			MMIO32((x) + 0x114)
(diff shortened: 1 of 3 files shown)


karlp:
I haven't looked at this yet, but as long as there are no _conflicts_ it's ok to have the common files contain a superset.  You only need to break them out when they start conflicting.

include/libopencm3/sam/pdc.h
```diff
@@ -0,0 +1,61 @@
...
+ */
+
+#ifndef SAM3X_PDC_H
+#define SAM3X_PDC_H
+
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/sam/memorymap.h>
+
+
+/* --- PDC registers ----------------------------------------------------- */
```
karlp: Normally this is the point where you can describe what PDC means...

include/libopencm3/sam/pdc.h
```diff
@@ -0,0 +1,61 @@
...
+#define PDC_TNCR(x)			MMIO32((x) + 0x11C)
+#define PDC_PTCR(x)			MMIO32((x) + 0x120)
+#define PDC_PTSR(x)			MMIO32((x) + 0x124)
+
+/* --- PDC counter values ------------------------------------------------ */
+
+/* Bits [31:16]: Reserved. */
+#define PDC_COUNTER_MASK			(0xFFFF << 0)
+
+/* --- PDC_PTCR values --------------------------------------------------- */
```
karlp: And here you can write out the full name of what PTCR means...

include/libopencm3/sam/pdc.h
```diff
@@ -0,0 +1,61 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef SAM3X_PDC_H
+#define SAM3X_PDC_H
```
karlp: if it's sam comon, it probably shouldn't have sam3x here...

include/libopencm3/sam/rstc.h
```diff
@@ -0,0 +1,66 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef SAM3X_RSTC_H
+#define SAM3X_RSTC_H
```
karlp: same here.

include/libopencm3/sam/rstc.h
```diff
@@ -0,0 +1,66 @@
...
+#include <libopencm3/sam/memorymap.h>
+
+
+/* --- RSTC registers ---------------------------------------------------- */
+
+#define RSTC_CR				MMIO32(RSTC_BASE + 0x00)
+#define RSTC_SR				MMIO32(RSTC_BASE + 0x04)
+#define RSTC_MR				MMIO32(RSTC_BASE + 0x08)
+
+/* --- RSTC_CR values ----------------------------------------------------- */
```
karlp: same here in this file.  The TC file has done it much better, you can see what the regs are.

PR #388, opened 2014-12-18
USB/SOF problem corrected by Uwe Bonnes

BuFran:

PR #439, opened 2015-05-18
STM32 FS USB support for L0, F0, F1

kuldeepdhaka:
This pr include.
* stm32l0 usb support
* stm32f0x2 usb support
* stm32f103 and stm32f0x2 commonized code.
* fix for endpoint callback that do not read the data on time (earlier it cause infinite loop)
* pm-top configurable value
* included changes of @UweBonnes #274 and #388 for stm32f103
* disconnect support (only for stm32f0x2)

some of the fix were only meant for stm32f0x2 but since f103 was pulled in, it made the change more broad.
most of the code is old. new code is low (mostly bug fix)

### DIFF
--- a/include/libopencm3/stm32/common/usb_common_all.h
+++ b/include/libopencm3/stm32/common/usb_common_all.h
@@ -1,0 +1,309 @@
+/** @defgroup adc_defines USB Defines
+
+@brief <b>Defined Constants and Types for the STM32F1xx USB Module</b>
+
+@ingroup STM32F1xx_defines
+
+@version 1.0.0
+
+@author @htmlonly &copy; @endhtmlonly 2009
+Piotr Esden-Tempski <piotr@esden.net>
+
+@date 11 March 2013
+
+LGPL License Terms @ref lgpl_license
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2009 Piotr Esden-Tempski <piotr@esden.net>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**@{*/
+
+/** @cond */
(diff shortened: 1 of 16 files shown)


include/libopencm3/stm32/common/usb_common_all.h
```diff
@@ -0,0 +1,305 @@
...
+ *
+ * USB status field is changed using an awkward toggle mechanism, that
+ * is why we use some helper macros for that.
+ */
+#define USB_SET_EP_RX_STAT(EP, STAT) \
+	TOG_SET_REG_BIT_MSK(USB_EP_REG(EP), USB_EP_RX_STAT_TOG_MSK, STAT)
+
+#define USB_SET_EP_TX_STAT(EP, STAT) \
+	TOG_SET_REG_BIT_MSK(USB_EP_REG(EP), USB_EP_TX_STAT_TOG_MSK, STAT)
+
```
fenugrec: These macros are different from the originals in stm32/usb.h : they can cause CTR_TX and CTR_RX to be reset involuntarily.

include/libopencm3/stm32/common/usb_common_all.h
```diff
@@ -0,0 +1,305 @@
...
+ */
+#define USB_CLR_EP_NTOGGLE_BIT(EP, BIT) \
+	CLR_REG_BIT_MSK(USB_EP_REG(EP), USB_EP_NTOGGLE_MSK, BIT)
+
+#define USB_CLR_EP_RX_CTR(EP) \
+	USB_CLR_EP_NTOGGLE_BIT(EP, USB_EP_RX_CTR)
+
+#define USB_CLR_EP_TX_CTR(EP) \
+	USB_CLR_EP_NTOGGLE_BIT(EP, USB_EP_TX_CTR)
+
```
fenugrec: Same here with the USB_CLR_EP_{RX,TX}_CTR macros : the original version protected the other CTR_{TX,RX} bit from being accidentally cleared

kuldeepdhaka:
this pr is open for reference purpose until stm32{f0, l0, l1} usb code is not merged.
this pr will be divided into sub-pr and will be send for resubmitted.
(this pr will not be merged)
[applies same for #282 and #479 - both of these commits have gone through the same process]

include/libopencm3/stm32/common/usb_common_v2.h
```diff
@@ -0,0 +1,105 @@
...
+/*****************************************************************************/
+
+/* USB_CNTR Values ----------------------------------------------------------*/
+
+#define USB_CNTR_L1REQM		(1 << 7)
+#define USB_CNTR_L1RESUME	(1 << 5)
+
+/* USB_ISTR Values ----------------------------------------------------------*/
+
+#define USB_ISTR_L1REQ		(1 << 5)
```
fenugrec: This is wrong, should be 1<<7 (current code has it defined as 0x80 which is correct)

karlp:
I believe it's only disconnect support still in this pull.  I'd like to get that pulled out separately if possible, I don't think there's any major benefit in keeping this tree around futher.

PR #441, opened 2015-05-18
stm32l0 registers

rkreis:
This adds register definitions for SYSCFG, the clock recovery system and the low power timer.


rkreis:
Updating to include PWR, will reopen shortly.

PR #445, opened 2015-05-22
stm32: gpio: add gpio_get_mode()

andreamerello:
As the name suggests, this function is the reciprocal for gpio_set_mode().

I kept the function signature/parameters-type as much similar as all the other functions in gpio.c. Nevertheless, since this function works on only one pin at a time, passing an integer to identify the pin could probably also be considered. Any opinion on this ?

To extract the pin index from the bitfield I'm using GCC __builtin_ctz(). This maps to one single ASM instruction in CM3 MCUs. I assumed libopencm3 is intended to be compiled only on GCC. Is this true ?

Signed-off-by: Andrea Merello <andrea.merello@gmail.com>

### DIFF
--- a/include/libopencm3/stm32/f1/gpio.h
+++ b/include/libopencm3/stm32/f1/gpio.h
@@ -968,6 +968,8 @@ BEGIN_DECLS
 
 void gpio_set_mode(uint32_t gpioport, uint8_t mode, uint8_t cnf,
 		   uint16_t gpios);
+void gpio_get_mode(uint32_t gpioport, uint8_t *mode, uint8_t *cnf,
+	           uint16_t gpio);
 void gpio_set_eventout(uint8_t evoutport, uint8_t evoutpin);
 void gpio_primary_remap(uint32_t swjenable, uint32_t maps);
 void gpio_secondary_remap(uint32_t maps);
--- a/lib/stm32/f1/gpio.c
+++ b/lib/stm32/f1/gpio.c
@@ -131,6 +131,33 @@ void gpio_set_mode(uint32_t gpioport, uint8_t mode, uint8_t cnf, uint16_t gpios)
 }
 
 /*---------------------------------------------------------------------------*/
+/** @brief Get GPIO Pin Mode
+
+Gets the mode (input/output) and configuration (analog/digitial and
+open drain/push pull), for a given GPIO pin on a given GPIO port.
+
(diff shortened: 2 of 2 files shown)


ChuckM:
Hi, I've not tested it but the code looks reasonable. I'm curious about the use case though, generally pins are set by the program and don't get reset except under program control, so when would the program **not** know what the current state of the pin was implicitly?

andreamerello:
Hi!
Yes, as you said, there shouldn't be any configuration change unless the program itself does it.. I know that this is not really an essential feature :)

In my case I really could store pins' settings in a local variable in the program, but I'd simply like the library doing this for me, being kind enough to provide also the "get" function :)

BTW, searching for exotic use cases..
Reading the configuration back from the HW, instead of buffering it, IMHO might be useful (apart for saving few variables) in a couple of scenarios:
1) maybe if you come from a bootloader you may like to find out what it did with gpios.
2) for debugging reasons, to ensure the final configuration is correct. 

About 2), for example, this is a real case happened to me some time ago (it was another MCU, another library, but this can apply also here):
After several SW modules had completed the initialization, one gpio was found misconfigured. The SW module that actually needed that pin was indeed configuring it properly, while another SW module (ported from another board that had some differences) was overwriting the configuration by mistake while configuring its pins. This had been hard to find out because oddly the buggy SW module was still working..

I finally got this snooping the GPIO regs by using a HW debugger via JTAG, but one might be using a bootloader to flash his/her program, having no any JTAG adapter at home.. So the SW way may be the only one :)

karlp:
libopencm3 should also compile with clang, or anything really, I don't personally want to go very far adding gcc built ins...

Further, I'm also of the case that having _code_ to do this in the _library_ seems out of scope at best.  You can always read this out of registers yourself, and that's the way I certainly recommend it.  (set mode didn't work? you want to use the same code to read it back again?!)

karlp:
Further, even though I don't really have any motivation to merge this myself, if it _was_ going to be merged, it should be supported at _least_ for all stm32 families, not just for the f1.

esden:
I do think this is a useful addition. There are situations where a bootloader or something external can configure a pin and our code want's to read that out.

That said, this needs the compiler portability issue addressed and it needs to be implemented for a bit bigger selection of targets.

I will mark this as stale. Anyone is welcome to pick this up to get this added. I will close this issue in a month if no owner can be found.

PR #446, opened 2015-05-22
[usb-f103] Fix lockup due to control race

gmccollister:
If the control endpoint is in the LAST_DATA_IN state sometimes
CTR_RX and CTR_TX both become set but the OUT handler is called, resulting
in an endless stall loop. This situation doesn't happen frequently, I haven't
been able to figure exactly why it happens though I'm able to reproduce it
within a few minutes by blasting the device with non-stop control and bulk
endpoint traffic that also generates USART traffic. One theory I have is that
another interrupt is sometimes occuring after LAST_DATA_IN, giving the
STATUS_OUT complete enough time to come in.

This change will make the IN handler get called first in that situation.

Signed-off-by: George McCollister george.mccollister@gmail.com

### DIFF
--- a/lib/usb/usb_f103.c
+++ b/lib/usb/usb_f103.c
@@ -309,9 +309,18 @@ static void stm32f103_poll(usbd_device *dev)
 	if (istr & USB_ISTR_CTR) {
 		uint8_t ep = istr & USB_ISTR_EP_ID;
 		uint8_t type = (istr & USB_ISTR_DIR) ? 1 : 0;
+		uint16_t epreg = *USB_EP_REG(ep);
+
+		/* If DIR is set in USB_ISTR, we must check if CTR_TX is set in
+		 * the EP reg and if control_state.state is LAST_DATA_IN we
+		 * must deal with the IN transaction first. */
+		if (!ep && type && epreg & USB_EP_TX_CTR &&
+		    dev->control_state.state == LAST_DATA_IN) {
+			type = 0;
+		}
 
 		if (type) { /* OUT or SETUP transaction */
-			type += (*USB_EP_REG(ep) & USB_EP_SETUP) ? 1 : 0;
+			type += (epreg & USB_EP_SETUP) ? 1 : 0;
 		} else { /* IN transaction */
 			USB_CLR_EP_TX_CTR(ep);
 		}


kuldeepdhaka:
hey @gmccollister,
can you test this pr.
https://github.com/libopencm3/libopencm3/pull/439.

karlp:
The code comments say "must" but the commit comments say "I have a theory"  Anyone else have opinions on this?

zyp:
Here's my analysis of the problem:

For the DIR bit, RX take precedence meaning that the bit gets set even if there's a pending TX too. This would as stated imply that the OUT handler gets called first. This is unexpected in the LAST_DATA_IN state, so the transaction is attempted stalled instead of trying to read out the packet.

Normally the host should be able to recover from a failed control request by starting a new control request, but since _usbd_control_out() doesn't clear the buffer when called in an unexpected state, the OUT packet is still stuck there, preventing a new SETUP token from arriving and resetting the state machine.

As for the status stage OUT packet arriving too early, this happens because EP0 OUT is set ready too early. It should be left in the NAK state until we reach the STATUS_OUT stage, i.e. after receiving the LAST_DATA_IN interrupt. After all, STATUS_OUT shouldn't be ACKed before the host has read everything. Even in the NAK state, EP0 OUT will handle a new SETUP packet as long as the buffer is free.

I'd say that this patch is only a workaround, and that you should rather fix the two root causes.

kuldeepdhaka:
"After all, STATUS_OUT shouldn't be ACKed before the host has read everything"
i do not think so, and afaiu not possbile by firmware.
in usb doc, i could not find any reference that state that the host have to read all data before sending status-out.
and afaiu status-out is intented to tell that device exit the transaction exectuion state,and wait for another setup packet.

so if status-out is send before reading all by host, should we enter a unknown state? or wait for another packet?
i think the latter one is better.

zyp:
> in usb doc, i could not find any reference that state that the host have to read all data before sending status-out.

It doesn't. It can however not expect it to be ACKed either, with the reasonable result being a timeout.

What good is a status stage if you always ACK it regardless of whether the data stage finished in full or not?

> and afaiu status-out is intented to tell that device exit the transaction exectuion state,and wait for another setup packet.

The device must always be ready to accept a SETUP packet, since that would abort any unfinished control request and start a new. The status stage is for reporting whether the request completed successfully or not.

zyp:
> so if status-out is send before reading all by host, should we enter a unknown state? or wait for another packet?
> i think the latter one is better.

If you don't set EP0 OUT to ACK it before you're done sending all data, it won't even hit the software, so we'll remain in the current state, waiting. If we get stuck there, we'll be able to recover when the next SETUP token arrives, assuming the issue of clearing the RX buffer also gets resolved.

For the reported problem, we would transition into the STATUS_OUT stage before the host times out, so all that'll happen is that the host would have to retry a few times.

karlp:
This seems like a valid bug worthy of looking at further, but I'd like to look at it again now after the st-usbfs merge, and ideally see if we can setup a test case to exercise this.

karlp:
I believe this should be resolved via the merge of 26ab78a7 but if not, we'll need tests to investigate further.

karlp:
on re-reading, this is the "endless loop" that's occasioanlyl referred to.

karlp:
or, that was fixed by #785, and the comment referenced above fixed a regression from that.  Either way, there's been a variety of usb fixes, both in core, and in platform code since then.

PR #457, opened 2015-05-29
partially fixed #454 adc power off

### LINKED ISSUE #454: stm32f0/f3 ADC poweroff
According to the reference manual (DocID018940 Rev 7, page 233) following procedure should be used to disable an ADC.

*Set ADDIS=1 to disable the ADC and put the ADC in power down mode. The ADEN
and ADDIS bits are then automatically cleared by hardware as soon as the ADC is fully
disabled.*

But libopencm3 instead tries to clear the ADEN bit, which is not possible according to the manual (the bit has `rs` access).

This should be the correct way:
```C
    ADC_CR(ADC) |= ADC_CR_ADDIS;
    while (ADC_CR(ADC) & ADC_CR_ADEN);
```
In order to re-calibrate an ADC it is essential to make sure the ADC is fully powered off i.e. `ADEN=0`, so there should be a call like `adc_power_off_wait_finish(...)` since `adc_power_off(...)` is async.

Relevant implementation:
https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/f0/adc.c

trichner:
This is the correct way to disable the ADC according to the reference manual. Note that this is asynchronous, the ADC will not stop immediatly, you need to check the ADEN flag to ensure its actually down.
Smoke-tested.

### DIFF
--- a/include/libopencm3/stm32/f3/adc.h
+++ b/include/libopencm3/stm32/f3/adc.h
@@ -870,7 +870,7 @@
 BEGIN_DECLS
 
 void adc_power_on(uint32_t adc);
-void adc_off(uint32_t adc);
+void adc_power_off(uint32_t adc);
 void adc_enable_analog_watchdog_regular(uint32_t adc);
 void adc_disable_analog_watchdog_regular(uint32_t adc);
 void adc_enable_analog_watchdog_injected(uint32_t adc);
--- a/lib/stm32/f0/adc.c
+++ b/lib/stm32/f0/adc.c
@@ -471,7 +471,7 @@ void adc_disable_eoc_interrupt(uint32_t adc)
 
 void adc_power_off(uint32_t adc)
 {
-	ADC_CR(adc) &= ~ADC_CR_ADEN;
+	ADC_CR(adc) |= ADC_CR_ADDIS;
 }
 
 /*---------------------------------------------------------------------------*/
--- a/lib/stm32/f3/adc.c
+++ b/lib/stm32/f3/adc.c
@@ -99,7 +99,7 @@
 
 void adc_off(uint32_t adc)
 {
-	ADC_CR(adc) &= ~ADC_CR_ADEN;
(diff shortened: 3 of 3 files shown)


karlp:
Can you please fix this for F3 as well?

trichner:
I have no stm32f3 hardware around, so cannot test.

karlp:
it doesn't need to be tested on the hardware, it's a pretty clear bug from the manual, with a clear simple fix.

trichner:
just letting you know its untested :)

karlp:
11:51 < n1b> I suspect there are some more issues there, I had problems setting ADDIS and then waiting for ADEN 
             to go to low.
11:51 < n1b> seemed to hang

trichner:
I suggest you merge it as it is the correct implementation (according to the reference manual for STM32F0). The issues I had might as well be in other parts in my project.

trichner:
any status update?

trichner:
What stops this from merging?

karlp:
round tuits mostly.  It certainly wasn't helpful that you said, "even with this, it doesn't work" and didn't follow up on it.  I haven't found the time to follow up on it myself either, so it's in the queue of "looks good, but seemed like it needed some verifying"   Also, adding in the rename (which is a good idea) means someone (me) wants to review the documentation and the common sections, make sure they're all documented as doing the same thing, make sure there's no slight differences in the different families, and that all families end up getting the same thing.  It's never quite as simple as "I'll just fix this for the one chip I'm working on"  The current string of commits don't have great titles or text, but generally just a lack of maintainer time.  I'm currently ignoring all but the most basic fixes that happen to land fully formed and easy to verify until I wrap up the USB infrastructure testing and merging.   It will happen, and when I have free time I do normally start with the oldest pulls and go through them methodically.

trichner:
Thanks for the quick response, if I find the time I'll do some testing on STM32F0 and report back.

karlp:
this has now landed (equivalents) via a large tree of adc cleanups for f0/f3/l4

Issue #236, opened 2014-01-10
Populate current processor clock speed


BuFran:
Current versions of RCC module populates variables  rcc_ppre1_frequency and  rcc_ppre1_frequency, that has not descriptive name.

The idea of this issue is to create a global set of variables, common for all stm32 family, correctly filled with current clock speeds of appropriate busses.

rcc_frequency_core
rcc_frequency_apb1
rcc_frequency_apb2
rcc_frequency_ahb

The core clock speed information is missing in the current lib.

mox-mox:
I just ran across the same issue and created a quick patch to populate at least the AHB clock speed (which should be equal to the core speed, correct?). It just adds one data member to the clock_scale_t struct which defines the AHB speed and sets rcc_ahb_frequency in void rcc_clock_setup_hse_3v3(const clock_scale_t *clock).

See https://gist.github.com/mox-mox/e21ee91e9ee9417badb5

Could you please merge this into libopencm3?

karlp:
@mox-mox please, if you have changes you want commited, make a regular pull request for them.

mox-mox:
Done :)
https://github.com/libopencm3/libopencm3/pull/467

karlp:
I've been back looking at things like this, and I'm now convinced of the validity of keeping the existing globals and adding one for core clock.

mox-mox:
Honestly, I don't care anymore.

karlp:
no, that's fine, I'm just pulling in parts of it now, and leaving comments for anyone else following the ticket

karlp:
Via 5738e9515d1a83dab72aada1865c3a2982754bea and the partial commits before it, all of the ahb frequencies are in, doesn't seem to be any particular need for a separate core clock.

PR #467, opened 2015-06-29
Partial fix for "Populate current processor clock speed #236"

mox-mox:
I added a small fix for those STM32 processors that use the same clock initialisation system as the STM32F4, that will set the variable rcc_ahb_frequency just like rcc_apb1_frequency and rcc_apb2_frequency when the clock is initialised.
This work is tested on my board with an STM32F429IIT6 and works there, all other CPUs are untested but this is a very small change that really should not cause any problems.


karlp:
I belive, after 5738e9515d1a83dab72aada1865c3a2982754bea all the rcc_ahb_frequency changes have come in.  The rest of this branch is out of scope for this pull request however.

PR #471, opened 2015-07-09
Adc

brabo:


karlp:
a bunch of power_on/power_off and adc api cleanup has just landed, this is one of the things that came in with it.

PR #472, opened 2015-07-15
Removed include/libopencm3/stm32/usb_desc.h

### LINKED ISSUE #462: stm32/usb_desc.h?
https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/stm32/usb_desc.h
what is usb_desc.h doing in stm32/

kuldeepdhaka:
as it is not required and include/libopencm3/usb/usbstd.h should be used instead.
fixes #462


karlp:
merged via rebase after editing the commit log

PR #473, opened 2015-07-15
validate SET_INTERFACE control requests

### LINKED ISSUE #302: Multiple configuration will NULL the control endpoint ( + many other related minors)

kuldeepdhaka:
Transaction will fail if host try to pass invalid configuration,
interface number or interface alternate setting.

Validate set-configuration wValue,
if wValue > 0, confirm that device have the configuration.
and do the reset, init part for alternate settings.

fixes #302

### DIFF
--- a/lib/usb/usb_standard.c
+++ b/lib/usb/usb_standard.c
@@ -18,6 +18,7 @@ LGPL License Terms @ref lgpl_license
  * This file is part of the libopencm3 project.
  *
  * Copyright (C) 2010 Gareth McMullin <gareth@blacksphere.co.nz>
+ * Copyright (C) 2015 Kuldeep Singh Dhaka <kuldeepdhaka9@gmail.com>
  *
  * This library is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
@@ -245,18 +246,31 @@ static int usb_standard_set_configuration(usbd_device *usbd_dev,
 					  struct usb_setup_data *req,
 					  uint8_t **buf, uint16_t *len)
 {
-	int i;
+	unsigned i;
+	const struct usb_config_descriptor *cfg;
 
 	(void)req;
 	(void)buf;
 	(void)len;
 
-	/* Is this correct, or should we reset alternate settings. */
-	if (req->wValue == usbd_dev->current_config) {
-		return 1;
+	if(req->wValue > 0) {
+		if (req->wValue > usbd_dev->desc->bNumConfigurations) {
+			return USBD_REQ_NOTSUPP;
+		}
 	}
 
 	usbd_dev->current_config = req->wValue;
+
+	if (usbd_dev->current_config > 0) {
+		cfg = &usbd_dev->config[usbd_dev->current_config - 1];
+
+		/* reset all alternate settings configuration */
+		for (i = 0; i < cfg->bNumInterfaces; i++) {
+			if (cfg->interface[i].cur_altsetting) {
+				*cfg->interface[i].cur_altsetting = 0;
+			}
+		}
+	}
 
 	/* Reset all endpoints. */
 	usbd_dev->driver->ep_reset(usbd_dev);
@@ -299,10 +313,16 @@ static int usb_standard_set_interface(usbd_device *usbd_dev,
 				      struct usb_setup_data *req,
 				      uint8_t **buf, uint16_t *len)
 {
-	const struct usb_config_descriptor *cfx = &usbd_dev->config[usbd_dev->current_config - 1];
+	const struct usb_config_descriptor *cfx;
 	const struct usb_interface *iface;
 
 	(void)buf;
+
+	if (!usbd_dev->current_config) {
+		return USBD_REQ_NOTSUPP;
+	}
+
+	cfx = &usbd_dev->config[usbd_dev->current_config - 1];
 
 	if (req->wIndex >= cfx->bNumInterfaces) {
 		return USBD_REQ_NOTSUPP;
@@ -335,7 +355,13 @@ static int usb_standard_get_interface(usbd_device *usbd_dev,
 				      uint8_t **buf, uint16_t *len)
 {
 	uint8_t *cur_altsetting;
-	const struct usb_config_descriptor *cfx = &usbd_dev->config[usbd_dev->current_config - 1];
+	const struct usb_config_descriptor *cfx;
+
+	if (!usbd_dev->current_config) {
+		return USBD_REQ_NOTSUPP;
+	}
+
+	cfx = &usbd_dev->config[usbd_dev->current_config - 1];
 
 	if (req->wIndex >= cfx->bNumInterfaces) {
 		return USBD_REQ_NOTSUPP;


lib/usb/usb_standard.c
```diff
@@ -247,19 +248,30 @@ static int usb_standard_set_configuration(usbd_device *usbd_dev,
 					  struct usb_setup_data *req,
 					  uint8_t **buf, uint16_t *len)
 {
-	int i;
+	uint8_t i;
```
karlp: there's absolutely no reason for doing this.

lib/usb/usb_standard.c
```diff
@@ -247,19 +248,30 @@ static int usb_standard_set_configuration(usbd_device *usbd_dev,
...
+		return USBD_REQ_NOTSUPP;
 	}
 
 	usbd_dev->current_config = req->wValue;
 
+	if (usbd_dev->current_config) {
+		cfg = &usbd_dev->config[usbd_dev->current_config - 1];
+
+		/* reset all alternate settings configuration */
+		for(i=0; i < cfg->bNumInterfaces; i++) {
```
karlp: style

lib/usb/usb_standard.c
```diff
@@ -247,19 +248,30 @@ static int usb_standard_set_configuration(usbd_device *usbd_dev,
...
+	const struct usb_config_descriptor *cfg;
 
 	(void)req;
 	(void)buf;
 	(void)len;
 
-	/* Is this correct, or should we reset alternate settings. */
-	if (req->wValue == usbd_dev->current_config) {
-		return 1;
+	if (req->wValue > usbd_dev->desc->bNumConfigurations) {
```
karlp: >= here I think?  if you only have 1 configuration, then the only wValue that we accept is 0 right?  wValue of 1 would be invalid wouldn' tit?

lib/usb/usb_standard.c
```diff
@@ -247,19 +248,30 @@ static int usb_standard_set_configuration(usbd_device *usbd_dev,
...
+		return USBD_REQ_NOTSUPP;
 	}
 
 	usbd_dev->current_config = req->wValue;
 
+	if (usbd_dev->current_config) {
```
karlp: So this is resetting the altsetting for all interfaces in this what, the current config?  Why does it need to do that, if we're going to be switching to a new config anyway?

lib/usb/usb_standard.c
```diff
@@ -301,11 +313,17 @@ static int usb_standard_set_interface(usbd_device *usbd_dev,
 				      struct usb_setup_data *req,
 				      uint8_t **buf, uint16_t *len)
 {
-	const struct usb_config_descriptor *cfx = &usbd_dev->config[usbd_dev->current_config - 1];
+	const struct usb_config_descriptor *cfx;
 	const struct usb_interface *iface;
 
 	(void)buf;
 
+	if (! usbd_dev->current_config) {
```
karlp: style

lib/usb/usb_standard.c
```diff
@@ -301,11 +313,17 @@ static int usb_standard_set_interface(usbd_device *usbd_dev,
...
+		return USBD_REQ_NOTSUPP;
+	}
```
karlp: Is it really necessary to add this check?  It seems like paranoia.  Can you ever set_interface without having a config?

lib/usb/usb_standard.c
```diff
@@ -337,7 +355,13 @@ static int usb_standard_get_interface(usbd_device *usbd_dev,
 				      uint8_t **buf, uint16_t *len)
 {
 	uint8_t *cur_altsetting;
-	const struct usb_config_descriptor *cfx = &usbd_dev->config[usbd_dev->current_config - 1];
+	const struct usb_config_descriptor *cfx;
+
+	if (! usbd_dev->current_config) {
+		return USBD_REQ_NOTSUPP;
+	}
```
karlp: same questions here, same style problem too if you want it.

lib/usb/usb_standard.c
kuldeepdhaka: yes, if wValue == 0, device goes into address state.
else if wValue >= bNumConfigurations: device goes into configured state.
[bConfigurationValue (in config desc) are value >= 1]
clearly my fault

lib/usb/usb_standard.c
kuldeepdhaka: this is initalizing the new config alternate settings value to 0.
this has to be done every time to a new config (either in user code or ocm3 code).
so, it is better to do in ocm3 because these values are also used by ocm3 too.

lib/usb/usb_standard.c
kuldeepdhaka: usb_20.pdf (p259)
9.4.10 Set Interface
Default state: Device behavior when this request is received while the device is in the Default state is not specified.
Address state: The device must respond with a Request Error.
Configured state: This is a valid request when the device is in the Configured state.

if the usb stack is tested for usual bugs, this will come up very nicely
also, the behaviour has been clearly mentioned.

lib/usb/usb_standard.c
kuldeepdhaka: though not a big deal, but i usually use unsigned values for indexing array.
also, bNumInterfaces is uint8_t and imo using int will only add confusion

kuldeepdhaka:
@karlp update everything except the below one.

what is wrong with "if (! usbd_dev->current_config) {"?
do you mean to say "if (usbd_dev->current_config == 0) {"?

lib/usb/usb_standard.c
karlp: then just use "unsigned" if you must.  or "uint8_least_t" or something that will let it use a bigger variable if needs to.  we're on a 32bit platform.

lib/usb/usb_standard.c
```diff
@@ -247,19 +248,40 @@ static int usb_standard_set_configuration(usbd_device *usbd_dev,
...
 
 	(void)req;
 	(void)buf;
 	(void)len;
 
-	/* Is this correct, or should we reset alternate settings. */
-	if (req->wValue == usbd_dev->current_config) {
-		return 1;
+	if(req->wValue > 0) {
+		if (req->wValue > usbd_dev->desc->bNumConfigurations) {
```
karlp: shoulnd't this be >= to as well? same reason as in some other pull we were looking at?

lib/usb/usb_standard.c
karlp: right, I must have missed that usbd_dev->current_config had already been updated on line 264, I was thinking this was clearing out the alt settings on the _old_ config.

lib/usb/usb_standard.c
```diff
@@ -337,13 +375,26 @@ static int usb_standard_get_interface(usbd_device *usbd_dev,
...
 
 	if (req->wIndex >= cfx->bNumInterfaces) {
 		return USBD_REQ_NOTSUPP;
 	}
 
 	*len = 1;
+	/* FIXME: assuming (req->wIndex) offset
+	 *   gives the interface descriptor "iface" which has
+	 *   iface->bInterfaceNumber == req->wIndex and
+	 *     iface->bAlternateSetting == 0 */
```
karlp: are these really fixmes? isn't this how it's kept?  it's not an assumption, they _are_ kept like that by the stack?

kuldeepdhaka: right, but then we should document that the config's array provided in the exact format. (same for interface array).
most examples are following the norm but, we still need to document that.
https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/usb/usbstd.h#L165
https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/usb/usbd.h#L64
note: the descriptor is _provided_ by application code. and is "kept" inside.
i will remove all these FIXME if we put it somewhere to follow the norm in doc

lib/usb/usb_standard.c
kuldeepdhaka: when the host send a configuration value 0, it goes to address state (ie unconfigured).
afaiu, this should not fail at all if configuration value 0
```
Default state: Device behavior when this request is received while the device is in the Default state is not specified.

Address state: If the specified configuration value is zero, then the device remains in the Address
state. If the specified configuration value matches the configuration value from a
configuration descriptor, then that configuration is selected and the device enters the
Configured state. Otherwise, the device responds with a Request Error.

Configured state: If the specified configuration value is zero, then the device enters the Address state. If the specified configuration value matches the configuration value from a
configuration descriptor, then that configuration is selected and the device remains in
the Configured state. Otherwise, the device responds with a Request Error.
```

karlp:
Some more descriptive text for reference, "it validate set-configuration wValue, if wValue > 0, confirm that we do have  the configuration. and do the reset, init part for alternate settings"

kuldeepdhaka:
use this script to validate the pr

don't forget to change idVendor and idProduct.
any usb libopencm3-example can be used to test this pr (with and without pr).

```
#/bin/python
import usb.core

dev = usb.core.find(idVendor=0xcafe, idProduct=0xcafe)
if dev is None:
	raise ValueError('Device not found')

try:
	dev.ctrl_transfer(0x00,0x09, 0x0A)
	print("failed, device accepted. Bad!")
	print("fix not applied, apply fix")
except usb.core.USBError as e:
	print("failed, device rejected. Good! exception=```%s'''" % str(e))
	print("fix works :)")
```

karlp:
or... (for people who don't know what allt he magic numbers in a control request are)

```
#/bin/python
import usb.core
import usb.control

dev = usb.core.find(idVendor=0xcafe, idProduct=0xcafe)
if dev is None:
    raise ValueError('Device not found')

try:
    dev.set_configuration(69) # Any unknown configuration index
    raise ValueError("Should not accept illegal configuration values")
except usb.core.USBError as e:
    print("Correctly failed to set an invalid configuration, device replied with exception:" ,e)
```

kuldeepdhaka:
@karlp your version will not  work. (i tried your way first, but it failed)

karlp:
If it doens't work, it should be reported as a bug in pyusb.

kuldeepdhaka:
@karlp it is not a pyusb problem.
when an invalid bConfigurationValue is passed.
https://github.com/walac/pyusb/blob/master/usb/core.py#L126
it try to search the configuration before proceeding

kuldeepdhaka:
@karlp added descriptive text in commit message.

lib/usb/usb_standard.c
karlp: This is currently failing the (admittedly pedantic) test case of having two configurations with bConfigurationValue of 2 and 3 (there's absolutely no requirement on those values being zero based and sequential)  I've fixed in my merge branch to scan the set of known configs.

kuldeepdhaka: yes, i know that. when i raised the same issue, i was told that they need to be sequencial and must start from 1.
same design problem for interface-number, interface-alternate-setting.

karlp:
the configuration portion of this has landed, set interface still needs validation

karlp:
There's also the following issue for set interface: https://github.com/libopencm3/libopencm3/issues/526  These go hand in hand.

PR #474, opened 2015-07-15
SOF enable/disable notification to backend.

kuldeepdhaka:
SOF should be enabled from backend when requested.
Backend will enable/disable SOF once not every time in poll().

### DIFF
--- a/lib/usb/usb.c
+++ b/lib/usb/usb.c
@@ -18,6 +18,7 @@ LGPL License Terms @ref lgpl_license
  * This file is part of the libopencm3 project.
  *
  * Copyright (C) 2010 Gareth McMullin <gareth@blacksphere.co.nz>
+ * Copyright (C) 2015 Kuldeep Singh Dhaka <kuldeepdhaka9@gmail.com>
  *
  * This library is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
@@ -114,6 +115,11 @@ void usbd_register_resume_callback(usbd_device *usbd_dev,
 void usbd_register_sof_callback(usbd_device *usbd_dev, void (*callback)(void))
 {
 	usbd_dev->user_callback_sof = callback;
+
+	/* backend enable/disable */
+	if (usbd_dev->driver->enable_sof) {
+		usbd_dev->driver->enable_sof(usbd_dev);
+	}
 }
 
 void _usbd_reset(usbd_device *usbd_dev)
--- a/lib/usb/usb_f103.c
+++ b/lib/usb/usb_f103.c
@@ -2,6 +2,7 @@
  * This file is part of the libopencm3 project.
  *
  * Copyright (C) 2010 Gareth McMullin <gareth@blacksphere.co.nz>
+ * Copyright (C) 2015 Kuldeep Singh Dhaka <kuldeepdhaka9@gmail.com>
  *
  * This library is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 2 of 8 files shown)


lib/usb/usb_fx07_common.c
```diff
@@ -342,3 +336,12 @@ void stm32fx07_disconnect(usbd_device *usbd_dev, bool disconnected)
 		REBASE(OTG_DCTL) &= ~OTG_FS_DCTL_SDIS;
 	}
 }
+
+void stm32fx07_register_sof_callback(usbd_device *usbd_dev)
```
karlp: I like what you're doing with this PR, but I don't think the naming is right.  The outer layer registers a callback, the innerlayer is enabling sof interrupts.

lib/usb/usb_private.h
```diff
@@ -157,6 +157,7 @@ struct _usbd_driver {
 				   void *buf, uint16_t len);
 	void (*poll)(usbd_device *usbd_dev);
 	void (*disconnect)(usbd_device *usbd_dev, bool disconnected);
+	void (* register_sof_callback)(usbd_device *usbd_dev);
```
karlp: so this is really a "enable sof interrupts" ...

kuldeepdhaka: i could not come up with a better one, please suggest some.

kuldeepdhaka:
@karlp riped.

lib/usb/usb_private.h
```diff
@@ -157,6 +158,7 @@ struct _usbd_driver {
 				   void *buf, uint16_t len);
 	void (*poll)(usbd_device *usbd_dev);
 	void (*disconnect)(usbd_device *usbd_dev, bool disconnected);
+	void (* enable_sof_callback)(usbd_device *usbd_dev);
```
karlp: Even with what you've got here, the backend method is _not_ "enable_callback", you're enabling the IRQ in the backend method.  The frontend method is enabling the callback (and providing said callback) the frontend is just asking to enable the SOF IRQ so that it can dispatch the callback.

lib/usb/usb.c
```diff
@@ -114,6 +115,11 @@ void usbd_register_resume_callback(usbd_device *usbd_dev,
 void usbd_register_sof_callback(usbd_device *usbd_dev, void (*callback)(void))
 {
 	usbd_dev->user_callback_sof = callback;
+
+	/* backend enable/disable */
+	if (usbd_dev->driver->enable_sof_callback) {
```
karlp: is this really optional?  I kinda feel that if you're going to add support for registering a SOF callback, then it's either always available, or the feature is not available at all, without manually handling it.  Do you really need to check for this?  (maybe an assert... even though we have no asserts right now.....)

lib/usb/usb_f103.c
```diff
@@ -41,6 +42,7 @@ static uint16_t stm32f103_ep_write_packet(usbd_device *usbd_dev, uint8_t addr,
 static uint16_t stm32f103_ep_read_packet(usbd_device *usbd_dev, uint8_t addr,
 					 void *buf, uint16_t len);
 static void stm32f103_poll(usbd_device *usbd_dev);
+static void stm32f103_enable_sof_callback(usbd_device *dev);
```
karlp: I do feel, personally, that as a backend method name, that this should either be, "enable sof interrupt" or "enable interrupt(sof)"  but it has nothing to do with a callback in this layer right?

kuldeepdhaka:
another way of doing this could 
"usbd_register_sof_callback(dev, cb) { dev->driver->register_sof_callback(dev, cb) }"
and the backend do what ever it want.
backend has to do "usbd_dev->user_callback_sof = cb" itself.

karlp:
yeah, but I'm not sure that the backend should know about the callback, I like the way you have already done it in that regard...

kuldeepdhaka:
updated

PR #477, opened 2015-07-20
32F103 : fix USB_COUNTn_RX buffer size calculation

### LINKED ISSUE #476: USB: st_usbfs : bad ep_rx_bufsize calculation
in lib/usb/usb_f103.c ::usb_set_ep_rx_bufsize() ( https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_f103.c#L88 ), the code will produce invalid values for the NUM_BLOCK field of the USB_COUNTn_RX register (refer to RM0008 table 177 / RM0091 table 127) :
- sizes <= 62 are rounded up (ok) but not shifted >>1
- sizes > 62 are rounded down (inconsistent) and not re-adjusted
- (minor) : size==0 is not checked; but is not allowed for NUM_BLOCK

The effect of this is that the actual size set (and used by the peripheral) is larger than the requested size.

This is conjecture, but a bug (admittedly very improbable) would occur if :
- the user sets a first EP RX buffer size, and assumes the next buffer (could be any EP buffer) starts at &RX_BUFFER + <asked_size>,
- data is written at the real &TX_BUFFER (not where the user expected)
- the host sends a packet that is longer than <asked_size>, but respects the actual size set in the USB_COUNTn_RX register. This over-writes the next buffer in packet memory.

fenugrec:
Fix #476 ; good for F102, F103 and F0 devices.


lib/usb/usb_f103.c
```diff
@@ -81,6 +81,7 @@ static void stm32f103_set_address(usbd_device *dev, uint8_t addr)
 
 /**
  * Set the receive buffer size for a given USB endpoint.
+ * Valid for F102, F103 and F0 devices
```
karlp: no need for this here, it's already out of date, and is actually "good for the synopsis dwc usb core used here" or whatever the name is. just leave this out.

lib/usb/usb_f103.c
```diff
@@ -89,16 +90,18 @@ static void usb_set_ep_rx_bufsize(usbd_device *dev, uint8_t ep, uint32_t size)
...
 	(void)dev;
 	if (size > 62) {
-		if (size & 0x1f) {
-			size -= 32;
-		}
-		USB_SET_EP_RX_COUNT(ep, (size << 5) | 0x8000);
+		size += 31;	//round up to nearest mult. of 32
+		size >>= 5;
+		size -= 1;
+		size |= 0x20;	//set BL_SIZE bit
```
karlp: no C++ comments please.  describe the algorithm, if necessary, in a block comment.

karlp:
The commit comment itself should describe what was wrong, and why it needs to be changed.  commit comments like "fix x" aren't enough.  By allmeans include a link to "this fixes github issue: http://blahlbhalbha" but the commit log should be useful without linking out.

fenugrec:
For reference, I wrote a test program to compare three implementations of the buffer size calculation:
 * f1 : current locm3 code (seems to be imperfectly based on ST code)
 * f2 : new code (this PR), based on datasheet
 * f3 : original ST code from their USB_FS lib

Here is the output of the test program. Values are the top 6 bits of USB_COUNTn_RX == (BL_SIZE | NUM_BLOCK)
```
requested size      f1 (old locm3)  f2 (new)        f3 (from ST)
 1 (0x01)       0x02            0x01            0x01
 2 (0x02)       0x02            0x01            0x01
 3 (0x03)       0x04            0x02            0x02
31 (0x1F)       0x20            0x10            0x10
32 (0x20)       0x20            0x10            0x10
33 (0x21)       0x22            0x11            0x11
63 (0x3F)       0x20            0x21            0x21
64 (0x40)       0x22            0x21            0x21
65 (0x41)       0x21            0x22            0x22
96 (0x60)       0x23            0x22            0x22
```
Unfortunately github gist doesn't work here(?), so the source is temporarily pastebinned @ http://pastebin.ca/3105120

rkreis:
Notably, the old code configures an RX buffer of 96 bytes for the control endpoint when bMaxPacketSize0 is set to 64 (a very common case). This causes a silent buffer overflow when long control requests are received, corrupting the next endpoint buffer.

This issue actually hit me in a real application, causing a lot of headache and debugging. If only this PR had been merged...

karlp:
@rkreis so, given that that's used here for instance: https://github.com/libopencm3/libopencm3/blob/master/tests/gadget-zero/usb-gadget0.c#L66  what is missing in the tests to have been catching this in the past?

Do I need to interleave some bulk in/out blah requests with a handful of >64 byte control requests and it should fail then?

PR #480, opened 2015-07-26
Remove "usbd_cable_connect" function defination, as it no where defined.

kuldeepdhaka:


karlp:
Anyone know what this was meant to be for?  Is this meant to be something to do with VBUS detection?

kuldeepdhaka:
@uwehermann @esden can give some hints?

karlp:
Merged via rebase in 57c7ba3  I agree, if it's not used anywhere, and no-one knows what it's for, we can just drop it.  If something like this becomes desirable, it can be re-added.

PR #481, opened 2015-07-26
Use enum usbd_request_return_codes instead of hardcoded integer values.

kuldeepdhaka:


karlp:
I like it.  Do you feel really strongly about adding your name to usb_standard.c though?  Other that that, this is totally good to go, and looks like "a good thing"

kuldeepdhaka:
nah, just because the previous year is 2010, i though it would be a good idea to add 2015 :)

kuldeepdhaka:
@karlp updated usb_standard.c

ChuckM:
:+1:  from me.

karlp:
Actually, shouldn't this also include changing the function signatures to return the enum too?

I believe the full fix is actually https://github.com/karlp/libopencm3/commit/219ee2cb9588e99e843ee822fe2f03b6025927ee with the internals being updated too correct?

kuldeepdhaka:
#497 is a superset.

PR #482, opened 2015-07-26
Added some typedef to ease in future.

kuldeepdhaka:


karlp:
I like it, this all seems like good, clear, straightforward improvements.

ChuckM:
:+1:

karlp:
merged via rebase in 4b89272  Note the expanded commit comments.  (I removed the copyright line you added, as you didn't add it to the docs above, and there's no code changes in this one, and you only added it to one file.  let's not just start putting all git contributors into the header of every file. if we can avoid it)

PR #484, opened 2015-07-26
Added basic working I2C support for all STM32 MCUs

KivApple:
The current version of the functions to work with I2C in libopencm3 for STM32 microcontrollers, in principle, does not allow to work with I2C bus - completely missing functions to check I2C module status. I realized the main (not all, but basic) functions that allow you to check the conditions necessary for the I2C master mode.

### DIFF
--- a/lib/stm32/common/i2c_common_all.c
+++ b/lib/stm32/common/i2c_common_all.c
@@ -458,59 +458,77 @@ void i2c_clear_dma_last_transfer(uint32_t i2c)
 	I2C_CR2(i2c) &= ~I2C_CR2_LAST;
 }
 
+/*---------------------------------------------------------------------------*/
+/** @brief I2C check bus busy
+
+@param[in] i2c Unsigned int32. I2C register base address @ref i2c_reg_base.
+*/
 uint8_t i2c_busy(uint32_t i2c)
 {
-	if ((I2C_SR2(i2c) & I2C_SR2_BUSY) != 0) {
-		return 1;
-	}
-	return 0;
+	return ((I2C_SR2(i2c) & I2C_SR2_BUSY) != 0);
 }
 
+/*---------------------------------------------------------------------------*/
+/** @brief I2C check start condition generated
+
+@param[in] i2c Unsigned int32. I2C register base address @ref i2c_reg_base.
+*/
 uint8_t i2c_start_generated(uint32_t i2c)
 {
-	if ((I2C_SR1(i2c) & I2C_SR1_SB) != 0) {
-		(void)I2C_SR1(i2c);
-		return 1;
-	}
-	return 0;
+	return ((I2C_SR1(i2c) & I2C_SR1_SB) != 0);
 }
 
+/*---------------------------------------------------------------------------*/
+/** @brief I2C check slave address sent by master
+
+@param[in] i2c Unsigned int32. I2C register base address @ref i2c_reg_base.
+*/
 uint8_t i2c_address_sent(uint32_t i2c)
 {
 	if ((I2C_SR1(i2c) & I2C_SR1_ADDR) != 0) {
-		(void)I2C_SR1(i2c);
-		(void)I2C_SR2(i2c);
+		//(void)I2C_SR2(i2c);
 		return 1;
 	}
 	return 0;
 }
 
+/*---------------------------------------------------------------------------*/
+/** @brief I2C check our address received (in slave mode)
+
+@param[in] i2c Unsigned int32. I2C register base address @ref i2c_reg_base.
+*/
 uint8_t i2c_address_received(uint32_t i2c)
 {
 	return i2c_address_sent(i2c);
 }
 
+/*---------------------------------------------------------------------------*/
+/** @brief I2C check is byte transfer finished (used in transmitter mode)
+
+@param[in] i2c Unsigned int32. I2C register base address @ref i2c_reg_base.
+*/
 uint8_t i2c_byte_transfer_finished(uint32_t i2c)
 {
-	if ((I2C_SR1(i2c) & I2C_SR1_BTF) != 0) {
-		return 1;
-	}
-	return 0;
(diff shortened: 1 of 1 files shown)


ChuckM:
This isn't going to work as you might hope it will. The interesting case is the 'read 1 byte' which involves sending the 'start', 'addr', reading the byte, and then 'stop'. If you look at the description on the F4 in the reference manual (page 831 in Rev 7 of RM0090) under 'Master Receiver' section, you will see things like this:
```
3. In case  single byte has to be received, the Acknowledge disable is made during EV6
   **(before ADDR flag is cleared)** and the STOP condition is made after EV6.
```

Basically you end up having to do this in-line or your 'address sent' code will reset the address too soon and the NAK won't go out in the right place and poof your slave will hang thinking it needs to send another byte when the master has gone away.

KivApple:
You mean that I still need to add the implementation of the set and clear of the ACK bit register I2C_CR1?

KivApple:
I have tested my functions on STM32F4-DISCOVERY board.
I2C initialization code:

rcc_periph_clock_enable(RCC_GPIOB);
gpio_mode_setup(GPIOB, GPIO_MODE_AF, GPIO_PUPD_NONE, GPIO6 | GPIO9);
gpio_set_output_options(GPIOB, GPIO_OTYPE_OD, GPIO_OSPEED_2MHZ, GPIO6 | GPIO9);
gpio_set_af(GPIOB, GPIO_AF4, GPIO6 | GPIO9);
rcc_periph_clock_enable(RCC_I2C1);
i2c_reset(I2C1);
i2c_peripheral_disable(I2C1);
i2c_set_standard_mode(I2C1);
i2c_set_clock_frequency(I2C1, I2C_CR2_FREQ_36MHZ);
i2c_set_ccr(I2C1, 360);
i2c_set_trise(I2C1, 36);
i2c_peripheral_enable(I2C1);

I2C read device register function:
uint8_t i2c_read_byte_reg(uint32_t i2c, uint8_t addr, uint8_t reg) {
	while (i2c_busy(i2c));
	i2c_send_start(i2c);
	while (i2c_start_generated(i2c) == 0);
	i2c_send_7bit_address(i2c, addr, I2C_WRITE);
	while (!i2c_address_sent(i2c)) {
		if (i2c_nack_received(i2c)) {
			i2c_send_stop(i2c);
			return 0;
		}
	}
	i2c_send_data(i2c, reg);
	while (!i2c_byte_transfer_finished(i2c)) {
		if (i2c_nack_received(i2c)) {
			i2c_send_stop(i2c);
			return 0;
		}
	}
	i2c_send_start(i2c);
	while (i2c_start_generated(i2c) == 0);
	i2c_send_7bit_address(i2c, addr, I2C_READ);
	i2c_nack_next(i2c);
	while (!(i2c_address_sent(i2c) || i2c_nack_received(i2c)));
	while (!i2c_data_received(i2c));
	uint8_t value = i2c_get_data(i2c);
	i2c_send_stop(i2c);
	return value;
}

And now I able to read ID register of onboard CS43L22 (of coarse after switch PD4 to high level) by code i2c_read_byte_reg(I2C1, 0x4A, 0x01).

It would not be possible without the added functions in my commit.

ChuckM:
You misunderstood my comment I think, I did not mean to imply that your code would work with some peripherals, perhaps even most peripherals, I merely state that having the manipulation of the ACK/NAK/START/STOP/ADDR states as separate functions violates the design rules set out in the processor's manual for 1 and 2 byte transfers. (also if you put back quotes around your comment github will format it as code which makes it easier to read)

KivApple:
What should I do?

lib/stm32/common/i2c_common_all.c
```diff
@@ -458,4 +458,59 @@ void i2c_clear_dma_last_transfer(uint32_t i2c)
 	I2C_CR2(i2c) &= ~I2C_CR2_LAST;
 }
 
+uint8_t i2c_busy(uint32_t i2c)
+{
+	if ((I2C_SR2(i2c) & I2C_SR2_BUSY) != 0) {
+		return 1;
+	}
+	return 0;
```
karlp: you can just return the comparison here

lib/stm32/common/i2c_common_all.c
```diff
@@ -458,4 +458,59 @@ void i2c_clear_dma_last_transfer(uint32_t i2c)
...
+}
+
+uint8_t i2c_address_sent(uint32_t i2c)
+{
+	if ((I2C_SR1(i2c) & I2C_SR1_ADDR) != 0) {
+		(void)I2C_SR1(i2c);
+		(void)I2C_SR2(i2c);
```
karlp: what on earth are you doing here?

karlp:
As general guidelines, this is missing doxygen for added functions, and you can just return comparisons, no need to make if/else branches for them.

I've got some code that runs on the f4 already that can read the onboard codec, and while i2c is gross and chunky and verbose, I don't recall having to add any of these functions to get anything to work at all.   Could you clarify a bit which of these were causing you so much pain?

karlp:
fwiw, there's my f4 code, https://github.com/karlp/libopencm3-examples/tree/rough-i2c-f4l1/examples/stm32/f4/stm32f4-discovery/i2c-cs43l22

You're teh second person in a week to ask for this, so I should really clear up that branch (It was tryign to get i2c examples for f1, f4 and l1 boards working)

karlp:
so yeah, I see where you wanted some of the functions you added there now.  Still, they need doxygen, an example for the -examples repo that uses this.

KivApple:
I changed the functions, as you said, and also added a lines of the documentation (I hope I did it right).

karlp:
I believe this is obsoleted by the higher level helpers added in b40c72828df0aab74224df6fd18205ef9623a084

See https://github.com/karlp/libopencm3-tests/tree/master/tests/i2c-master for some usage.

PR #485, opened 2015-07-26
Pass usbd reference to {sof, reset, resume, suspend} callback.

kuldeepdhaka:
without this, application code has to mantain a global reference in its code.
this pr have little similarity with #482. (but are completly independent)


karlp:
:+1: This seems like a generally good idea.  I can't think of any real objections to doing this, other than that it breaks compatibility with old code.  (And that's fine, we just try to keep it to a minimum)

It shouldn't stop this commit, but we've got to sort out the usb documentation at some point too....

PR #486, opened 2015-07-26
USBD_PM_TOP preprocessor for configuring pm_top instead of hardcoding…

kuldeepdhaka:
… everywhere

### DIFF
--- a/lib/usb/usb_f103.c
+++ b/lib/usb/usb_f103.c
@@ -45,6 +45,14 @@ static void stm32f103_poll(usbd_device *usbd_dev);
 static uint8_t force_nak[8];
 static struct _usbd_device usbd_dev;
 
+/* Offset, within the dedicated packet buffer memory, of the first endpoint buffer(s).
+ * 0x40 (64) allows the full 8 EP buffer descriptors (8 bytes each) to be placed
+ * before the buffers themselves;
+ * Applications using fewer buffer descriptors could alter PM_TOP and therefore
+ * gain more usable buffer space.
+ */
+#define USBD_PM_TOP 0x40
+
 const struct _usbd_driver stm32f103_usb_driver = {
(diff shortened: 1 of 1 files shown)


karlp:
Not sure I see much need for this, I'd rather see an explanation of where the 0x40 comes from, rather than just hiding it in the top of the file behind a define.

kuldeepdhaka:
@karlp 0x40 is 64. the usb buffer first 64bytes are used for storing endpoint informations.
this can vary as per requirement. but to use all endpoints 64 byte of buffer need to be allocated for ep information (rx len, tx len, buf offset etc configuration). see datasheet for exact registers.

lib/usb/usb_f103.c
```diff
@@ -45,6 +45,8 @@ static void stm32f103_poll(usbd_device *usbd_dev);
 static uint8_t force_nak[8];
 static struct _usbd_device usbd_dev;
 
+#define USBD_PM_TOP 0x40
```
fenugrec: Possible description:
```
/* Offset, within the dedicated packet buffer memory, of the first endpoint buffer(s). 
 * 0x40 (64) allows the full 8 EP buffer descriptors (8 bytes each) to be placed
 * before the buffers themselves;
 * Applications using fewer buffer descriptors could alter PM_TOP and therefore
 * gain more usable buffer space.
 */
```

kuldeepdhaka: good idea @fenugrec

kuldeepdhaka: "Applications using fewer buffer descriptors could alter PM_TOP" this was the original motive.

kuldeepdhaka:
@fenugrec added comment.

karlp:
this has landed via e121243

Issue #360, opened 2014-11-13
USB contineous callback problem


kuldeepdhaka:
if the usb callback didnt read the buffer [in the case if its ignorning the data OR isochronous endpoint connected to DMA) then,
https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_f103.c#L322 (effect originally seen in f0x72 usb code)
will not be cleared and continuously callback in the next poll because the callback didnt readed the data.

incase callback want  to safely ignore the data, it can clear the flag itself.

---
even if this is not an issue, might help other in future while debugging the same problem.

kuldeepdhaka:
<kuldeepdhaka> rkreis, if you get a callback from library when data is received on ep, as of current, you _have_ to read the endpoint. if you do not read, it will repeat till you do not read the data
.....
<zyp> kuldeepdhaka, that sounds dumb, why not just drop it if it's not read in time?
<zyp> I think the latter is what the dwc_otg driver does, and it's dumb to have different behavior between different lowlevel drivers
......
<zyp> repeating the callback is rather useless when it's not documented and consist behavior, and just has a potential for leading to a deadlock
...
<zyp> on dwc_otg, you absolutely have to read out the data in the callback, since it's a shared rx fifo, otherwise it will not be able to receive anything more on any endpoints
....
<zyp> so it's _poll() function will actually discard remaining parts of the packet if the callback doesn't read out everything

JonasNorling:
I see the same problem on an STM32F103 during enumeration with my Macbook, with latest libopencm3 master (7b29caed). Something happens on endpoint 0 that isn't handled in the switch in _usbd_control_out(), and the default case there does a stall and doesn't read the data. The USB implementation in my application is interrupt driven, so it just sits there servicing the same interrupt over and over again forever.

kuldeepdhaka's commit e243ebb7 fixes the problem.

kuldeepdhaka:
@JonasNorling thank you for your feeedback, and good to see that you solved the problem.
can you please try out this branch [usb/f103-lock-up-due-to-control-race](https://github.com/kuldeepdhaka/libopencm3/tree/usb/f103-lock-up-due-to-control-race) for the above described problem?
there is a slim chance that the problem could related could be related to #447.

JonasNorling:
@kuldeepdhaka Yes, that commit (1c2500d8) certainly appears to help as well. I can't reproduce any lockups during enumeration on that branch, but then it doesn't happen reliably on master either.

Thank you!

[...]

PR #487, opened 2015-07-26
Fix infinite loop when OUT endpoint receive data is not read by callback

kuldeepdhaka:
…lback.

fixes issue #360


karlp:
I have memories of this being discussed on irc, did we think it was "user error" or were we going to try and find a way to defer reading or what?  IIRC, if you don't read every packet, you can end up not seeing packets you do care about, due to shared/separate ep fifos?

Issue #448, opened 2015-05-23
Endpoint callback reference


kuldeepdhaka:
if null callback is passed, reference to older callback is maintained.
https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_f103.c#L125
https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_f103.c#L137

karlp:
I've pulled this locally, but i've allowed clearing the endpoint 0 callbacks, if you're going to allow writing them to null, we might as well allow it for all of them.  (We already allowed _overwriting_ the endpoint0 handlers, so it's more consistent anyway)

Issue #449, opened 2015-05-23
Control Endpoint callbacks


kuldeepdhaka:
issue: we should clear the EP_DIR bit since control endpoint do not have direction. (its bidirectional) and set the internal control ep callback functions (that means throw the user passed callback).
https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_f103.c#L123

kuldeepdhaka:
moved common-issue1 to another issue thread.

kuldeepdhaka:
since common-issue1 is not going to be implemented, implementing this is not required anymore since, we are assuming that their is only one control endpoint and we are not going to mess with its callback (only once set in init).

PR #488, opened 2015-07-26
Proper reference to callback.

kuldeepdhaka:
even if the callback is NULL, set it.
note: not modifying EP0 callback
fixes issue #448 and #449


karlp:
I don't understand the commit comment here. Even after reading the linked issues.  For commits, the commit comment itself has to have the relevant information, not just links to offsite.  I don't see how the code matches this comment here.

kuldeepdhaka:
if callback is NULL, endpoint callback array is not updated. (previous reference is used)
if ep0 is being configured, endpoint callback array is update. (not required, as it is done in other parts)
removed check for callback NULL, and added check to only update if the ep is not ep0.
though the change look small, but fixes two bugs

kuldeepdhaka:
closed in favour of #508 (issue #448 and #449 are automatically being fixed)

PR #492, opened 2015-07-27
STM32 timers : avoid R-M-W operation when clearing interrupt flags in TIM_SR

### LINKED ISSUE #392: stm32 timer_clear_flag() race condition
The following code in lib/stm32/common/timer_common_all.c:

```
void timer_clear_flag(uint32_t timer_peripheral, uint32_t flag)
{
        TIM_SR(timer_peripheral) &= ~flag;
}
```

introduces a race condition when clearing interrupt status flags of a timer.
If another interrupt flag asserts in between the read / modify / write, then
it will be lost.  This is because bits in this register are defined by the (STM32F407)
datasheet as rc_w0, meaning software can write a 1 with no effect, or write a
0 to clear a particular status.  The recommended fix is simple:

```
void timer_clear_flag(uint32_t timer_peripheral, uint32_t flag)
{
        TIM_SR(timer_peripheral) = ~flag;
}
```

There is no need to do the read / modify.  All bits in the register are rc_w0,
aside from reserved bits.  It is presumable that any future bits ST assigns
to this register would also be rc_w0.

fenugrec:
Fixes #392 : previous code did a read-modify-write operation on the TIMx_SR register, which could result in missed interrupts.
This code now only writes 0 to the specified flag, and 1 to the other bits.
The paranoid version of this would write 0 to the reserved bits (0 is the "reset value"), but this would require knowing which flags are valid on the actual platform, and adding the corresponding macros.

### DIFF
--- a/lib/stm32/common/timer_common_all.c
+++ b/lib/stm32/common/timer_common_all.c
@@ -280,7 +280,12 @@ tim_reg_base
 
 void timer_clear_flag(uint32_t timer_peripheral, uint32_t flag)
 {
-	TIM_SR(timer_peripheral) &= ~flag;
+	/* Avoid R-M-W operation !*/
+	TIM_SR(timer_peripheral) = ~flag;
+	/* Paranoid version : set reserved bits to "Reset value" (0)
+	 * by masking out undefined flags :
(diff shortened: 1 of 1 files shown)


karlp:
merged via rebase in 410f2dd thanks for this!

PR #498, opened 2015-07-29
usb: control complete callback has no return value

karlp:
It's never used, so let's make it void to avoid any confusion.


daniel-thompson:
I've just closed #544 in favour of this one. Note that the review comments in #544 included a comment from the author of commit 4b89272 confirming the original switch from void to int was unintentional. This should be good to pull.

karlp:
you're right, not really any reason to sit on this any longer.  merged in 7814d06

PR #483, opened 2015-07-26
Rename user endpoint ctr to user endpoint callback

kuldeepdhaka:
this commit rename "user_callback_ctr" to "user_endpoint_callback"
"user_callback_ctr" name is completly wrong and add only confusion to reader.
"user_callback_ctr" is the array which hold callback for endpoints (provided by user).
"user_endpoint_callback" is far far better name for communicating the use of array.
please merge this commit as it is important as well as required by newer code.

~~prereq: #482~~

PR #501, opened 2015-07-30
Use type defined usbd_endpoint_callback callback

kuldeepdhaka:
Use type defined usbd_endpoint_callback callback for callback instead of
 void (*callback)(usbd_device *usbd_dev, uint8_t ep) in source

preq: #483


karlp:
I can just take the second one right now can't I?

kuldeepdhaka:
closed in favour of #508

PR #502, opened 2015-07-31
stm32f1: remove artificial limit on gpio remap mask

karlp:
Connectivity line devices have more remaps available, and the existing code was
artificially and needlessly preventing those remaps from being set, even if
defined.

While implementing this, the existing code to handle SWJ remap was found to be
inconsistent with the reference manual and has been fixed.

Fixes github issue #369

Note: for archaeology, this supersedes pull request #394 

Please can someone sanity check that this is the "really" correct fix for this issue.  (A bug will be filed separately to indicate that you can't presently _clear_ remaps)

### DIFF
--- a/lib/stm32/f1/gpio.c
+++ b/lib/stm32/f1/gpio.c
@@ -161,14 +161,20 @@ this function as its current value cannot be ascertained from the hardware.
 
 @param[in] swjdisable Unsigned int8. Disable parts of the SWJ capability @ref
 afio_swj_disable.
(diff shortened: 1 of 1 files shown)


karlp:
I actually propose replacing this API with something like, remap(swjdisable, maps_set, maps_clear) or something, and just handling it all in one function.

karlp:
merged via rebase in eebef01 to at least make it possible to use.  Making a new api for set/clear the remap is still valid.  See https://github.com/libopencm3/libopencm3/issues/503

PR #505, opened 2015-07-31
Send ZLP if app send less data than host expecting and send length is multiple of bMaxPacketSize0

kuldeepdhaka:
[USB] If the application
 - send smaller length data than the host requested and
 - the send data is multiple of bMaxPacketSize0.
If the condition occur, a ZLP (Zero Length Packet) need to be send to the host to tell that
 the device is out of data and will not further send anymore data.

This fix only apply to Control IN Request

(if the second condition do not occur, a packet of length less than bMaxPacketSize0 is send to host. using that host can automatically infer that device is out of data)

use the firmware and python script to test the patch provided
https://github.com/kuldeepdhaka/libopencm3-examples/tree/usb-stack-tester

reference: #194

### DIFF
--- a/lib/usb/usb_control.c
+++ b/lib/usb/usb_control.c
@@ -89,7 +89,10 @@ static void usb_control_send_chunk(usbd_device *usbd_dev)
 		usbd_ep_write_packet(usbd_dev, 0,
 				     usbd_dev->control_state.ctrl_buf,
 				     usbd_dev->control_state.ctrl_len);
-		usbd_dev->control_state.state = LAST_DATA_IN;
+
+		usbd_dev->control_state.state =
+			usbd_dev->control_state.send_zlp_at_end ? DATA_IN : LAST_DATA_IN;
+		usbd_dev->control_state.send_zlp_at_end = false;
 		usbd_dev->control_state.ctrl_len = 0;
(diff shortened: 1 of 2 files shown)


kuldeepdhaka:
i originally got introduced to the bug from #194, but the patch provided by #194 (according to my findings) will break the usb stack. so, if this pr is merged, #194 can also be closed.

lib/usb/usb_control.c
```diff
@@ -71,6 +71,8 @@ int usbd_register_control_callback(usbd_device *usbd_dev, uint8_t type,
 	return -1;
 }
 
+int send_zlp_at_end = 0;
```
plaes: This variable seems to be unused?

kuldeepdhaka: my fault

karlp:
This has some minor stylecheck issues, and it should probably have a usb: tag in the commit header.  I think it's worth mentioning that this only applies to control requests too.

kuldeepdhaka:
@karlp done!

karlp:
The tests in this pull are great, and expose a proper bug in the control request handling.  need to pull this up into the unit tests and merge a fix.

karlp:
Fixed via merge of 3ed12b6fd944c4b3c921b8f73a6912f856edb68a

PR #506, opened 2015-07-31
Move usb_lm4f.c from lm4f/ to usb/

kuldeepdhaka:
Management of usb code is easier if everything is at one place


kuldeepdhaka:
i purposed pr because while doing development, i forgot about usb_lm4f.c for a while.
also this is a exception since all usb_* files are in usb/

karlp:
Merged via rebase after cleaning up the commit comment.  see e7eb61f

PR #479, opened 2015-07-24
STM32F0 USB support:

fenugrec:
[EDIT - Current version could be obsoleted by other pending PRs. For reference only. ]
- Reworked headers and common code
- Renamed USB_DADDR_ENABLE flag to USB_DADDR_EF  (already merged)
- Fixed typo : USB_BDCR* -> USB_BCDR* (already merged)
- Added USB_FNR register defines (already merged)
- Moved USB_PMA_BASE address definition to family-specific memorymap.h (already merged)
- Changed hard-coded pma-top definition to USBD_PM_TOP macro (split PR)
Based on the works of Frantisek Burian, Kuldeep Singh Dhaka and Robin
Kreis.

PR #509, opened 2015-08-01
[USB] This commit commonize the USB full speed device.

kuldeepdhaka:
Earlier, everything was kept in include/libopencm3/stm32/usb.h
 now, core common content has been moved to include/libopencm3/stm32/common/usb_common_all.h
and include/libopencm3/stm32/{f0, f1, f3, l0, l1}/usb.h is provided to include the above on family basis.
and a top include/libopencm3/stm32/usb.h has been added to route to the intended family.

---
this commit is inspire from pr #282 (but the commit 235095be191c57f82111f637c48e5b1849a2ae3a has been identified to contain modified content which is know to cause problem (in #439, see @fenugrec comments)

### DIFF
--- a/include/libopencm3/stm32/common/usb_common_all.h
+++ b/include/libopencm3/stm32/common/usb_common_all.h
@@ -1,0 +1,324 @@
+/** @defgroup usb_defines USB Defines
+
+@brief <b>Defined Constants and Types for the STM32F1xx, STM32F0xx, STM32L0xx_defines USB Module</b>
+
+@ingroup STM32F1xx_defines STM32F0xx_defines STM32L0xx_defines
+
+@version 1.0.0
+
+@author @htmlonly &copy; @endhtmlonly 2009
+Piotr Esden-Tempski <piotr@esden.net>
+
+@date 11 March 2013
+
+LGPL License Terms @ref lgpl_license
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2009 Piotr Esden-Tempski <piotr@esden.net>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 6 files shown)


fenugrec:
I structured the header inclusion guards a bit differently in PR #479 ; I think it would be better this way:

```
**** stm32/usb.h ****
#ifndef LIBOPENCM3_USB_H
#define LIBOPENCM3_USB_H	
[your new content with the #include <f?/usb.h> selection]

#endif

**** f[*]/usb.h ****
#ifndef LIBOPENCM3_USB_H
#	error Do not include directly !
#else

#include <.../usb_common_all.h>

#endif

**** usb_common_all.h ****
// same mechanism : error if included directly
#ifndef LIBOPENCM3_USB_H
#	error Do not include directly !
#else

[original #defines, etc]

#endif
```

kuldeepdhaka:
```
**** stm32/usb.h ****
#if STM32F0
#   include <libopencm3/stm32/f0/usb.h>
#elif STM32F1
# ......
#else
#    error "stm32 family not defined."
#endif

**** f[*]/usb.h ****
#ifndef LIBOPENCM3_USB_H
#define LIBOPENCM3_USB_H
#    include <.../usb_common_all.h>
#endif

**** usb_common_all.h ****#
#if LIBOPENCM3_USB_H
#ifdef LIBOPENCM3_USB_COMMON_ALL_H
#define LIBOPENCM3_USB_COMMON_ALL_H
[original #defines, etc]
#else
#   warning Do not include directly !

#endif
```

this is the structure i could infer from other files. will be updating acordingly

kuldeepdhaka:
modified code as per previous comment.

fenugrec:
Your last version doesn't protect against double inclusion of stm32/usb.h ; that's why I had "#ifndef LIBOPENCM3_H \n #define LIBOPENCM3_H ... " in that file instead of duplicated in the f*/usb.h headers.

fenugrec:
Also, #warning is gcc-specific, but #error is standard C. I think double / illegal #includes should cause an error, they clearly indicate a problem !

kuldeepdhaka:
header guard is for preventing including same content twice.
since stm32/usb.h just route to specific target header, there will not be any problem. (target header is having header guard)
and #warning is used in other part of code, so i feel it is acceptable. (btw libopencm3 only support gcc, so not a problem from this prespective)

fenugrec:
Concerning #warning : fair enough. Disregard my comment.
For stm32/usb.h : it doesn't *only* route to specific target header, since you #include <...cm3/common.h> and  <...stm32/memorymap.h> before including the specific headers ! so there would be duplicate def errors within common.h and memorymap.h...

kuldeepdhaka:
@fenugrec just like stm32/usb.h,   stm32/memorymap.h is routing to stm32/f*/usb.h which is preventing double inclusion using header guard.

karlp:
Thanks for all the hardwork, and pushing to get this rolled in.  This tree, like many, provided lots of valuable input to get the final version merged

PR #510, opened 2015-08-01
[USB] Commonize the OTG {FS, HS} with proper defination.

kuldeepdhaka:
Earlier include/libopencm3/stm32/otg_{fs, hs}.h contain common content (double copy, things could go wrong easily).
now, the common content has been moved to include/libopencm3/stm32/common/otg_common_all.h
and include/libopencm3/stm32/otg_{fs, hs}.h has been modified to use include/libopencm3/stm32/common/otg_common_all.h

Also, changes have been made to lib/usb/usb_{f107, f207, fx07_common}.c to use the intended copy only.
The cause of regression in lib/usb/usb_fx07_common.c (see e11b7d5d6df88b90efa9d53ebca80cf1cb3676dd) will be avoided since, none
of the include/libopencm3/stm32/otg_{hs, fs}.h have been included.

lib/usb/usb_fx07_common.c:
earlier, REBASE(offset) was used to access registers, now the registers can be accessed using USBD_OTG_*

lib/usb/usb_{f207, f107}.c:
they are modified to use include/libopencm3/stm32/otg_{fs, hs}.h

Overall the commit is cleaning up the double defination and prevent regression in future.
@H2OBrain have a look

### DIFF
--- a/include/libopencm3/stm32/common/otg_common_all.h
+++ b/include/libopencm3/stm32/common/otg_common_all.h
@@ -1,0 +1,332 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2010 Gareth McMullin <gareth@blacksphere.co.nz>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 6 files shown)


amirhammad:
some commonalization have been done in the past by me in PR #405, which is pretty old ... I also added some usb host-related definitions that were missing.

kuldeepdhaka:
a possible downside of this pr with comparision to #405 is that, 
inside lib/usb/usb_fx07_common.c lots of (USBD_OTG_*) macro's have been added [replaced REBASE()].
the older way REBASE(reg_offset) is still being followed in #405.
USBD_OTG_* would not give a superduper benifit.

one good side of this patch is that, `#define OTG_<REG>(base) MMIO32(base + reg-offset)` instead of `#define OTG_<REG> (reg-offset)`.

kuldeepdhaka:
closed in favour of #405

PR #511, opened 2015-08-01
Added USB_FNR register defines

kuldeepdhaka:
credit: @fenugrec
originally presented in #479 but the commit was a single, so had to seperate out the FNR work.

@fenugrec please review the commit to confirm that i have not missed anything.


fenugrec:
definitions are correct according to RM0091 (F0) and RM0008 (F1)

karlp:
merged via rebase in 0d1f5fc

PR #514, opened 2015-08-10
All STM32 family pwr.h must use LIBOPENCM3_PWR_H as include guard so …

ksarkies:
…that

pwr_common.h can detect that it has been referenced by pwr.h for
each family. F2 and F3 had the wrong include guard.


karlp:
merged via rebase in 4d31528  thanks for fixing this so promptly!

Issue #516, opened 2015-08-13
Stack pointer is not always 8-byte aligned during ISRs


forrestv:
Alignment of the stack pointer along 8-byte boundaries is required by the ARM ABI during function calls[1]. As was determined from a conversation on #libopencm3, this can be violated for interrupt handlers when the STKALIGN bit in SCB_CCR isn't set. It is set by default for Cortex revision r2p0, but not r1p1.

As the libopencm3 examples don't set STKALIGN manually or use \_\_attribute\_\_((interrupt)) [2] on ISRs, users that follow the examples will find that ISRs that use varargs (potentially among other things) will mysteriously fail occasionally, when the code running before the interrupt happened had the stack not 8-byte aligned (as is allowed).

References:
1. Last section of ARM IHI 0046B http://infocenter.arm.com/help/topic/com.arm.doc.ihi0046b/IHI0046B_ABI_Advisory_1.pdf
2. Last line of the interrupt section of https://gcc.gnu.org/onlinedocs/gcc/ARM-Function-Attributes.html

forrestv:
See pull request #517 for fix.

karlp:
closed via 4a21730

PR #517, opened 2015-08-13
Set STKALIGN in SCB_CCR on reset. Fixes issue #516 (Stack pointer is not always 8-byte aligned during ISRs).

forrestv:


karlp:
merged in 4a21730  (I expanded the commit comment a lot, I prefer to keep that sort of diagnosis in the commit message, where it lives in the git repo, rather than just being in the github issue, where it might one day disappear)

Thanks a lot for persevering with this, and convincing us of the real issue.

PR #522, opened 2015-08-23
stm32F3: fix memory map for ADC2_BASE and ADC4_BASE.

ventosus:


karlp:
merged via rebase in be56146  Thanks for this.

PR #523, opened 2015-08-28
Add nonblocking calls to usart

kungjohan:
Add nonblocking calls to check the status of the usart.
Instead of busy waiting you can do something else waiting to transmit next byte or waiting to receive data.

### DIFF
--- a/lib/stm32/common/usart_common_f124.c
+++ b/lib/stm32/common/usart_common_f124.c
@@ -77,7 +77,7 @@ uint16_t usart_recv(uint32_t usart)
 
 bool usart_is_send_ready(uint32_t usart)
 {
-	return ((USART_SR(usart) & USART_SR_TXE) == 0);
+	return ((USART_SR(usart) & USART_SR_TXE) == 1);
 }
 
 
@@ -109,7 +109,7 @@ usart_reg_base
 
 bool usart_is_recv_ready(uint32_t usart)
 {
-	return ((USART_SR(usart) & USART_SR_RXNE) == 0);
+	return ((USART_SR(usart) & USART_SR_RXNE) == 1);
 }
 
 
--- a/lib/stm32/common/usart_common_f124.c
+++ b/lib/stm32/common/usart_common_f124.c
@@ -99,6 +99,22 @@ void usart_wait_send_ready(uint32_t usart)
 }
 
 /*---------------------------------------------------------------------------*/
+/** @brief USART check if Received Data Available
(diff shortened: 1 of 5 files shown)


lib/stm32/common/usart_common_f124.c
```diff
@@ -65,6 +65,24 @@ uint16_t usart_recv(uint32_t usart)
...
+ * the next data word.
+ *
+ * @param[in] usart unsigned 32 bit. USART block register address base @ref
+ * usart_reg_base
+ * @returns boolean: transmit data buffer is ready to accept the next data word
+ */
+
+bool usart_is_send_ready(uint32_t usart)
+{
+	/* Wait until the data has been transferred into the shift register. */
```
karlp: copy/paste error?

lib/stm32/f0/usart.c
```diff
@@ -191,6 +191,23 @@ uint8_t usart_recv(uint32_t usart)
...
+ * the next data word.
+ *
+ * @param[in] usart unsigned 32 bit. USART block register address base @ref
+ * usart_reg_base
+ * @returns boolean: transmit data buffer is ready to accept the next data word
+ */
+
+bool usart_is_send_ready(uint32_t usart)
+{
+	/* Wait until the data has been transferred into the shift register. */
```
karlp: Same here

karlp:
Seems generally reasonable.  I have no objections to this.

karlp:
Oh, Please at least update the f3 file too, to at least keep the stm32 line consistent.  "soon" the f0, f3 and l0 peripherals will all get a common core pulled out, but until then you'll need to modify them all.  (except L0 doesn't exist just yet, so you're lucky there ;)

kungjohan:
Thanks for the review, updated this.

danielinux:
@kungjohan Thank you very much for this. We merged this change to our fork back in November and we strongly believe that such calls should be present in the unified UART driver for all platforms. Ref: https://git.io/voQGk

karlp:
TODO (I will do in merge if no-one gets there first)

- squish to single commit
- modify all the wait_ blocking functions to use the new is_ asynch functions, rather than duplicating the checks

OpenUAS:
Merge appreciated...

karlp:
@OpenUAS you can see the TODO if you "want to get there first"  making less for me to do is always a good way of making it easier for things to happen.  Rebasing it to be current too.

PR #525, opened 2015-09-07
stm32f3: ld: stm32f3[01]3?c have 40K RAM only.

ventosus:
Memory sizes for stm32f3[01]3?c in ld/devices.data need to be corrected from
RAM=48K to RAM=40K.


karlp:
Are you using this file? it's still not the "main" linker script used.  If you are, could you fix up the B devices too, they're in the same datasheet.

ventosus:
>> Are you using this file? it's still not the "main" linker script used.
I was experimenting with it, as I'd like to put code into CCM on stmf3.

>> If you are, could you fix up the B devices too, they're in the same datasheet.
I've done so.

karlp:
merged via rebase in 15a6103 thanks for this!

PR #527, opened 2015-09-09
stm32f3: ld: support initialized code/data on CCM.

ventosus:
This commit adds support for initialized code/data on CCM (Core Coupled
Memory) on stm32f3 by:
- Splitting .ccm section into two sections:
  - .ccm_data for initialized code/data with subsections:
    - .ccm_text for CCM program code
    - .ccm_rodata for CCM read-only data
    - .ccm_data for CCM read-write initialized data
  - .ccm_bss for CCM uninitialized data
- Adding corresponding initialization loops in the series' pre_main()
- Adapting automatic ld script generation tools to produce correct
  scripts for the whole f3 series (as not all of them have CCM RAM)

Usage:

Code can be put into CCM and sped up by up to ~30% with:

```
__attribute__((section(".ccm_text"))) void ex_handler(void) {}
```

Variables may also be put into CCM accordingly:

```
__attribute__((section(".ccm_rodata"))) const int ex_rodata = 1;
__attribute__((section(".ccm_data"))) int ex_data = 2;
__attribute__((section(".ccm_bss"))) int ex_bss;
```

### DIFF
--- a/ld/devices.data
+++ b/ld/devices.data
@@ -110,10 +110,10 @@ stm32f2[01][57]?e* stm32f2 ROM=512K RAM=128K
 stm32f20[57]?f* stm32f2 ROM=768K RAM=128K
 stm32f2[01][57]?g* stm32f2 ROM=1024K RAM=128K
 
-stm32f302?b* stm32f3ccm ROM=128K RAM=24K CCM=8K
-stm32f302?c* stm32f3ccm ROM=256K RAM=32K CCM=8K
-stm32f303?b* stm32f3ccm ROM=128K RAM=32K CCM=8K
-stm32f3[01]3?c* stm32f3ccm ROM=256K RAM=40K CCM=8K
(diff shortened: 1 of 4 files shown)


ChuckM:
This is a bit dangerous, in that there are restrictions on CCM that are not always obvious, the one that bites people most often is the use of CCM for stuff that needs to be fetched by DMA. That said, I like the idea of making the symbols available for people.

ventosus:
>This is a bit dangerous, in that there are restrictions on CCM that are not always obvious, the one that bites people most often is the use of CCM for stuff that needs to be fetched by DMA.

True indeed. Although this is clearly stated in the manual, it may easily be overlooked.

>That said, I like the idea of making the symbols available for people.

An alternative would be to add a minimal example project in libopencm3-examples to show people how to add CCM support with libopencm3 as-is, e.g. with a custom ld-script and initialization code at starting of main()?

karlp:
I like the idea of a self contained example folder showing how to use CCM nicely.

PR #534, opened 2015-09-27
Wait for ADC to be ready after in adc_power_on(..)

onnokort:
This waits for the ADC_ISR_ADRDY flag to become one in the ADC_ISR register. This makes adc_power_on(..) closer to the power on example in the reference manual (DM00031936.pdf) on page 935.


karlp:
I've not looked closely at these, but there's definitely some copypaste that went on with f0, and things that haven't been extracted out well to common code.  I believe there are a couple of other pulls touching adc poweron/off that all need to be standardized across the stm32 line too

karlp:
This is now fixed by a89cd8645400edf32ae99bce4b883cc9e5306f66

PR #535, opened 2015-09-27
STM32F0: adc_start_conv_reg() & continuous fix

onnokort:
ADSTART with not go back to zero when using the ADC in continuous conversion mode. This is according to the description on page 259 of DM00031936.pdf.

I am expecting that the functions provided in adc.h should also work for setting the ADC up in continuous mode?

This might theoretically break existing code that relied on the old behavior, so I guess this patch needs further tweaks.


karlp:
probably won't look at this much until I've finished reworking some of the f0/f3 adc basics but valid bug, needs to be addressed.  Probably via a bunch of asynch methods

esden:
This patch definitely conflicts with the current code and needs to be rebased.

Did the issue get addressed, should we revisit this?

PR #536, opened 2015-10-02
Drafts/near final usb

karlp:
What's missing? doxygen review


karlp:
This has merged, with the tests now.

PR #538, opened 2015-10-06
make: pass EXTRA_CFLAGS variable to CC

tidklaas:
Allow user to pass additional flags to the compiler without modifications
to the Makefiles. Contents of EXTRA_CFLAGS will be passed along to the CC
in addition to the CFLAGS generated by the build system.

This is useful if libopencm3 is build as part of a bigger project or gets linked against a library built with non-default flags. An example would be the newlib-nano currently shipped by Debian testing, which was compiled with -fshort-wchar.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,7 @@ them as environment variables, for example:
     $ VARIABLE=value make
 
 * `FP_FLAGS` - Control the floating-point ABI
+
    If the Cortex-M core supports a hard float ABI, it will be compiled with
    floating-point support by default. In cases where this is not desired, the
    behavior can be specified by setting `FP_FLAGS` Currently, M4F cores default
@@ -99,6 +100,17 @@ them as environment variables, for example:
 
         $ FP_FLAGS="-mfloat-abi=soft" make               # No hardfloat
         $ FP_FLAGS="-mfloat-abi=hard -mfpu=magic" make   # New FPU we don't know of
+
+* `CFLAGS` - Add to or supersede compiler flags
+
+   If the library needs to be compiled with additional flags, they can be
+   passed to the build system via the environment variable `CFLAGS`. The
+   contents of `CFLAGS` will be placed after all flags defined by the build
+   system, giving the user a way to override any default if necessary.
+
+   Examples:
+
+        $ CFLAGS="-fshort-wchar" make    # compile lib with 2 byte wide wchar_t
 
 Example projects
 ----------------
--- a/lib/Makefile.include
+++ b/lib/Makefile.include
@@ -42,7 +42,7 @@ $(SRCLIBDIR)/$(LIBNAME).ld: $(LIBNAME).ld
 
 %.o: %.c
 	@printf "  CC      $(<F)\n"
-	$(Q)$(CC) $(CFLAGS) $(EXTRA_CFLAGS) -o $@ -c $<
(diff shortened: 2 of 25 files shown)


karlp:
Does it not work to just set CFLAGS yourself?  locm3 can be made to only add to CFLAGS, if it's not already?

tidklaas:
CFLAGS gets assigned in the target specific Makefile:
```
CFLAGS		= -Os \
		  -Wall -Wextra -Wimplicit-function-declaration \
		  -Wredundant-decls -Wmissing-prototypes -Wstrict-prototypes \
		  -Wundef -Wshadow \
		  -I../../../include -fno-common \
		  -mcpu=cortex-m3 $(FP_FLAGS) -mthumb -Wstrict-prototypes \
		  -ffunction-sections -fdata-sections -MD -DSTM32F1
CFLAGS          += $(DEBUG_FLAGS)
```
This overrides values put in the environment by the user or top level Makefile. If, on the other hand, CFLAGS is set on the make command line, the assignment in the Makefile is completely ignored. Using an extra variable removes the ambiguities of having different behaviour depending on the variable's origin (environment or command line.) 

If you want to allow adding values directly to CFLAGS, you would have to use  `CFLAGS += -Os ...` in all the targets' Makefiles. This way there is only the Makefile.include to patch.

Besides, having an explicit variable for user flags might make tracking down build problems easier, as it is easier to distinguish whether a problematic flag came from the build system or was set by the user.

karlp:
I'm somewhat inclined to follow autotools behaviour here[1], which is that CFLAGS is for the user to use.  I know that EXTRA_CFLAGS isn't uncommon either, and I agree that _a_ solution for this is "a good thing"

It's basically irrelevant how big the changes are internally, whether it's one makefile or 10, the goal is better for the end user.  

FWIW, any change for this would need to at least update the readme or something to try and help the user find this setting.  Having to guess that EXTRA_CFLAGS (which, though not uncommon, is no standard) exists and can be used, or tracking down the makefile rules manually to find it isn't good enough for a user facing functional knob.


[1] I might think autotools is an abomination, but it's a rather standard abomination.

tidklaas:
So, the right way would be to go through all target Makefiles and Makefile.include, renaming CFLAGS to something like TARGET_CFLAGS and EXTRA_CFLAGS to just CFLAGS. The availability of CFLAGS to users for adding custom flags should then be mentioned in the README.md, the way it is done for FP_FLAGS already. Anything else?

karlp:
That looks good to me, anyone else have opinions on this?  (Got a link to this debian issue? I presume there's an issue for it even if it's wont fix?)

tidklaas:
I have not checked of there is a bug filed for this, but it is mentioned in the Debian changelog: http://metadata.ftp-master.debian.org/changelogs//main/n/newlib/newlib_2.2.0+git20150830.5a3d536-1_changelog

fenugrec:
Is it not possible to replace all "CFLAGS = ..." lines with "CFLAGS += ..." ? i.e. always add to CFLAGS, whether or not it was defined outside ?

karlp:
the reason not to do that is that it means library flags come after user flags, potentially overwriting them.

karlp:
@tidklaas yeah, I found the changelog, but like all shitty moronic changelogs it says what it did, but not why it did it. :(

karlp:
merged thanks!

PR #540, opened 2015-10-07
Fix offset for GPIO_BRR on stm32f0

devanlai:
Per section 9.4.11 of the reference manual, the offset for the BRR register for F0 chips should be 0x28, not 0x24


karlp:
merged with minor summary line changes in 5270c11

PR #541, opened 2015-10-10
STM32F4 USB bugfix: RVFLVL set while rxbcnt=0

znuh:
OTG_GINTSTS_RXFLVL can be set while rxbcnt is zero.
usbd_ep_read_packet will return 0 then, triggering a stall on EP0 in usb_control.c lines 108/109. This leads to -EPIPE errors when a linux driver tries to send a control message to the device.

The STM32F4 reference manual (RM0090 rev. 10 pages 1343/1344) suggests checking the byte count after RXLVL is set. This is what the attached code does.


karlp:
Please use separate branches for unrelated pulls.  Doing it all on master means that your pull gets auto updated when you make further commits.

I've just pulled in the linker generator changes for you.

For the RVFLVL, do you have any way of reproducing this?

karlp:
If you _can_ reproduce this, adding tests in the tests directory would be _awesome_!

znuh:
Sorry, forgot about the extra branch.
As for the test: yes I can reproduce this, but a test for this involves the host sending several control messages to the device. It can be done using the libusb I guess, but I don't have dedicated code for this at the moment.

karlp:
if you can outline the steps clearly enough, and what the firmware needs to do, in it's control callback, if anything, we can probably work it out.  If you have a look at the tests/gadget0 directory, you can see the sort of tests I've setup for a  bunch of other cases.

PR #542, opened 2015-10-13
Fix 2 comments

Sh4rK:


karlp:
Thanks, merged (after cleaning up the commit comment)  the whole f1 rcc code is just gross and prone to this sort of thing.

Sh4rK:
Thanks, I will pay attention next time :)

PR #543, opened 2015-10-13
Improved comments of USB API:

fenugrec:
usbd.h: more comments for bmRequestType and bmAttributes bitfield macros
usbd.h: migrated and added API doxyblocks from implementations (*.c)


karlp:
merged with commit comment cleanups in c899273

PR #544, opened 2015-10-13
usb: Fix definition of usbd_control_complete_callback

daniel-thompson:
Commit 4b892724cf74 ("usb: provide typedefs for all the function
callbacks.") appears to have acidentally changed the return type of
the control complete callback (from void to int) resulting in spurious
warnings building USB applications.

Restore the original type.


fenugrec:
Indeed, but there's already https://github.com/libopencm3/libopencm3/pull/498 that is identical :)

kuldeepdhaka:
@daniel-thompson Indeed, it has.

daniel-thompson:
Ok. I'll close this pull request.

Would be great to see #498 merged soon though. I spent quite a bit of time chasing through the source trying to find out what the return value was for (so I could fix my firmwares after updating libopencm3).

PR #497, opened 2015-07-29
usb: Use enumerated return codes

karlp:
The enum usbd_request_return_codes has been available for some time.  It
should be used internally, not just by users of this code.

Note: this is a superset of #481 (it completely replaces that pull request, by expanding it to not just use the return defines, but also fixes the signatures of those methods.

PR #550, opened 2015-10-18
More USB documentation:

fenugrec:
- Removed old "FIXME: Have class/vendor requests as well" comment  from usb_standard.c, since those requests are already handled by the usbd_register_control_callback() mechanism;
- Added Doxyblocks for callback signatures;
- Replaced some hardcoded request return codes to facilitate eventual
migration to "enum usbd_request_return_codes" return type. This duplicates some of the changes of https://github.com/libopencm3/libopencm3/pull/497 ;

- NOTE : I may have misinterpreted the use of "usbd_endpoint_callback" (see comment in diff)
- NOTE : The doxyblock for usbd_control_callback should be double-checked too.


include/libopencm3/usb/usbd.h
```diff
@@ -102,18 +102,40 @@ extern void usbd_register_resume_callback(usbd_device *usbd_dev,
...
+/** Callback function pointer for standard Set Interface (alternate setting) requests
+ * @param wIndex Interface number
+ * @param wValue Alternate setting
+ */
 typedef void (*usbd_set_altsetting_callback)(usbd_device *usbd_dev,
 		uint16_t wIndex, uint16_t wValue);
 
+/** Callback function pointer for endpoint CTR event (?)
+ *
+ * Called when an OUT, IN, or SETUP token is received (?)
```
fenugrec: I wasn't too sure about this. It seems up to each driver to decide when to call an usbd_endpoint_callback; they all appear to do so from within their _poll() function according to the type of token received. A second opinion would be required to validate this, though.

esden:
This would be great to add to the code. This is some useful documentation.

I am keeping this open as I would love if someone can pick this up, rebase on the current code so that I can merge it.

PR #552, opened 2015-10-22
Add inital RCC subsystem support for the STM32F7.

Sync-:
Just uses the default reset clock. Actual PLL values coming later.

### DIFF
--- a/include/libopencm3/stm32/f7/flash.h
+++ b/include/libopencm3/stm32/f7/flash.h
@@ -1,4 +1,15 @@
-/* This provides unification of code over STM32F subfamilies */
+/** @defgroup flash_defines FLASH Defines
+ *
+ * @ingroup STM32F7xx_defines
+ *
+ * @brief Defined Constants and Types for the STM32F7xx FLASH Memory
+ *
+ * @version 1.0.0
+ *
+ * @date 14 January 2014
+ *
+ * LGPL License Terms @ref lgpl_license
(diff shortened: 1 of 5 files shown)


karlp:
@Sync- this branch has gotten a  bunch of ethernet commits in it to that are out of scope for this pull, can you leave a different branch with just the f7 rcc pull bits please?

Sync-:
Did so.

karlp:
I want to merge this, and will do so, but do people have any opinions on which API they prefer?

1. uint32_t rcc_system_clock_source(void) and void rcc_set_sysclk_source(uint32_t clk)  (uses bit definitions) (used on f1, f3,f4,l4 and here, f7)
2. enum rcc_osc rcc_system_clock_source(void) and void rcc_set_sysclk_source(enum rcc_osc osc) (uses enums, used on f0, l0)

I kinda feel that API 2 is "nicer" for the user, though it's "fatter" library code.  Without comments, I guess I should just file this as an API issue?

karlp:
I've pulled this in after modifying all the namespace prefixes as was done for other families after this PR was submitted.  There's definitely room for some future extraction of   common api code for RCC stuff, but   now's not the time for that.  Same for the blocking/non-blocking. It's just not the time for that in this commit.

karlp:
I've pulled in (most?all?) of this in 3fc0c9d..d2540e5

PR #553, opened 2015-10-22
Add linker scripts for L100RC and F407x6

joostrijneveld:


karlp:
I've merged the l100, with a name change, and the f4 I just fixed the existing file (somewhat)  It still needs the right name, but the docs imply it should work for all stm32f4xxG parts.

We really need to move everyone on to bufran's generated ldscripts branch!

PR #559, opened 2015-11-03
Use ?= for TARGETS in Makefile

MightyPork:
Allows to limit build to only what you really want by using eg. `make TARGETS=stm32/f3`.

Also corrected some indentation (tabs where spaces should have been).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,11 +34,11 @@ space:=
 space+=
 SRCLIBDIR:= $(subst $(space),\$(space),$(realpath lib))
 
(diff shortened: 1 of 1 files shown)


flixr:
Um... this already works without the `?=`:
```
make lib TARGETS="stm32/f3"
```

have always only been building targets I actually need.

MightyPork:
@flixr hmm.. that's odd.

Seems `?=` will allow env variables tho, might be handy if someone makes a build script or something.

MightyPork:
okay I guess this is not a useful PR then

karlp:
Probably related to the expansion of := vs = in makefiles.  fyi, if you want to make a pull request for something like the targets change, don't put other unrelated changes in the same branch. (like the indentation changes you added)

Issue #548, opened 2015-10-17
F3 ADC definitions renamed


karlp:
I'm planning on renaming a pile of definitions for the F3 ADC peripheral (F30x style) to use naming more in line with f0/l0/l4, that share the same peripheral.  It's not much, just 
```
- ADC_CFGR_XXX
+ ADC_CFGR1_XXX
```

This means we can use the same code for both of them.

karlp:
After conversations, I'm going to attempt to add the ADC_CFGR defintions for f3 only, as aliases for ADC_CFGR1

esden:
@karlp this patch looks good to me. Should we make the "old macros" generate compiler warnings that the CFGR style is legacy and might be removed in a future release?

esden:
Ahh I see, this is part of a series, that is a draft pull request #555

karlp:
this tree has landed

PR #562, opened 2015-11-05
STM32 ADC standardization and cleanup

karlp:
~~DEPENDS ON https://github.com/libopencm3/libopencm3/pull/561~~ merged

This is further work cleaning up some of the common functionality across the ADCs, paving the way for upcoming patches that reduce duplication and bring in L0 and L4 support, and merge the f0/f3 adc code.

### DIFF
--- a/include/libopencm3/stm32/common/adc_common_v1.h
+++ b/include/libopencm3/stm32/common/adc_common_v1.h
@@ -357,7 +357,7 @@ specific memorymap.h header before including this header file.*/
 BEGIN_DECLS
 
 void adc_power_on(uint32_t adc);
-void adc_off(uint32_t adc);
+void adc_power_off(uint32_t adc);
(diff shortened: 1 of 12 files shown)


karlp:
@esden This one is "ready to go" I was just leaving it in my local tree while I worked on the rest of the "v2 adc" peripheral stuff for f0/f3/l0/l4 but it's just stalled out in my christmas overload.  All these parts are good to go though.  (#548 had some debate so I held it off for a bit longer, and it's in the adv c2 drafts branch instead of here)

ChuckM:
:+1: looks good to me

karlp:
(Also, it was waiting in the hopes someone else would see what I was doing and yea/nei it ;)

karlp:
merged now, or equivalents at least.

PR #563, opened 2015-11-08
RCC namespace cleanups [BREAKING]

karlp:
The "stage 1" of @esden's overhaul of RCC naming. (See #410)  This all seems like a good idea, so I've finished it off for the rest of the stm32 series.

Yes, this is breaking for a lot of downstream users, but we're doing a lot of breakage right now already to get things more cleaned up, so now's a great time to get this in.  This got nothing but generally positive reviews months ago, so I'll probably run with this as the base for my adc v2 work, to give it some testing.

### DIFF
--- a/include/libopencm3/stm32/common/pwr_common_l01.h
+++ b/include/libopencm3/stm32/common/pwr_common_l01.h
@@ -77,15 +77,15 @@
 
 /* --- Function prototypes ------------------------------------------------- */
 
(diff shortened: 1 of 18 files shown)


brabo:
:+1:

esden:
Awesome! Thanks! Merged with 3a7cbec776bc9b9b53eefc26d5bf365466977e5a

PR #566, opened 2015-11-14
make: Use gcc-ar instead of ar to fix LTO build of the library

Sh4rK:
This PR allows to compile the library with `CFLAGS="-flto"`, which in my (little, toy) case provides a ~30% reduction in code size (2600 -> 1796 bytes), and enables inlining of small functions (for example `gpio_clear`, `gpio_set`) which results in faster code.


karlp:
If I understand it correctly, this is just an interim fix until binutils gets a new release.  Until the project as a whole simply moves to LTO, I'm not sure I want to bring in this change.  We're kinda tied to gcc already, but we should try and make sure it stays building with clang/llvm as well.  if you are already setting CFLAGS to add the -flto, just set AR as well?

Sh4rK:
I would, but all makefiles set the AR variable unconditionally (as can be seen in the commit). Would you accept a PR that changes these assignments to conditional ones? Also, I wouldn't mind adding this, since gcc is hard-coded anyway in all files, and this is right next to that, so whenever someone wants to implement a more general way, they would notice this for sure.

karlp:
Oh absolutely, sorry, I hadn't even noticed.  They should all be ?= not explicit =.  absolutely correct.

Sh4rK:
Superseded by #567.

PR #567, opened 2015-11-16
make: Allow overriding of CC and AR in Makefiles

Sh4rK:


karlp:
Did you try this?  it doesn't work out of the box now, as CC isn't being exported or something.  It only works if you set CC=arm-none-eabi-xxxx explicitly now unfortunately.  What I'd _really_ like to see is someone taking the time to pull up all those common sets of warnings, compiler definitions, ARFLAGS, and so on into a common file.  sort of like, https://github.com/karlp/libopencm3/commit/786a6fc934e29f66c506b5fe099ebbb62c126e6d only that has the same problem as this patch :)

Sh4rK:
You're right, I didn't try it :)

In the long term, I think it would be nice to migrate to something other than hand-written makefiles (I like CMake nowadays).

PR #573, opened 2015-11-23
STM32 USART: Support for usart_rx_data_ready()

danielinux:
In our opinion, uart/usart needs a mechanism to poll for new data in the RX buffer before issuing a read. 

In a OS context, the scheduler needs to kick in when the device does not have anything to read, until the next IRQ from the serial port, so the usart_wait_recv_ready() is not usable as it blocks on a tight loop instead.


kungjohan:
I think its a good idea.
Similar to
https://github.com/libopencm3/libopencm3/pull/523

karlp:
Why should this patch be chosen over #523?

danielinux:
Indeed #523 looks better and it is more complete. Please consider pulling #523 instead.

PR #575, opened 2015-12-06
Drop "install" target and documentation.

karlp:
This has hurt many many many people over it's lifetime, by confusing their
multilib toolchains.  Simply drop it outright.  People who _really_ know what
they're doing are still perfectly entitled to "install" portions of this
project in suitable locations for their own use.


ChuckM:
:+1: I'd give it +100 if could :-)

PR #577, opened 2015-12-12
stm32f0: Add BOOT_SEL and nBOOT0 definitions for FLASH_OBR

devanlai:
Checked against RM0091rev8


fenugrec:
Also OK according to RM0091 rev7 p-)

karlp:
Thanks, merged in c28d972

PR #580, opened 2015-12-18
Fixed misc. errata found in efm32/lg include files.

kbob:
Looking through the datasheets, I noticed a few small errors in the efm32/lg include files.
Fixed here.


kuldeepdhaka:
:+1:  looks good to me

esden:
Thanks! :)

PR #581, opened 2015-12-18
First step toward SiLabs EFM32WG "Wonder Gecko" support.

kbob:
Just enough to make miniblink run.

This is more of a request for comments than a pull request.

The EFM32LG and WG chips are 98% the same, at least in these modules.  Do we want to put all this into a "common" directory and have the 2% in the lg and wg directories?


karlp:
I don't know what the efm code looks like already, but one comment is that on most other platforms, where there's a macro for setting an option, the option definitions are the values, not the combined again values...

So,

CMU_CTRL_CLKOUTSEL1(v)  (mask and shift from 0,1,2,3)
CMU_CTRL_CLKOUTSEL1_LFRCO	CMU_CTRL_CLKOUTSEL1(0)
CMU_CTRL_CLKOUTSEL1_blah	  CMU_CTRL_CLKOUTSEL1(1)
CMU_CTRL_CLKOUTSEL1_wop	CMU_CTRL_CLKOUTSEL1(2)

is what you've got here,   normally, the _LFCRO, _blah, _wop would be jsut the defintiions 0,1,2, not the fully shifted values.  

So final usage is often, CMU_CTRL_CLKOUTSEL(CMU_CTRL_CLKOUTSEL_BLAH)  on other platforms...  maybe that's too wordy, but it is how it's done.  Doing that here would result in very unexpected behaviour.

kbob:
I'm happy to change it, but the efm32/lg/ files, which I cribbed from, do it the same way.  So they should be made consistent.

esden:
Agreed for consistency sake we have to change the behavior to the way it is done on other platforms. I missed it when I merged the gecko support. Thanks for catching that @karlp

esden:
umm wait... I have looked into that ... I am not sure I understand you correctly @karlp. For example the stm32f3 rcc does the following for register values:
```
#define RCC_CR_PLLRDY                           (1 << 25)
#define RCC_CR_PLLON                            (1 << 24)
#define RCC_CR_CSSON                            (1 << 19)
#define RCC_CR_HSEBYP                           (1 << 18)
#define RCC_CR_HSERDY                           (1 << 17)
#define RCC_CR_HSEON                            (1 << 16)
/* HSICAL: [15:8] */
/* HSITRIM: [7:3] */
#define RCC_CR_HSIRDY                           (1 << 1)
#define RCC_CR_HSION                            (1 << 0)
```

So these are fully shifted values that you get when using the defines. As far as I can tell EFM32 code does the same only by using a macro to do the shifting instead of having the shift in place. It ends up being more verbose (not sure that is useful on it's own) but the usage would be the same... Am I missing something?

esden:
Do you mean the fact that the values are being masked in the process too? I can see that being an inconsistency to other platforms. But the result should be the same after all, the masking seems to be redundant as far as I can tell. The values of the macros should be the same though...

ChuckM:
My 2 cents on this; 

Situation 1; you had a bit in the register (a flag) then that is pre-shifted. So you have

`#define FOO_REGISTER_FLAG        (1 << constant_shift)`

Situation 2: you have a value that is in the middle of register and its a numeric value you provide the mask and shift
```
#define FOO_REGISTER_VAL_SHIFT         nn
#define FOO_REGISTER_VAL_MASK          0x3
```

Situation 3: you have a value that is in the middle of a register and its a numeric value but the
values represent individual choices and may or may not be continuous

```
#define FOO_REGISTER_VAL_SHIFT      nn
#define FOO_REGISTER_VAL_MASK      0xn
#define FOO_RESISTER_VAL_value       ((FOO_REGISTER_VAL_MASK & n) << FOO_REGISTER_VAL_SHIFT)
```

Usually I'd make that last define a macro so that I could just say `REGISTER_VAL(x)` and have all the shifting and masking done.  But macros do make it easier to get an error introduced and propagated through a bunch of values. 

Alternatively we could define an enum for the values and do that assignment in the LOC3 code (which is done in the RCC code for the ST series). This makes the code clearer, but you can't match the bits 1:1 with their counter part in the register itself.

karlp:
Those RCC_CR defines are individual bits, not the values of a multi bit field, so not really the same thing, look at things like RCC_CFGR_MCO_* or any of the GPIO_ macros.  Have we been perfectly consistent in all ways on this? no, but generally, on stm32 at least, the _values_ for multi bit fields have been defined unshifted.  It doesn't matter at all for people using the convenience functions, only has any impact on people _writing_ the convenience functions, or using just the .h files from libopencm3.

kbob:
This is superceded by #606.

PR #583, opened 2015-12-21
[usb-otg] added missing field-definitions for GOTGINT register

H2OBrain:
SEDET can be used to detect 'cable disconnect'


karlp:
merged in https://github.com/libopencm3/libopencm3/commit/92845bc50898974b87de9a8362f9d49116efd166 while I was reviewing #585 thanks!

PR #585, opened 2015-12-21
Usb otg cable disconnect

H2OBrain:
Resets usb on cable disconnect
(does what usb-reset would do at the moment..)


karlp:
Hi, I've cherry-picked the field definitions in 92845bc, but I'd like to leave off the reset device proposal for later.  Small steps I know, but progress.

H2OBrain:
nice, I completely missed that you merged this one :D

karlp:
What's left in this one?  It  looks like it adds an interrupt, and then handles it itself?  What does that bring? What does user code need to do with this?

H2OBrain:
That was a long time ago.. I think it helps the USB stack surviving a cable disconnect and reconnect on a running device in some cases. Or maybe I used it to kind of reset the USB device by replugging the cable.
Are you working on USB again? I could test my devices behaviour with and without this patch..

PR #582, opened 2015-12-20
Testgadget0 stm32f429i

H2OBrain:
Contents
- Adds test-support for gadget0 for the stm32f429i-disco board
- Adds a (failing) test case - control-loopback ("1 word lost in the beginning of control out transfer")
- Adds an interrupt driven usb version for stm32f429i, because there the test fails more often (felt)
- Fixes the new rcc timing names for stmf4 and stmf429

Output on windows with libusbK(device crashes at the EEE..-part)
```
$ python test_gadget0.py
.........ss.F.....EEEEEEE

>>> removed Errors (they don't appear on linux) <<<

======================================================================
FAIL: test_loopback (__main__.TestControlTransfer_Reads)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "test_gadget0.py", line 279, in test_loopback
    for i in range(5,ep0_size) : inner(i)
  File "test_gadget0.py", line 277, in inner
    self.assertEqual(buf_in, buf_out, "Buffers don't match!\n - buf_out : %r\n - buf_in  : %r"%(buf_out.tostring(),buf_in.tostring()))
AssertionError: Buffers don't match!
 - buf_out : '5678901234567890123456789012345678901234567'
 - buf_in  : '9012345678901234567890123456789012345676\x00\x00\xa7'

----------------------------------------------------------------------
```

PS: Could we remove the "# serial of my f4 disco board." line from openocd.stm32f4disco.cfg?

PR #591, opened 2016-01-05
[test_gadget-zero] added stm32f429i-disco support

H2OBrain:
replaces #582

Contents
- Adds test-support for gadget0 for the stm32f429i-disco board
- Fixes the new rcc timing names for stmf4 and stmf429
- Adds an interrupt driven usb version for stm32f429i, which is disabled by default


karlp:
Merged with changes in  5996ac6

PR #592, opened 2016-01-05
Gadget0 failing control loopback test

H2OBrain:
Contents
- Adds test-support for gadget0 for the stm32f429i-disco board

### DIFF
--- a/tests/gadget-zero/main-stm32f429i-disco.c
+++ b/tests/gadget-zero/main-stm32f429i-disco.c
@@ -33,27 +33,50 @@
     do { } while (0)
 #endif
 
+
+//#define USE_USB_INTERRUPT
+
+
+static usbd_device *usbd_dev;
+
+#ifdef USE_USB_INTERRUPT
+
+void otg_hs_isr(void)
+{
+	usbd_poll(usbd_dev);
+}
+
+#endif
+
+
 int main(void)
 {
 	rcc_clock_setup_hse_3v3(&rcc_hse_8mhz_3v3[RCC_CLOCK_3V3_168MHZ]);
-	rcc_periph_clock_enable(RCC_GPIOA);
-	rcc_periph_clock_enable(RCC_OTGFS);
+	rcc_periph_clock_enable(RCC_GPIOB);
+	rcc_periph_clock_enable(RCC_OTGHS);
 
-	gpio_mode_setup(GPIOA, GPIO_MODE_AF, GPIO_PUPD_NONE,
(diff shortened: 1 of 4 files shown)


H2OBrain:
WIKI proposal

how to test usb (linux) :
- install openocd, pyusb, GCC ARM Embedded
- clone libopencm3
- cd libopencm3
- make
- cd tests/gadget-zero
- edit openocd.stm32f4disco.cfg
    - adapt hla_serial to your debugger board (dmesg?) (just comment it out :))
- edit test_gadget0.py
    - uncomment DUT_SERIAL belonging to your board
- make -f Makefile.stm32f429i-disco all flash
- python -m unittest test_gadget0

karlp:
Aim here is to make the loopback test behave the same way as the linux kernel gadget 0 loopback test.  (Late thoughts, this is control req loopback? not data loopback?)

H2OBrain:
It's definitely a control transfer.. (loopback like..)
- send known data to device (store in static buffer)
- read size of known data from device (read from static buffer)
- compare known data with read data

the device fails in the first step with the word-loss error)
looks like this
```
 - buf_out : '5678901234567890123456789012345678901234567'
 - buf_in  : '9012345678901234567890123456789012345676\x00\x00\xa7'
```

tests/gadget-zero/main-stm32f429i-disco.c
```diff
@@ -0,0 +1,82 @@
...
+	
+#ifndef USE_USB_INTERRUPT
+	while (1) {
+		usbd_poll(usbd_dev);
+	}
+#else
+	nvic_enable_irq(NVIC_OTG_HS_IRQ);
+	
+	volatile int i = 1;
+	while (i);
```
fenugrec: "while (i)" ? Is there something else modifying "i" ?

H2OBrain: no, it's just a while (1) asm(""); or maybe even while (1);
i'm never sure what the compiler is going to optimise out in the end

fenugrec: I think the standard "while (1) {} " construct is pretty invincible, and compiles always to an endless loop.

karlp:
I've pulled an ~equivalent ctrl loopback test code locally, (and can confirm some of the other bugs) but I've based it more on the linux gadget0/intel compliance code, rather than making up our own new codes for it.  Will close this once it lands.

karlp:
These tests (or equivalent) have merged, along with the fixes for them. thanks for pushing on this.

PR #593, opened 2016-01-05
Fix broken mutex_trylock()

MightyPork:
Implemented zyp's fix for broken mutex. If it's 1 (= failure) by default, the function works fine.

irc log for reference
```
<zyp> strex returns 1 if it fails, 0 if it's successful
<zyp> but if the mutex is already locked, line 57 skips the status update and status gets remains at the initial value which means successful, which is wrong
<zyp> changing line 54 to status = 1 should do the trick
```


karlp:
merged in fb410a4 after amending the commit header thanks for this!

PR #596, opened 2016-01-07
Return configuration = 0 in Addressed state. 

svofski:
Return configuration = 0 in Addressed state in usb_standard_get_configuration(). This enables passing of USB-IF Chapter 9 tests.


karlp:
Looks good.  I'd like to add a testcase for thsi to the tests/gadget-zero python tests to enforce this behaviour is preserved too.

karlp:
Thanks for this.  I've added the test, fixed some whitespace issues and reworded the commit log a little, but always great to get more in line with the specs!

Issue #475, opened 2015-07-16
cm3/cortex.h mixed bool/uint32_t usage


karlp:
As indicated in the comments for https://github.com/libopencm3/libopencm3/pull/323 there's some somewhat unusual mixes of bool and uint32_t in the cm3/cortex.h file.

Arguably, many of the bool types should be uint32_t, as they're dealing with registers anyway.  Further, the register keyword is almost entirely irrelevant, but should probably be double checked with the inline asm.

fenugrec:
(TL;DR summary at the end)
For the functions cm_is_masked_interrupts() and similar, in include/libopencm3/cm3/cortex.h,  ( https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/cm3/cortex.h#L87 ),
the cast to bool for the return value is quite OK in C99.
Ref C99 n1256, sec 6.3.1.2, "When any scalar value is converted to _Bool, the result is 0 if the value compares equal to 0; otherwise, the result is 1.".
This is great, it amounts to " return (result !=0)? 1:0"

For the 'register' keyword, I would agree that it should be irrelevant. Parsing the __asm__ line, the compiler already knows the instruction writes to a register, that eventually needs to be cast to bool (should be a simple "cmp rn, 0" ). Plus, original CMSIS code uses a similar
```
uint32_t __get_PRIMASK(void) {
  uint32_t result;
  __ASM volatile ("MRS %0, primask" : "=r" (result) );
  return(result);
}
```
**** however ****
tests with arm-gcc 5.2.0 give interesting results : (tested only for cortex-M0 output, since it has the most limited instruction set and may show subtle differences)

1- with no optimizations, as expected the code is silly and the generated asm implements quite literally what was written, i.e. a local variable on stack if the "register" keyword is removed. Interestingly, gcc spews 5 instructions in addition to "cmp rn, 0" : 
```
	MRS r3, PRIMASK
	movs	r4, r3
	movs	r3, r4
	subs	r2, r3, #1
	sbcs	r3, r3, r2
	uxtb	r3, r3
	cmp	r3, #0
```

2- with -Os and -O2, gcc takes liberties and exposes an issue: in some cases, depending on the test code, it can optimize away some "MRS rn, primask" instructions if cm_is_masked_interrupts() is called multiple times in the same context / scope. This is most probably due to the lack of "volatile" keyword on the __asm line. From the gcc docs : 
" You can prevent an asm instruction from being deleted by writing the keyword volatile
[...]

PR #599, opened 2016-01-12
Cortex : changed interrupt handling functions to use uint32_t (GH issue #475)

fenugrec:
Cortex : changed interrupt handling functions to use uint32_t instead of bool.

This is more in line with the actual hardware (u32 registers), and will still work if PRIMASK or FAULTMASK ever have more than 1 bit defined.
The functions cm_is_masked_interrupts() and cm_is_masked_faults() are unchanged, since returning 'bool' fits with the function naming.
Fixes most of github issue #475 . What remains "unfixed" is the absence of functions to simply get the u32 value of PRIMASK and FAULTMASK registers.


karlp:
works for me.

karlp:
merged via rebase in 1d36685

PR #601, opened 2016-01-21
cm3: remove duplicate confusing MPU defines.

karlp:
The MPU RASR AP table has a duplicate entries for Privileged ReadOnly
and Usermode ReadOnly, in the source ARM document (Cortex M3 TRM)
Remove the duplicate here.

The MPU RASR Shareable, Bufferable and Cacheable bits are all individual
bits, and none of the existing defines appear to even match the ARM
documentation.  Remove them, but leave the definitions of the bit
positions.

Reported by MightyPork on IRC


karlp:
merged via rebase in 58f9909474041b83b91abbc798efa5df742eaf40

PR #603, opened 2016-01-25
USB Audio Support

svofski:
- Descriptors required to declare a USB Audio Class Device
- Extended generic endpoint that includes fields required for isochronous endpoints
- Constants


esden:
This needs to be rebased if we are considering adding this.

I am marking this as stale. If it is not picked up again in a month I will close this issue.

PR #604, opened 2016-01-25
USB Isochronous Endpoint Support

svofski:
Juggle ODD/EVN frame bits for ISO endpoints.


karlp:
The commit "	Support for isochronous endpoints in stm32fx07" looks generally good by itself. Do you have any test code that can exercise this?  Adding it either into, or as a new set of tests like the tests/gadget-zero test cases would be greatly appreciated.

svofski:
Yes I have a project that uses all 3 recent PRs together to implement an USB Audio DAC. I would like to make an example that would be a thing in itself but I don't have a standard board. Perhaps just sending back some text stats about received stream back over CDC would be an ok example? 

I think even just having a properly written down set of descriptors is already a worthy sample. I spent quite a bit of time creating them from scratch.

PR #605, opened 2016-01-25
Add ISO IN transfer incomplete callback

svofski:
void usbd_register_incomplete_callback(usbd_device *usbd_dev, void (*callback)(void));

Invoked when IISOIXFR interrupt flag is set (TX FIFO not empty at the end of frame).


karlp:
The commit "	Isochronous IN transfer incomplete callback …" looks good by itself, but this is only for fx07?

svofski:
The interface is generic but the implementation is currently only for fx07, correct.

PR #606, opened 2016-01-26
Split EFM32 into common + families.  Added efm32wg and ezr32wg families.

kbob:
Moved @kuldeepdhaka's EFM32LG changes into efm32/common and linked them from efm32/lg.
Added efm32wg and ezr32wg families.

Also collected several errata along the way.


kbob:
Update 2016-02-17.

While using this on the EZR32WG, I found out that libopencm3 wasn't being built with hard-fp compiler flags, and also that the interrupt vector table was not being initialized.  Fixed both, see above.

karlp:
a lot of this overlaps with the addition of the HG family too, and can probably all come in as well.

karlp:
this is all pulled in to my staging now.  I think the only thing to check is whether all the efm32 can use the refactored usb drivers that came in with the HG commits.  (I think they can) and then just drop the existing common one, and rename the HG driver.

karlp:
This all landed via 74316e69..dd87b5be

PR #607, opened 2016-01-26
tests: extract serials to optional local config

karlp:
Instead of having committed files containing a single developer's serials, use
optional includes to include a local config file for each board.

If you have only a single board connected, simply:

$ make -f Makefile.<board> clean all flash

If you have multiple boards connected, make sure to fill in the appropriate
hla_serial in your openocd.<board>.local.cfg file, and then run the same
commands.

### DIFF
--- a/tests/gadget-zero/.gitignore
+++ b/tests/gadget-zero/.gitignore
@@ -1,0 +1,1 @@
+openocd.*.local.cfg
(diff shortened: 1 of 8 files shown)


H2OBrain:
Looks fine for me. I have not tested it tough.
So if I do __not__ create this config files it will use the first stlink/v2 it finds?

Now we only need something similar for the python script (or an optional console argument?)

karlp:
That's correct, if the files don't exist, it's just using a regular "interface/stlink target/stm32xxxxx" pair, so it will attempt the first vid:pid matching it finds.

karlp:
Anyone up for testing that this works for them for multiple discovery boards connected with your own locally provided serial numbers?

karlp:
merged, seemed to work ok, and fixed a valid issue

PR #608, opened 2016-01-29
Added extra clock config

kevindehecker:
This happens to be the Pixhawk IO config.


karlp:
Is this used on an F100 or something?  24mhz sysclock isn't a very common or useful speed.  I'd rather see F1 refactored to take clock configs like most of the other stm32 series, rather than add yet more copypasta with minor tweaks for different speeds.

kevindehecker:
Yes, it's for the stm32f103C8T6 medium density device.

Agree about the copy pasta, the file is full of little mistakes currently (mainly in the comments but still). However, I don't have the time nor expertise  to do that.

karlp:
As a very non-standard speed (not full speed, and not a low power speed either) I'm not going to merge this as useful code.  It's the sort of thing that should just remain in user application code.  Even after f1 moves to a struct based init like most other platforms, it would still not be a desirable preconfigured struct

PR #609, opened 2016-01-31
added rcc_hsi16_div4() and rcc_set_pll_source() for stm32l0 (tested)

kaeipnos:
I started to use libopencm3  for a small project with an nucleo-L053R8 board, but It seems that some key functions were missing ( hsi16 divider and pll source switch ).
Of course those functions are trivial to implement but it makes source code a bit more readable.
It seems that stm32l0 targets aren't well supported.

I've also started a miniblink example for this board that uses those functions, and if you think it can be useful I can make another pull request on libopencm3-examples when it's ready.

### DIFF
--- a/include/libopencm3/stm32/l0/rcc.h
+++ b/include/libopencm3/stm32/l0/rcc.h
@@ -160,6 +160,8 @@
 #define RCC_CFGR_PLLMUL_MASK		0xf
 
 /* PLLSRC: PLL entry clock source */
+#define RCC_CFGR_PLLSRC_SHIFT 16
+#define RCC_CFGR_PLLSRC_MASK (1 << RCC_CFGR_PLLSRC_SHIFT)
 #define RCC_CFGR_PLLSRC_HSI16_CLK	0x0
 #define RCC_CFGR_PLLSRC_HSE_CLK		0x1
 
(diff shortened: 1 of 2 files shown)


karlp:
Correct, L0 support is very new

include/libopencm3/stm32/l0/rcc.h
```diff
@@ -160,6 +160,8 @@
 #define RCC_CFGR_PLLMUL_MASK		0xf
 
 /* PLLSRC: PLL entry clock source */
+#define RCC_CFGR_PLLSRC_SHIFT 16
+#define RCC_CFGR_PLLSRC_MASK (1 << RCC_CFGR_PLLSRC_SHIFT)
```
karlp: no need for mask and shift definitions for single bits.

lib/stm32/l0/rcc.c
```diff
@@ -408,5 +408,50 @@ void rcc_set_hpre(uint32_t hpre)
...
+ */
+
+void rcc_hsi16_div4(uint32_t hsi16diven)
+{
+	switch(hsi16diven) {
+		case RCC_CR_HSI16DIVEN:
+			RCC_CR = ( RCC_CR & ~RCC_CR_HSI16DIVEN) | RCC_CR_HSI16DIVEN;
+			break;
+		case 0:
+			RCC_CR = ( RCC_CR & ~RCC_CR_HSI16DIVEN);
```
karlp: stylecheck problems here.

lib/stm32/l0/rcc.c
```diff
@@ -408,5 +408,50 @@ void rcc_set_hpre(uint32_t hpre)
...
+			(RCC_CFGR_PLLSRC_HSE_CLK << RCC_CFGR_PLLSRC_SHIFT);
+		break;
+	case RCC_HSI16:
+		RCC_CFGR = (RCC_CFGR & ~RCC_CFGR_PLLSRC_MASK) |
+			(RCC_CFGR_PLLSRC_HSI16_CLK << RCC_CFGR_PLLSRC_SHIFT);
+		break;
+	default:
+		break;
+	}
+}
```
karlp: See for instance rcc_set_pll_source for the L1.  same bit, same configuration, we can make it the same code.

karlp:
pll source has landed, if people still want the div4 support we can make a new PR for that.

PR #610, opened 2016-02-02
UNTESTED: genlink: update README to correct omissions

karlp:
Reported by quitte on irc, but not actually tested at this point.


karlp:
quitte went on and did this all properly. this was just an example.

PR #572, opened 2015-11-22
Support for LM3S: RCC and UART

danielinux:
Hello, 
please consider the introduction of RCC and UART support for the LM3S platform. Even if old, this target allows to run on qemu-system-arm which is often useful.

PR #613, opened 2016-02-05
LM3S/Qemu support

danielinux:
This branch introduces support for LM3S stellaris RCC + basic UART.

This is largely based on #572 - taking into account all the comments from @karlp 

The patch is split into three commits:
- support for basic UART and RCC functionality in LM3S
- board-specific linker script for lm3s6965
- qemu examples in the test/ directory

Feel free to comment.


include/libopencm3/lm3s/usart.h
```diff
@@ -0,0 +1,136 @@
...
+#define USART_IC_BE             (0x01 << 9)
+#define USART_IC_PE             (0x01 << 8)
+#define USART_IC_FE             (0x01 << 7)
+#define USART_IC_RT             (0x01 << 6)
+#define USART_IC_TX             (0x01 << 5)
+#define USART_IC_RX             (0x01 << 4)
+/* Bits [3:0] - Reserved */
+
+#define USART_SR_RXNE  USART_IC_RX
+#define USART_SR_TXE   USART_IC_TX
```
karlp: we spoke about this before didn't we?

lib/lm3s/rcc.c
```diff
@@ -0,0 +1,88 @@
...
+{
+    uint32_t rcc = RCC_RESET_VALUE;
+    uint32_t rcc2 = RCC2_RESET_VALUE;
+
+    /* Initials: reset values */
+    RCC_CR = rcc;
+    RCC2_CR = rcc2;
+
+    /* Stage 1: Reset Oscillators and select configured values */
+    rcc = RCC_SYSDIV_VAL(4) | RCC_PWMDIV_DEFAULT | RCC_XTAL_8MHZ_400MHZ | RCC_USEPWMDIV;
```
karlp: I'm not really sure it's ok to have a single _init(void) function that does such a hardcoded clock selection.  Consider how f4 and l1 do it at least, or if you _must_ consider the way f1 does it.

lib/lm3s/usart.c
```diff
@@ -0,0 +1,151 @@
...
+void usart_enable(uint32_t usart)
+{
+	(void)usart;
+}
+
+void usart_disable(uint32_t usart)
+{
+	(void)usart;
+}
+
```
karlp: just don't even implement this stuff.  Functions that appear to all exist and if you don't happen to go all the way into the source you don't notice that they are actually no-ops aren't helping anyone.

danielinux: Ok, I'll remove the stubs

lib/lm3s/rcc.c
danielinux: Do you mean a more meaningful name for the rcc initialization such as `rcc_clock_setup_in_xtal_8mhz_out_50mhz` ? Should I also implement other clock configurations, since I can only test one target?

include/libopencm3/lm3s/usart.h
danielinux: can't recall... what in particular? Can you provide an example for defining single bits in a bitmask the opencm3 way?

karlp: wsan't it that you were using the actual regstier/bit definitions from stm32 familes as aliases for your own use?  It's useful in your own code, but out of scope for an lm3s definition file.

danielinux: those names (registries, values) are taken from the datasheet of the lm3s6965

karlp: USART_SR_RXNE? and USART_SR_TXE? not in the lm3s datasheet I'm looking at :)

lib/lm3s/rcc.c
karlp: either that, the wordy f1 style, that at least isn't confusingly vague and openended, or go the f4 style and allow more open configuration, or, entirely leave this out of the library altogether until someone else does, and just have this in your example application

include/libopencm3/lm3s/rcc.h
```diff
@@ -0,0 +1,94 @@
...
+ */
+
+#ifndef LIBOPENCM3_RCC_H
+#define LIBOPENCM3_RCC_H
+#include <libopencm3/cm3/common.h>
+
+/* Interface required types */
+
+#define RCC_CLK_ENA_E 0
+#define RCC_CLK_RST_E 1
```
karlp: what are these two used for?

include/libopencm3/lm3s/rcc.h
```diff
@@ -0,0 +1,94 @@
...
+
+/* --- RCC registers ------------------------------------------------------- */
+
+#define RCC_RIS             MMIO32(0x400FE050)
+#define RCC_CR              MMIO32(0x400FE060)
+#define RCC2_CR             MMIO32(0x400FE070)
+
+#define RCC_RESET_VALUE   (0x078E3AD1)
+#define RCC2_RESET_VALUE  (0x07802810)
```
karlp: interesting, what are you planning on using these for?

include/libopencm3/lm3s/rcc.h
```diff
@@ -0,0 +1,94 @@
...
+
+/* RCC1 bits */
+#define RCC_SYSDIV    (1 << 23)
```
karlp: SYSDIV is 4 bits, this is a very non-standard way of defining that.

include/libopencm3/lm3s/rcc.h
```diff
@@ -0,0 +1,94 @@
...
+#define RCC_SYSDIV_MASK  (0x0F << 23)
+#define RCC_SYSDIV_VAL(X)    (((X - 1) << 23) & (RCC_SYSDIV_MASK) )
```
karlp: I'd have to check, but from memory, other places that need -1s don't do it in the macros, the user puts in the value they want, and the macro just takes care of putting the value in correctly.

include/libopencm3/lm3s/rcc.h
```diff
@@ -0,0 +1,94 @@
...
+
+#define RCC_USESYSDIV (1 << 22)
+#define RCC_USEPWMDIV (1 << 20)
+
+#define RCC_PWMDIV    (1 << 17)
+#define RCC_PWMDIV_MASK (0x07 << 17)
+#define RCC_PWMDIV_DEFAULT (0x07 << 17)
```
karlp: is this _default useful for anything?  same comments on the definitons of multibit fields here.   It all appears self-consistent at least.

include/libopencm3/lm3s/rcc.h
danielinux: leftovers. Removing.

include/libopencm3/lm3s/rcc.h
danielinux: sorry, these were hardcoded values used for testing. This is not right, removing these values.

include/libopencm3/lm3s/rcc.h
danielinux: This is marked as default for PWMDIV in the datasheet.

lib/lm3s/rcc.c
```diff
@@ -0,0 +1,89 @@
...
+ */
+
+#include <stdint.h>
+#include <libopencm3/lm3s/rcc.h>
+#include <libopencm3/cm3/sync.h>
+
+/* PLL/Systick Configured values, default settings for LM3S6965EVB */
+#define LM3S_SYSDIV_VAL        (3)
+#define LM3S_PWMDIV_VAL        (7)
+#define LM3S_XTAL_VAL          (14)
```
karlp: what are these for?

include/libopencm3/lm3s/rcc.h
danielinux: Will change into ...PWMDIV_64

lib/lm3s/rcc.c
danielinux: To be removed

lib/lm3s/rcc.c
danielinux: resolved using wordy f1 style

include/libopencm3/lm3s/rcc.h
danielinux: Actually, these values are OK. They come from the datasheet, and they are marked as "magic" reset values (in RCC_CR, this also includes a reserved bit on pos 12).

This seems the only way to reset RCC, so I will leave it like this.
Typical TI, there are hardcoded values for register reset. Check datasheet section 6.4 "Register Descriptions".

include/libopencm3/lm3s/rcc.h
danielinux: Roger that, changing into mask + values.

include/libopencm3/lm3s/rcc.h
danielinux: Changing defines into readable values

danielinux:
I am preparing a pull request for the example repo with the qemu application, as suggested by @karlp

include/libopencm3/lm3s/rcc.h
```diff
@@ -0,0 +1,114 @@
...
+
+/* Register 8: RCC
+ * Type R/W, reset 0x078E.3AD1
+ */
+#define RCC_RESET_VALUE   (0x078E3AD1)
+
+/* Register 10: RCC2
+ * Type R/W, reset 0x0780.2810
+ */
+#define RCC2_RESET_VALUE  (0x07802810)
```
karlp: My query on these was not what they were, but what you would ever do with them.  At what point in your application lifecycle are you going to do RCC = RCC_RESET_VALUE?  It's not going to do other things that need to be reset, and resetting individual settings doesn't require a definition of the entire registers' reset state.  I just don't understand what they're for, and adding "RW, xxxx" doesn't seem to make that easier to understand :)

lib/lm3s/rcc.c
```diff
@@ -0,0 +1,79 @@
...
+#include <libopencm3/cm3/sync.h>
+
+int rcc_clock_setup_in_xtal_8mhz_out_50mhz(void)
+{
+    uint32_t rcc = RCC_RESET_VALUE;
+    uint32_t rcc2 = RCC2_RESET_VALUE;
+
+    /* Stage 0: Reset values applied */
+    RCC_CR = rcc;
+    RCC2_CR = rcc2;
```
karlp: Really?  Is this how you want to use them?  Don't you just set the bits you need to do what you need in the steps below?  That's how stm32 and co do it.  DId you have a look at the lm4f at all?

lib/lm3s/rcc.c
```diff
@@ -0,0 +1,79 @@
...
+    __dmb();
+
+    /* Stage 1: Reset Oscillators and select configured values */
+    rcc = RCC_SYSDIV_50MHZ | RCC_PWMDIV_64 | RCC_XTAL_8MHZ_400MHZ | RCC_USEPWMDIV; 
+    rcc2 = RCC2_SYSDIV2_4;
+    RCC_CR = rcc;
+    RCC2_CR = rcc2;
```
karlp: FWIW, while I don't really mind per se, this style is rather more verbose than seems necessary.  in most other families, you just set the register directly.
```
RCC_CR = RCC_SYSDIVxxx | blah | wop;
RCC2_CR = RCC2_SYSDIV2_4;
```

lib/lm3s/rcc.c
danielinux: Seems like RCC reset value contains a reserved bit as well. Can you help me finding a better way to set these values at startup? Looks like RCC needs these values set before initiating the clock setup, which implies multiple steps.

lib/lm3s/rcc.c
danielinux: Agree, will do

include/libopencm3/lm3s/rcc.h
```diff
@@ -0,0 +1,114 @@
...
+
+
+#define RCC_OSCRC_MASK   (0x03 << 4)
+#define RCC_OSCRC_MOSC   (0x00 << 4)
+#define RCC_OSCRC_IOSC   (0x01 << 4)
+#define RCC_OSCRC_IOSC_Q (0x02 << 4)
+#define RCC_OSCRC_30KHZ  (0x03 << 4)
+
+#define RCC_IOSCDIS   (1 << 1)
+#define RCC_MOSCDIS   (1 << 0)
```
karlp: you're missing quite a few parts here.

include/libopencm3/lm3s/rcc.h
```diff
@@ -0,0 +1,114 @@
...
+/* RCC2 bits */
+#define RCC2_USERRCC2       (1 << 31)
+#define RCC2_SYSDIV2_MASK   (0x7f << 23)
+
+#define RCC2_SYSDIV2_1      (0  << 23)
+#define RCC2_SYSDIV2_2      (1  << 23)
+#define RCC2_SYSDIV2_4      (3  << 23)
+#define RCC2_SYSDIV2_8      (7  << 23)
+#define RCC2_SYSDIV2_16     (15 << 23)
+#define RCC2_SYSDIV2_32     (15 << 23)
```
karlp: this is wrong, also, i don't think these need to be individual defines, the divider is just the number you want -1.

lib/lm3s/usart.c
```diff
@@ -0,0 +1,98 @@
...
+#include <libopencm3/lm3s/usart.h>
+
+void usart_send(uint32_t usart, uint16_t data)
+{
+	USART_DR(usart) = data;
+}
+
+uint16_t usart_recv(uint32_t usart)
+{
+	return USART_DR(usart) & 0xff;
```
karlp: are you sure you want an api indiciated a uin16t, but you're masking it to 8bits?

lib/lm3s/usart.c
```diff
@@ -0,0 +1,98 @@
...
+}
+
+void usart_wait_send_ready(uint32_t usart)
+{
+	while ((USART_FR(usart) & USART_FR_TXFE) == 0);
+}
```
karlp: possibly, usart_is_send_ready()  and is_recv_ready() instead? then send_blocking can just do while (!is_ready()) and you have a better async api available for the user?

lib/lm3s/usart.c
```diff
@@ -0,0 +1,98 @@
...
+
+bool usart_is_recv_ready(uint32_t usart)
+{
+	return((USART_FR(usart) & USART_FR_RXFE) == 0);
+}
+
+bool usart_is_send_ready(uint32_t usart)
+{
+	return((USART_FR(usart) & USART_FR_BUSY) == 0);
+}
```
karlp: ah ok, so what do we still need the other two blocking versions for?

karlp:
I've got this staged and mangled a bit in my current tree:   https://github.com/libopencm3/libopencm3/pull/646

include/libopencm3/lm3s/rcc.h
danielinux: My point here is that the datasheet recommend to set these "magic" values to reset the RCC setting before programming these two registers.

include/libopencm3/lm3s/rcc.h
danielinux: What in particular?

include/libopencm3/lm3s/rcc.h
danielinux: OK, just trying to stick with other arch definitions here, but if you request so, I will indeed change these values to their own dividers.

lib/lm3s/usart.c
danielinux: I seem to remember that other architectures export usart_send()/recv() with the same set of arguments, and then only output/input one byte

lib/lm3s/usart.c
danielinux: Our OS needs those is_send_ready/is_recv_ready for tasks to properly sleep, I provided the blocking versions only for compatibility with other archs (e.g. stm32)

include/libopencm3/lm3s/rcc.h
karlp: Sorry, what were trying to copy from other archs here?  I was saying that having the div 16 and the div 32 being the same define was wrong, and then, as a separate issue, questioning the entire value of having a partial set of these dividers defined at all, when you simply write the divider value as is into the field.

include/libopencm3/lm3s/rcc.h
karlp: That's not how I read the datasheet, but I'll re-read it again :)

include/libopencm3/lm3s/rcc.h
karlp: there's more values for those fields.  I'd rather see a field _not_ defined, than only having some of it's values defined.

karlp:
This has (finally!) landed.  hooray!  all the async is_xxx() handlers are out of scope though, but are valid and will continue to be worked on

Issue #612, opened 2016-02-03
Missing devices in devices.data file


quitte:
The devices.data file has no entries for lpc43xx, lpc43xx_m0, lm4f and vf6xx devices despite the library files being created. This means that the linkerscript generation of mk/genlink doesn't work for any of those.

quitte:
lm4f and vf6xx are added in https://github.com/libopencm3/libopencm3/pull/646.

The lpc43 situation has been partially addressed and will now default to a cortex-m4 build. Some mechanism needs to be added for the user to be able to select wether they want to target the m0 or m4 core. Until then the build of either the m0 or m4 or both variants must be special cased.
This is similar to the situation as-is which is unaffected.

PR #614, opened 2016-02-06
mk/genlink: change devices.data format to remove gcc specific options

quitte:
As discussed with karlp on irc the devices.data file should not contain
gcc specific command line options.

For that reason the command line options for gcc are now generated from
the variables FAMILY,CPU,FPU by the rules in the mk directory.

This breaks the genlink tests.

### DIFF
--- a/mk/README
+++ b/mk/README
@@ -103,7 +103,7 @@ LDSCRIPT	(replaced)
  * No needed to handle this variable if you use module <gcc> too.
 
 LDLIBS (appended)
- - LDLIBS += -l$(OPENCM3_LIBNAME) is appended to link against the
+ - LDLIBS += -lopencm3_$(family) is appended to link against the
    matching library.
 
 LDFLAGS (appended)
--- a/mk/README
+++ b/mk/README
@@ -15,34 +15,31 @@ where you are defining variables (near the beginning of the file), and file
 <module>-rules.mk should be included in the rules part of makefile (somewhere
 near to the end of file).
 
-Example makefile using gcc compiler module:
+Example makefile using the gcc compiler module together with the linker script
+generator module:
 
 >>>>>>
-OBJS		+= foo.o bar.o
+DEVICE          =
+OPENCM3_DIR     =
+OBJS            += foo.o
 
-CFLAGS		+= -O0 -g
-CPPFLAGS	+= -MD -MP $(@F).d
-CPPFLAGS	+= $(DEFS)
-CPPFLAGS	+= $(INCS)
-LDFLAGS		+= --static --nostartfiles
-LDLIBS		+= -Wl,--start-group -lc -lgcc -lnosys -Wl,--end-group
-# parameters for gcc module
-PREFIX		= arm-elf
+CFLAGS          += -Os -ggdb3
+CPPFLAGS	+= -MD
+LDFLAGS         += -static -nostartfiles
+LDLIBS          += -Wl,--start-group -lc -lgcc -lnosys -Wl,--end-group
 
+include $(OPENCM3_DIR)/mk/genlink-config.mk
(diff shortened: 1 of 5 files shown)


quitte:
This is not tested very well. The devices.data format now has less redundant information.

The biggest issue is the mismatch of devices.data format and documentation.

karlp:
see also: https://github.com/libopencm3/libopencm3/issues/612

quitte:
To days changes have been tested by building libopencm3 itself. That seems to work fine, again.

karlp:
one note while I remmeber, is it still possible to override the FP_FLAGS with this stuff?

mk/README
```diff
@@ -81,7 +78,8 @@ genlink
 
   This module adds an support for the user to the linker script generator. The
```
fenugrec: while we're haxing this file... this sentence doesn't make a whole lot of sense. This module adds support for what ?

mk/genlink-config.mk
```diff
@@ -18,17 +18,61 @@
...
+endif
+
+ifeq ($(genlink_fpu),soft)
+ARCH_FLAGS	+= -msoft-float
+else ifeq ($(genlink_fpu),hard-fpv4-sp-d16)
+ARCH_FLAGS	+= -mfloat-abi=hard -mfpu=fpv4-sp-d16
+else ifeq ($(genlink_fpu),hard-fpv5-sp-d16)
+ARCH_FLAGS      += -mfloat-abi=hard -mfpu=fpv5-sp-d16
+else
+$(warning No match for the FPU flags)
```
fenugrec: Maybe this should instead generate an error for "no match found", and say nothing for "no FPU at all" ?

mk/genlink-config.mk
```diff
@@ -18,17 +18,61 @@
 ##
 
 ifeq ($(DEVICE),)
-$(error no DEVICE specified for linker script generator)
+$(warning no DEVICE specified for linker script generator)
```
fenugrec: Why is this now a warning instead of an error ? Wouldn't the following awk commands fail if DEVICE isn't specified ?

karlp:
damnit, lost some commentary in progress.  I've taken this and am working with it in my tree abit.  I propose squishign the first three commits down into one, and have a couple of other tweaks to some of the warning text.  I've got them staged here: https://github.com/libopencm3/libopencm3/pull/646  but might add a few more things there yet.

You mention that the tests fail now?  How were they even run?

karlp:
related to the "tests", the Makefile.linker file and the example there don't seem to be used/functional?

mk/README
quitte: for linker script generation. I'll change that

mk/genlink-config.mk
quitte: Indeed. This arose out of backwards compatibility to the old data file format. Doesn't make sense anymore

quitte: Looking at it again I recall being very cautious about breaking builds that would otherwise work. There are more places where warnings could as well be interpreted as errors in the devices.data file. It's also imaginable that the user actually intended to mix and match things. Therefore I opted for outputting warnings in all cases where I could remotely imagine the configuration to be working for now.

I'd like to revisit those warnings once it's clear if and how the linkerscript generation and target configuration by devices.data is actually used.

mk/genlink-config.mk
quitte: If DEVICE not being specified causes the build to fail there will be a good indication for the reason.
However it's possible that a Makefile includes genlink-config.mk while not needing it (yet). The above check is executed no matter what. That means that any and all make targets would fail because of it for no good reason.

karlp:
I believe all of this has now landed.

PR #617, opened 2016-02-12
RCC, PWR, FLASH, EXTI, SYSCFG and TIMER libraries for l4

benjaminlevine:


karlp:
Can you squish this down so as not to have one commit introducing something then a new commit immediately changing it?  Also, you have whitespace issues in this patch.

karlp:
For exti, did you look at the common code? exti_common_all looks like it has a lot of what you want...

include/libopencm3/stm32/l4/syscfg.h
```diff
@@ -0,0 +1,139 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* THIS FILE SHOULD NOT BE INCLUDED DIRECTLY, BUT ONLY VIA SPI.H
```
karlp: wrong wordsss

karlp:
Further, I recommend setting your "git name" properly.  It's currently 
```
Author: benjaminlevine <benjaminmileslevine@gmail.com>
```
which doesn't match the email given in the files, and doesn't look very nice...

benjaminlevine:
Thanks for your comments.

Regarding flash.c: There are some minor changes between the common code (specifically f24 and f234) e.g. ICE -> ICEN (in line with the reference manual for l4) but the main change is the removal or references to the function "flash_set_program_size" which appears in a number of other functions throughout the common code of f24. I couldn't find an equivalent function for the l4.

I can appreciate that the code copied from f234 is extremely similar (only one substantive change of FLASH_SR_PGPERR -> FLASH_SR_PROGERR, in line with the reference manual).
If you'd rather, I can make the l4 code use the f234 flash library and do without the helper functions of f24. That way we wouldn't need need flash.c

Regarding number of commits: I will squish it down as suggested and look into the whitespace issue

Regarding EXTI: I did look at the common code but I had a problem with it, as I mentioned in comment on ed0e361. More specifically, the EXTI common code relies on bit masks for each EXTI being passed to the helper functions. However, because the l4 splits EXTI over TWO registers this method won't work on l4 e.g. EXTI0 and EXTI32 have the same value. So I had to rewrite the helper functions to differentiate between the two registers by passing in, as a plain number (i.e. not a bit mask), the EXTI value e.g. '1' for EXTI1, '32' for EXTI32 etc. If the value is less than 32 then the value is assigned to a first EXTI register, if it's greater than 32 it's assigned to a second EXTI register. It's not elegant but I couldn't think of a better way to do it.

Regarding SYSCFG.h: Will sort the words

benjaminlevine:
I've fixed a lot of the issues.
Specifically:

- Sorted the words in syscfg.h
- Fixed the email address
- Set up the git name properly
- Fixed the whitespace issues
- Squished the commits

I'll wait for your comments on flash.c and exti

mrnuke:
Hi benjaminlive,

I realize I'm a little late to this review. Would it be possible to split up into several smaller commits, for history purposes? For example, you could start by adding the basic infrastrucure in one commit, then add RCC in another commit, and so on. Every peripheral would get its own commit. This makes it easier for reviewers, and provides a cleaner history.

benjaminlevine:
Sure thing. Good idea

karlp:
I'm picking bits of this as I need it, but this still has multiple style issues, primarily whitespace.

benjaminlevine:
For reference, can you be more specific on the style issues?
I used "make stylecheck" and I few issues with this commit.
Are you talking about the tabbing e.g. in the defines not being consistent?

Thanks

include/libopencm3/stm32/l4/flash.h
```diff
@@ -0,0 +1,256 @@
...
+#define FLASH_ACR_ICEN			    (1 << 9)
+#define FLASH_ACR_PRFTEN		    (1 << 8)
+
+#define FLASH_ACR_LATENCY_SHIFT     0
+#define FLASH_ACR_LATENCY_MASK      0x03
+
+#define FLASH_ACR_LATENCY_0WS		0x00
+#define FLASH_ACR_LATENCY_1WS		0x01
+#define FLASH_ACR_LATENCY_2WS		0x02
+#define FLASH_ACR_LATENCY_3WS		0x03
```
romainreignier: I think a `FLASH_ACR_LATENCY_4WS` is needed.
See in 3.7.1 in L4x6 reference manual ([RM0351](http://www.st.com/resource/en/reference_manual/dm00083560.pdf)).

br101:
@benjaminlevine:

I've been looking at the code for L4 flash.

It contains flash_program_word() but in the reference manuals
(RM0351 and RM0394) i can only find the capability to program double
words (64 bit). Definetly my chip (STM32L433CC) can only do double
words... I wonder, did I miss a L4 variant or did you not test your code
 sorry no offense, I just want to confirm if I missed something... so
I can fix it in case it's necessary.

benjaminlevine:
@br101 - you're quite right. should be to program double words.

br101:
@benjaminlevine thanks will make a pull request soon. btw your email which is mentioned in the code bounces.

benjaminlevine:
Weird. It is correct

br101:
@benjaminlevine <benjamin@jesco.karoo.co.uk>: host mx.vip.karoo.l.kcom.com[213.249.242.244]
    said: 550 #5.1.0 Address rejected. (in reply to RCPT TO command)

br101:
I think all functionalty from this PR has been merged already and it can be closed.

karlp:
ok, thanks, wasn't sure if there were still some pieces outstanding.

PR #619, opened 2016-02-15
fix stm32f4 ahb

smallzzy:


karlp:
You can  make a case for the ahb_ changes here, but you need to remove the change to the default values, and you need to fix the whitespace, and you need to explain what it's for properly in your commit notice.

smallzzy:
#620

karlp:
as noted by author, superseded by #620

PR #620, opened 2016-02-15
stm32f4: provide correct AHB frequency

smallzzy:


karlp:
I see these are missing for f3, f4, l1.  Any chance you could expand this to cover those? clearly missing from the earlier commits in 9ddfcb0  and 40d9ffc

smallzzy:
I have made additional comments. I never used these chips so I suppose they share the same properties.

karlp:
:+1:

karlp:
merged via rebase, thanks

PR #578, opened 2015-12-13
add CAN support for stm32f3 (rebased)

devanlai:
Rebased version of [stm32f3_can](/nuft/libopencm3/tree/stm32f3_can) to resolve merge conflicts.
See #384 for previous discussion.

Since I'm rebasing @nuft 's commits, I'll be explicit with how I handled the merge conflicts so that he can verify that I'm not authoring heinous code in his name.

There were two commits that had conflicts with some of the commits from that occurred in the last year during rebasing:

**"stm32: add CAN support for f3" (99b35da)**
- Conflicted with "Surround all macro parameters with ()" (c72f3d5)
- Conflicted in `stm32/can.h` because most of the definitions were moved into `stm32/common/can_common_all.h` (renamed in a later commit).
- Resolved by re-applying wrapping the macros in `stm32/common/can_common_all.h` with parens and using the new copy of `stm32/can.h` that dispatches to the series-specific versions.
- Rebased commit is ~~1ddf5c6~~ 6220b8d

**"stm32: add CAN support for f0." (acbae65)**
- Conflicted with "usb: Moved USB_PMA_BASE address definition to family-specific memorymap.h" (5c73d60)
- The conflict in `stm32/f0/memorymap.h` was pretty trivial; just a new `USB_PMA_BASE` definition on the line before `BX_CAN1_BASE` was renamed to `BX_CAN_BASE`.
- It also conflicted with several f0 Makefile changes.
- Resolved by appending `can_common_bxcan.o` after `rcc_common_all`.
  - Is there any particular scheme for deciding which object files get appended as part of separate statements vs doing further line continuations? `crs_common_all.o` seems like it logically belongs with the rest of the `_common` object files.
- Rebased commit is ~~b970c70~~ fde0a9e

The rest of the commits automerged during the rebase without requiring intervention.

I did some basic testing with an STM32F042 chip and it seemed okay, but I'm operating with only a CAN_RX pin in silent mode, so it's not a particularly thorough test.

PR #624, opened 2016-02-16
Support CAN on stm32f0

florolf:
The stm32f0 family contains a bxCAN-controller which is compatible with the already existing code. The same seems to go for all other STM32 families, but I haven't been able to check that in detail.

This pull request moves the CAN code to the common/ hierarchy and enables it for the f0 build.


florolf:
This is redundant, see #578.

PR #627, opened 2016-02-29
STM32F3: flash programming operations added

cemizm:


lib/stm32/f3/flash.c
```diff
@@ -50,14 +50,105 @@
 /*---------------------------------------------------------------------------*/
 /** @brief Clear All Status Flags
 
-Clears program error, end of operation, busy flags.
-*/
+ Program error, end of operation, write protect error, busy.
+ */
```
karlp: I undertand adding write protect, but why remove the fairly important "clears" verb?

lib/stm32/f3/flash.c
```diff
@@ -50,14 +50,105 @@
...
+
+/*---------------------------------------------------------------------------*/
+/** @brief Read All Status Flags
+
+ The programming error, end of operation, write protect error and busy flags
+ are returned in the order of appearance in the status register.
+
+ @returns uint32_t. bit 0: busy, bit 2: programming error, bit 4: write protect
+ error, bit 5: end of operation.
+ */
```
karlp: Probably better to just refer to the bits by name, or with @ref and use grouping doxygen around the flags, rather than trying to explicitly number the bits here.

lib/stm32/f3/flash.c
```diff
@@ -50,14 +50,105 @@
...
+
+void flash_erase_all_pages(void)
+{
+	flash_wait_for_last_operation();
+
+	FLASH_CR |= FLASH_CR_MER; /* Enable mass erase. */
+	FLASH_CR |= FLASH_CR_STRT; /* Trigger the erase. */
+
+	flash_wait_for_last_operation();
+	FLASH_CR &= ~FLASH_CR_MER; /* Disable mass erase. */
```
karlp: no need for the extra /* */ comments here.

karlp:
Apart from the general style comments above, please note that this is all common code with f0.  Please combine the f0 code with this in some way, rather than duplicating it further.

cemizm:
Before duplicating I started to combine it with F0 Code. But it lead to rework the whole flash module over all families. In my opinion that should not be part of this pull request.

If you gonna accept this pull request as it is, I could fix the style comments you mentioned. Otherwise please reject and create an issue.

karlp:
Ok well if I merge this, it just makes the work of finally merging the flash code even harder.  So I'm not super motivated to do that.    It's very easy to add code that "improves things for this one platform" but that doesn't help anyone next year.

karlp:
It should be a good beginning ot just make a common/flash_v3.c or something (v1 being f1, v2 something else) and use the _v3 file in f0 and f3.  The current naming of the "common" files makes it very hard to follow, I agree, but you shouldn't have to overhaul the entire flash driver just to start work o n f3 and f0 combining

Issue #537, opened 2015-10-02
TX buffer size is not a multiple of 2 in usb_f103 usb


kuldeepdhaka:
https://github.com/libopencm3/libopencm3/blob/master/lib/usb/usb_f103.c#L123
2 byte alignment is not done, it will cause problem to future endpoint setup

karlp:
Is this a new issue in anyway? is this not the same as https://github.com/libopencm3/libopencm3/issues/461 ?

kuldeepdhaka:
"USB_ADDRn_TX  
Bit 0 Must always be written as ‘0 since packet memory is half-word wide and all packet buffers
must be half-word aligned." [872/988] [DocID018940 Rev 6] [RM0091]

if the max_size is not a multiple of 2 it could result in known behaviour

```
if (dir || (addr == 0)) {
       /* convert max_size multiple of 2 (up) */
	if (max_size & 1) {
		max_size += 1;
	}
       ......
       .....
}
```

karlp:
thanks for the extra detail, much appreciated.

fenugrec:
Update : since the st_usbfs code merge, this is now in st_usbfs_ep_setup() :
https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/common/st_usbfs_core.c#L62

karlp:
I've tried this a lot more, and while I do totally agree that according to the ref man this needs to be "ensured" I'm not at all convinced that you can actually do anything useful with this.  You need to odd byte sized IN endpoints just for _starters_ and then somehow expect those to actually work.

I can't even get the pure source/sink functions to work with odd sized endpoints and mismatched in/out sizes, let alone loopback and witness a buffer trampling.  Unless someone can demonstrate this, I'm closing this as irrelevant.

PR #628, opened 2016-02-29
some STM32 USB fixes

dmitrystu:
1. For both TX and RX endpoints reserved PMA memory should be halfword aligned. (see datasheet)
2. Fixed PMA memory size and NUM_BLOCKS[4:0] calculations for RX endpoints.
3. Fixed looping when endpoint RX transaction stalled. (CTR_RX not cleared after stall and USB machine handles incorrect transaction on each usbd_poll call)


karlp:
Is it possible you can set up a test case for the stalling RX transaction?  I'm pretty sure there's already pull requests for the block size calculation.

dmitrystu:
I will try.
I'ts 100% occurs in case of short (8 byte) control endpoint and WIN10 when host asks a device descriptor or any other data larger that control endpoint. I have no HW bus analyzer and i don't know why the new control RX transaction occurs (CTR_RX set) when TX not yet finished (control_state.state = DATA_OUT). USB machine stalls this transaction but CTR_RX  remains active for this endpoint. As a result "Usb device not recognized"

Tested on STM32L052

About blocksize. First of all TX also should be halfword aligned. And i think here is a bit simple implementation

dmitrystu:
Under Debian 4.2.6-1 x86_64 has the same problem

karlp:
At least part of this seems to overlap with #880 but this simply had too many things in the same PR to track nicely.

fenugrec:
Disconnect : why are we removing `SET_REG(USB_BCDR_REG, USB_BCDR_DPPU);` from the init code ?

karlp:
We're not, (obviously, or obviously not) I'm not merging code that just comments out some stuff.  I was more looking at just keeping the .disconnect registration.

karlp:
1 was covered by https://github.com/libopencm3/libopencm3/issues/537 
2 was handled by #880 and relatives
the disconnect is in, and will land soon.

that only leaves the stall issue, which I'd love more information on reproducing the issue.

PR #630, opened 2016-03-04
STM32F0/3 extended crc unit

cemizm:
Implementation of extended crc unit in f0 and f3


karlp:
This looks perfectly reasonable, but please just name the file crc_v2 or something, this peripheral is also available on the L4 and L0 already.

lib/stm32/common/crc_common_f03.c
```diff
@@ -0,0 +1,113 @@
...
+
+/*---------------------------------------------------------------------------*/
+/** @brief Polynomial coefficient
+
+ Set the coefficients of the polynomial to be used for CRC calculation.
+ If the polynomial size is less than 32-bits, the least significant bits have to be used to program the correct value.
+
+ @note To obtain a reliable CRC calculation, the change on-fly of the polynomial
+ value or size can not be performed during a CRC calculation. As a result, if a CRC calculation is ongoing,
+ the application must either reset it or perform a CRC_DR read before changing the polynomial.
```
karlp: Some long lines here, and maybe some wording that can be clarified around "on-fly"

karlp:
As a general note, there's no need to put the type information in the doxygen, the code already contains that information.

mrnuke:
Sorry to intervene, but why rename it to crc_v2? What sort of information does that convey? Is it a new revision of the same peripheral? How do I know which peripheral I have in my chip? Does it depend on the die revision?

How is _v2 any clearer than crc_f03, or crc_f0_f3, which actually states the product families to which this applies?

karlp:
mrnuke: because it _also_ applies to l0, and l4.  and the continual renaming as new parts appear that could/should use the same code is silly.

it's _v2  (we could call them style_A, style_B or similar if you prefer) because it's the second form of the CRC peripheral seen in STM32 parts.  it's not better or worse, just a different one that came later, and makes it easier to say what's really applicable, rather than having files like, "common_f0347l04" filenames.  and is that an l or a 1.

mrnuke:
"_v2" has no correlation with datasheets. So while you, as an expert, will immediately know what this means, someone learning how to use libopencm3, or who has had no previous experience withe the "v1" peripheral, will have no clue know what the "v2" is in reference to.

"I spoke with github person v24" vs "I spoke with karlp"

"I'm meeting with girfriend v13" vs "I'm meeting Anne"

You're setting a dangerous precedent in which naming has little to do with reality. Want to shorten it, just use crc_fl_04. I know that however you put it, these names suck. You can at least make them parsable.

zyp:
A name that contains no information is better than a name that contains incorrect/incomplete information, because that won't trick people into making wrong assumptions based on it.

Maintaining a list of all supported chips in all peripheral driver filenames is a huge maintenance burden since they would have to be renamed all the time, which not only requires the file to be renamed, but all files referencing it by name also needs to be updated.

Your name argument doesn't work, since these peripherals doesn't have distinct names, they are both «CRC». We still need to distinguish the variants, and numbering them in the order they appeared is the simplest solution.

Trying to work out a consistent parsable scheme is something that would turn ugly really quickly, especially when you consider that there are now peripheral differences even within the families. As an example; F303xC got USB v1 while F303xE got USB v2.

karlp:
What he said :)

While I appreciate the sentiments, (I think I even proposed this sort of naming as being helpful in the past) it's proven to be a drag.  We get pull requests that ignore "common" code because it's labelled the wrong way, duplicating into a "new" chip instead; and users who are confused with why on earth they use the stm32f107 usb driver and the stm32f207 driver on their f4, depending on whether they're using FS or HS usb!

karlp:
Oh, and the other one is that the user just imports "crc.h"  and gets the functions they are provided.

mrnuke:
Then give them codenames, like _chupacabra, and _diablo. Much easier to remember than numbers ;)

include/libopencm3/stm32/common/crc_common_f03.h
```diff
@@ -0,0 +1,110 @@
+/** @addtogroup crc_defines
+
+ @author @htmlonly &copy; @endhtmlonly 2016 Cem Basoglu <cem.basoglu@web.de>
+
+ */
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2016 Cem Basoglu <cem.basoglu@web.de
```
fenugrec: missing ">" here too P-)

include/libopencm3/stm32/common/crc_v2.h
```diff
@@ -0,0 +1,110 @@
+/** @addtogroup crc_defines
+
+ @author @htmlonly &copy; @endhtmlonly 2016 Cem Basoglu <cem.basoglu@web.de>
+
+ */
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2016 Cem Basoglu <cem.basoglu@web.de
```
fenugrec: (my previous comment was on the outdated diff, sorry) - missing ">" here !

fenugrec:
@cemizm same here, squashing those commits may help getting this merged faster.

Regarding F0 : the F0x0 chips seem to lack the "CRC_POL" register, so the crc_set_polynomial() function will write to an unmapped address. I'm not sure how this should be fixed ?

Gussy:
Is there any interest in merging this? This would replace my local implementation for the F3.

I can rebase/test if there's interest?

karlp:
Sure, I just never got back to it after it was updated

Gussy:
@karlp this can be closed, #843 superseded this. thanks!

PR #634, opened 2016-03-12
lm4f: Enable FPU using the vector_chipset hook

mrnuke:
By order of his High Priesthood KARLP

### DIFF
--- a/lib/lm4f/vector_chipset.c
+++ b/lib/lm4f/vector_chipset.c
@@ -1,0 +1,29 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2016 Alexandru Gagniuc <mr.nuke.me@gmail.com>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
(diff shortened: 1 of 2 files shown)


karlp:
by order of high priesthood wtf?  :rage1:

lib/lm4f/vector_chipset.c
```diff
@@ -0,0 +1,29 @@
...
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+static void pre_main(void)
+{
+	/* Enable FPU */
+	SCB_CPACR |= SCB_CPACR_FULL * (SCB_CPACR_CP10 | SCB_CPACR_CP11);
+	/* Wait for store to complete */
+	__asm__("DSB");
+	/* Reset pipeline. Now the FPU is enabled */
```
karlp: not entirely sure I agree with the comments here.  Also, are these even necessary? the end of function will have the same effect surely?  No other parts are doing this.

fenugrec:
Note (semi related), the programming manual for STM32F3/F4 (PM0214) has an example of enabling the FPU with both DSB + ISB. All similar cores should probably be changed to something like that...

mrnuke:
I've hauled this code around for years in my projects. I don't remember the exact reason, but these things were needed to reliably not crash.

ChuckM:
Page 7-72 of the [ARM Technical Reference Manual](http://infocenter.arm.com/help/topic/com.arm.doc.100166_0001_00_en/arm_cortexm4_processor_trm_100166_0001_00_en.pdf) gives this guidance:

7.3.2     Enabling the FPU
---------------------------------

Example code sequence for enabling the FPU in both privileged and user modes. The processor must be in privileged mode to read from and write to the CPACR.

Enabling the FPU
```
; CPACR is located at address  0xE000ED88
    LDR.W   R0, 0xE000ED88
; Read CPACR
    LDR     R1, [R0]
; Set bits 20-23 to enable CP10 and CP11 coprocessors
    ORR     R1, R1, #0xF << 20)
; Write back the modified value to the CPACR
    STR     R1, [R0]
'''

It does not mention any need for the DSB barrier instruction.

karlp:
This was merged back in 6f30e76c61ceab409302117f1bb66d72f43a0de4 sorry I didn't update and close it then.

PR #637, opened 2016-03-19
STM32 desig.c cleanup and Arch ARM GCC 5.3.0 bug workaround

urjaman:
This is a take it or close it kind of thing, but I'm just documenting what I had to do to get on with my stuff.

The archlinux arm gcc gives a "internal compiler error: in expand_expr_addr_expr_1, at expr.c:7736"
with that array version :(

### DIFF
--- a/lib/stm32/desig.c
+++ b/lib/stm32/desig.c
@@ -24,25 +24,26 @@ uint16_t desig_get_flash_size(void)
 	return DESIG_FLASH_SIZE;
 }
 
-void desig_get_unique_id(uint32_t result[])
+void desig_get_unique_id(uint32_t *result)
 {
-	result[0] = DESIG_UNIQUE_ID2;
-	result[1] = DESIG_UNIQUE_ID1;
-	result[2] = DESIG_UNIQUE_ID0;
+	*result++ = DESIG_UNIQUE_ID2;
+	*result++ = DESIG_UNIQUE_ID1;
+	*result = DESIG_UNIQUE_ID0;
 }
 
 void desig_get_unique_id_as_string(char *string,
 				   unsigned int string_len)
 {
 	int i, len;
-	uint8_t device_id[12];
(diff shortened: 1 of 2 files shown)


kuldeepdhaka:
i think, this should be merged even if the compiler do not give error. :+1:

urjaman:
oh you should also change the header if you take this, apparently i was a bit fast with this and gcc didnt warn... (or i can do the change if wanted, but trivial.)

lib/stm32/desig.c
```diff
@@ -24,7 +24,7 @@ uint16_t desig_get_flash_size(void)
 	return DESIG_FLASH_SIZE;
 }
 
-void desig_get_unique_id(uint32_t result[])
+void desig_get_unique_id(uint32_t *result)
 {
 	result[0] = DESIG_UNIQUE_ID2;
```
ChuckM: So why not then write this as 
```
    *result++ = DESIG_UNIQUE_ID2;
    *result = DESIG_UNIQUE_ID1;
```

urjaman:
Changed it to be a bit more generic cleanup while working around the bug at the same time.,
comments?

ChuckM:
Looks good to me :+1:

fenugrec:
Also +1, the original code casting " uint8_t device_id[]" to a (uint32_t *) is evil.

karlp:
I just want to check that the resulting serials are the same, but yes, this is probably the right thing to do :)

karlp:
serials are the same,but the code gets bigger? wat?
```
   text	   data	    bss	    dec	    hex	filename
   6172	     24	    532	   6728	   1a48	old/usbmidi.elf
   6204	     24	    532	   6760	   1a68	new/usbmidi.elf
```

karlp:
whatever, not going to even chase that, merged in d3fff11 thanks!

PR #639, opened 2016-03-27
fix timer trigger selection names

tsaitgaist:
correct TIMx_SMCR TS[2:0]: Trigger selection name are TI1F_ED, TI1FP1, TI2FP2 according to RM0008 and RM0090


karlp:
merged via rebase after fixing up summary line in c759ba5 thanks!

Issue #451, opened 2015-05-23
read modify write cycle to be avoided as per doc


kuldeepdhaka:
As per documentation, Read modify write for ISTR and Endpoint in usb not recommended.
(p861, p868 RM0091)

fenugrec:
I agree, but I doubt that commit you guys added will fix the issue. You're still doing a variant of RMW by doing
1) tempvar = USB_ISTR (or USB_EP)
2) clear one bit
3) write USB_ISTR or USB_EP.

For USB_ISTR, something like this is needed :
` USB_ISTR = ~(USB_ISTR_XYZ); `
(note : this writes 1 in some reserved bits on F0, so it needs to be refined slightly)

For USB_EP, it's more complicated since there are "rw" and "toggle" bits we need to avoid as well as the "rc_w0" flags we're interested in. As it turns out, I think the macros of libopencm3 are **already** doing this properly. The desired operation is thus:
```
/* clearing TX_CTR : */
uint32_t ep_temp = USB_EP(addr) & 0x8F8F & USB_EP_TX_CTR;	//clear "t" bits, keep "rw" bits
ep_temp |= USB_EP_RX_CTR;	//do not clear the other CTR bit (also "rc_w0")
USB_EP(addr) = ep_temp;

/* clearing RX_CTR : */
uint32_t ep_temp = USB_EP(addr) & 0x8F8F & USB_EP_RX_CTR;	//clear "t" bits, keep "rw" bits
ep_temp |= USB_EP_TX_CTR;	//do not clear the other CTR bit (also "rc_w0")
USB_EP(addr) = ep_temp;
```

karlp:
what commit are you referring to?

fenugrec:
It's a commit kuldeep and gmc added to their repos that claim to fix the issue, but as you pointed out in PR #641 there's no issue after all. Well only the issue of writing 1 to reserved bits in ISTR, not RMW.

kuldeepdhaka:
?? @fenugrec please explain what you are "saying".

> It's a commit kuldeep and gmc added to their repos that claim to fix the issue

gmc repo repo referenced the issue earlier, probebly both are my commit.

> but as you pointed out in PR #641 there's no issue after all

did i say anything in #641 ?

fenugrec:
@kuldeepdhaka let me rephrase :
- yes, it's your commit in gmc's tree too. No, you didn't say anything in #641 , but karlp did, he pointed out I was wrong in my comments in 641.
[...]

PR #641, opened 2016-04-02
stm32: usb: redefine the USB_CLR_ISTR_* macros to avoid R-M-W.

fenugrec:
This is my proposed fix for GH issue #451 .
The current implementation of st_usbfs_poll() invokes the USB_CLR_ISTR_* macros to clear some flags in the USB_ISTR register, but those macros imply a read-modify-write cycle.

The macro definitions had to be moved from st_usbfs_common.h to the version-specific st_usbfs_v*.h because the USB_ISTR register has different "Reserved" bits according to the peripheral version.
To avoid a read-modify-write cycle when clearing the bits, the USB_CLR_ISTR_* macros now simply write '1' to every defined bit except the one being cleared.


include/libopencm3/stm32/common/st_usbfs_v1.h
```diff
@@ -36,6 +36,30 @@
...
+					USB_ISTR_SUSP | \
+					USB_ISTR_RESET | \
+					USB_ISTR_SOF | \
+					USB_ISTR_ESOF | \
+					USB_ISTR_DIR | \
+					USB_ISTR_EP_ID)		/* these bits are all "rc_w0" : writing 1 is safe */
+
+/* --- USB interrupt status register manipulators -------------------------- */
+
+/* Note: CTR and DIR are read only! */
```
karlp: and EP_ID, if we're going there ;)

karlp:
Please explain to me how the existing CLR_REG_BIT macro implies a RMW?  The only difference after this change is that instead of writing all 1s, except the bit to clear (and therefore writing 1s to the "reserved" and "read only" bits, you're no writing 0s for the reserved and read only bits.  Arguably the 0 for the reserved is better, but it's a (very) slim case, and it's not the case you're making in the description here.  writing 1s or 0s for the read only bits is equivalent.

CLR_REG_BIT_MSK_AND_SET does a RMW, yes, but that's not what's used here.

fenugrec:
This is embarassing, for some reason I thought I saw  " &= ~BIT" in the expansion of the USB_CLR_ISTR_* macros.
And thus you're right, the only real difference this PR makes is writing 0 to the reserved bits.

karlp:
Do you still want any part of this?

fenugrec:
I cheated and reworded everything - the PR name + description and the commit message.
It's now more of a "strictly-correct" fix rather than a bug fix, so no hurry to merge.

PR #643, opened 2016-04-06
libopencmsis is buggy (missing volatiles _at least_ )

stoyan-shopov:
Without the 'volatile' qualifier the gcc compiler generates code that does not work as intended.

### DIFF
--- a/lib/stm32/common/adc_common_v1.c
+++ b/lib/stm32/common/adc_common_v1.c
@@ -588,16 +588,16 @@ void adc_set_regular_sequence(uint32_t adc, uint8_t length, uint8_t channel[])
 		if (i <= 6) {
 			first6 |= (channel[i - 1] << ((i - 1) * 5));
 		}
-		if ((i > 6) & (i <= 12)) {
+		if ((i > 6) && (i <= 12)) {
 			second6 |= (channel[i - 1] << ((i - 6 - 1) * 5));
 		}
-		if ((i > 12) & (i <= 18)) {
+		if ((i > 12) && (i <= 18)) {
 			third6 |= (channel[i - 1] << ((i - 12 - 1) * 5));
 		}
-		if ((i > 18) & (i <= 24)) {
+		if ((i > 18) && (i <= 24)) {
 			fourth6 |= (channel[i - 1] << ((i - 18 - 1) * 5));
 		}
-		if ((i > 24) & (i <= 28)) {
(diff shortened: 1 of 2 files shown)


karlp:
Are you using efm32? I'd rather see this systick made to follow it's own commentary and use the existing cm3/systick.h routines now.  they've long since been pulled out of stm32 specific code.  Also, this set of headers arguably has zero purpose these days, now that the original CMSIS headers have "proper" " more suitable" licenses.   While your fix may be correct, I'm concerned that the entire use case could be improved, rather than fixing a symptom down the road.

stoyan-shopov:
I am using stm32, f0, the actual part is stm32f070rb. I needed some precise timekeeping for an otherwise simple project, so I got down to polling the iron directly, which does not work without 'volatile'.
I thought someone else might have stumbled on the same rake, so I put it here.
Which headers do you mean? Do you recommend using CMSIS directly from ARM? I haven't taken a look on those for ages...

Thanks

karlp:
I mean, use libopencm3/cm3/systick headers, rather than libopencmsis/*.  the libopencmsis headers were written to work with some efm32 rom code that required them.   If you _want_ to use cmsis headers, you can use the original ones from ST/ARM now, as they're more permissively licensed.

karlp:
perhaps https://github.com/libopencm3/libopencm3-examples/blob/master/examples/stm32/f1/other/systick/systick.c might help?

fenugrec:
if "libopencmsis"  is broken (definitely missing a volatile there, I agree), or not relevant anymore, I think they should be either fixed, or removed altogether ? Or at the very least, warn of possible brokenness / irrelevance and point to the recommended alternatives !

quitte:
We discussed wether opencmsis should be replaced with cmsis quite a while ago. I did give it a try and this is what came out of it. I barely remember any details - but whatever I did was pretty straight-forward.
That being said - here it is for reference.
https://github.com/quitte/libopencm3-cmsis

karlp:
I've retitled this to at least keep all the discussion on this issue here, rather than closing this and making an empty issue.  I'm not super keen on merging this, it would look like I endorsed the usage of this awful libopencmsis stuff.

fenugrec:
704 commits !? You know this is never going to be merged, right...

jhalmen:
Isn't this missing a rebase or something?

karlp:
this is _exactly_ why people shouldn't send PRs from their own master branch :)

PR #645, opened 2016-04-07
stm32f7: fix build failure when CFLAGS is set

tidklaas:
The Makefile for the stm32f7 library still used CFLAGS instead of TGT_CFLAGS, leading to a build failure when passing CFLAGS as a make option.


karlp:
please rebase this properly to have just the commit you need for this.

tidklaas:
Sorry, fixed now. I should not touch git after 6pm...

karlp:
np, thanks for cleaning it up, saves me work :)  this is now merged, thanks!

PR #651, opened 2016-04-11
Fixed documentation for 8 bit parameters in stm32/f0 usart routines.

ha-ko:


karlp:
I'd actually rather see these simply dropped.  The type information is already available from the functon signature, adding it in the comments in this cases provides no additional value, and only serves to be out of date.

Further, commits headlines should be something like, "module: summary" so a good rewording of this commit would be "stm32f0: usart: drop incorrect/confusing documentation"

ha-ko:
Ah, good point, I just stumbled across the conflicting doc for the usart and thought I make a quick fix. 
So all unnecessary type descriptions in the stm32f0 documentation should be removed?

Thanks for the feedback and the commit msg style heads up!

fenugrec:
Problem : the F0 USART can use 9 bits of the USARTx_RDR register ( the extra bit being parity), I assume this is why the other API functions use uint16_t arguments...

ha-ko:
This caught my eye yesterday as well.
Is there a reason why the F0 functions are the only ones using 8 instead of 16 bit?

karlp:
Going to fix both issues.  Dropping all the redundant type info from doxygen, and also fixing all the definitions for the f0 usarts.

karlp:
this has now landed in 6dd7b3e and 7ee1d94 thanks!

Issue #460, opened 2015-06-01
Merge of USB host library


amirhammad:
I would be happy to get the [usb host library](https://github.com/libusbhost/libusbhost) merged into libopencm3. 

This issue was opened with intent to discuss whether and possibly when and how could it be done.

ChuckM:
It is an interesting question Amir, I've been developing a bunch of higher level APIs for things like SDIO and graphics which I've been building into a monolithic mini sort of OS for my projects. I've added yours to get some experience with it (most of my work is with the STM32F4 series anyway) and it lined up with my future need to add a mouse and keyboard input.

But the challenge with adding it to libopencm3 directly is that, not all that many variants have USB-OTG peripherals. The Tiva C does, the STM32F4 does, and I believe the SAM4 does but those are all Cortex-M4 variants. For one offs it makes sense to just link in an additional library.

amirhammad:
@ChuckM The library was meant to be as multiplatform as it could be, but as of this time, only the low level driver for stm32f4 was written. I think, developing usb device drivers (for example: mouse, keyboard drivers) is easier when it is independent of the underlying hardware (OTG in case of stm32f4). However, USB-OTG is present in stm32f4 f2 and f1 (f105 or something), device drivers could be written independently of the low level implementation. 

The case could be: The programmer wants to use some special platform for which low-level driver does not exist yet. He, then, instead of writing his own low+high-level driver, could write only low level driver and then enjoy all the features device drivers contained in libopencm3 offer.

amirhammad:
I think this issue is no more considered and so I close it.

jhgorse:
@ChuckM @karlp @gsmcmullin would it make sense to make a note about the scope of opencm3 at the top level so that folks can know that it is not intended for USB Host stack or other things which negatively affect interoperability?

Also, it's been a while. I am glad to see this is still moving forward. I read through some of the USB device code stack and it looks quite nice. =)

Cheers,
Joe

karlp:
We've got nothing against bringing host support in, but it's not going to happen via some random issue pointing at some external code.  :)  I'm certainly not going out to external repos and restructuring them to bring them in.

Issue #401, opened 2015-01-24
for ARM (on STM32) usb descriptor and usbd_control_buffer alignment


kuldeepdhaka:
struct usb_device_descriptor and uint8_t usbd_control_buffer should atleast be 16bit aligned.
these are used to by 16 oriented copy operation for STM32 usb PMA.
if not, on ARM hard-fault occur due to aligned memory access constraint.
(Please Document this somewhere to save someone from headache).

fenugrec:
+1000 !, just got hit by this. I think this can be generalized to all ST "usbfs_v2" peripherals (F0 and L0) since both M0 and M0+ cores do not support unaligned memoery access ?

karlp:
Fixed via merge of https://github.com/libopencm3/libopencm3/pull/653

PR #653, opened 2016-04-15
tests: gadget 0: added test for unaligned buffer read/writes.

fenugrec:
[STATUS - this is ready to merge to the best of my knowledge]

This adds a test to the gadget0 framework to test unaligned vs aligned buffer access in the usb stack, and the corresponding fix.
Without the fix, this fails on stm32F072, which is expected (all F0 / L0 mcus should fail) but not normal. Note that once the unaligned test fails, the target is in HardFault and needs to be reset.

See issues #401, #460.

I am only able to test on a F072 target, it would be nice if someone could test on something other than F0* / L0*, for example F1 which should not exhibit the issue.

The fix is copied from Kuldeep's tree; I haven't modified the st_usbfs_copy_to_pm() function however,  since it's already a bytewise copy.

### DIFF
--- a/tests/gadget-zero/test_gadget0.py
+++ b/tests/gadget-zero/test_gadget0.py
@@ -318,3 +318,54 @@ class TestControlTransfer_Reads(unittest.TestCase):
         self.assertEqual(len(q), 10, "In this case, should have gotten wLen back")
 
 
+class TestUnaligned(unittest.TestCase):
+    """
+    M0 and M0+ cores don't support unaligned memory accesses. These test
+    how the stack behaves with aligned vs unaligned buffers.
+    https://github.com/libopencm3/libopencm3/issues/401
+    https://github.com/libopencm3/libopencm3/issues/461
+    """
+
+    def setUp(self):
+        self.dev = usb.core.find(idVendor=0xcafe, idProduct=0xcafe, custom_match=find_by_serial(DUT_SERIAL))
+        self.assertIsNotNone(self.dev, "Couldn't find locm3 gadget0 device")
+
+        self.cfg = uu.find_descriptor(self.dev, bConfigurationValue=2)
+        self.assertIsNotNone(self.cfg, "Config 2 should exist")
+        self.dev.set_configuration(self.cfg);
+        self.req = uu.CTRL_OUT | uu.CTRL_TYPE_VENDOR | uu.CTRL_RECIPIENT_INTERFACE
+        self.intf = self.cfg[(0, 0)]
+        # heh, kinda gross...
+        self.ep_out = [ep for ep in self.intf if uu.endpoint_direction(ep.bEndpointAddress) == uu.ENDPOINT_OUT][0]
+        self.ep_in = [ep for ep in self.intf if uu.endpoint_direction(ep.bEndpointAddress) == uu.ENDPOINT_IN][0]
+
+    def tearDown(self):
+        uu.dispose_resources(self.dev)
+
+    def set_unaligned(self):
+        # GZ_REQ_SET_UNALIGNED
+        x = self.dev.ctrl_transfer(self.req, 4, 0, 0)
+
+    def set_aligned(self):
+        # GZ_REQ_SET_ALIGNED
+        x = self.dev.ctrl_transfer(self.req, 3, 0, 0)
+
+    def do_readwrite(self):
+        """
+        transfer garbage data to/from bulk EP; alignment issues will hardfault the target
+        """
+        data = [x for x in range(int(self.ep_out.wMaxPacketSize / 2))]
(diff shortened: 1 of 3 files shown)


tests/gadget-zero/test_gadget0.py
```diff
@@ -318,3 +318,56 @@ def test_mean_limits(self):
...
+        self.assertEqual(len(data), read_size, "Should have read as much as we asked for")
+
+    def test_aligned(self):
+        self.set_aligned()
+        self.do_readwrite()
+
+    def test_unaligned(self):
+        self.set_unaligned()
+        self.do_readwrite()
+        self.set_aligned()
```
karlp: shouldn't need this last line

tests/gadget-zero/usb-gadget0.c
```diff
@@ -159,7 +162,7 @@ static const struct usb_config_descriptor config[] = {
 	}
 };
 
-static char serial[] = "0123456789.0123456789.0123456789";
+static char serial[] = "stm32f072disco";
```
karlp: why did you need to do this?

tests/gadget-zero/usb-gadget0.c
```diff
@@ -181,37 +184,60 @@ static struct {
...
 static void gadget0_ss_in_cb(usbd_device *usbd_dev, uint8_t ep)
 {
 	(void) usbd_dev;
 	trace_send_blocking8(0, 'I');
-	uint8_t buf[64];
+	uint16_t buf[BULK_EP_MAXPACKET / 2 + 1];
+	uint8_t *src;
+
+	if (test_unaligned) {
+		src = ((uint8_t *) buf) + 1;
```
karlp: how do you know that it was aligned before? is that because you used a uint16_t buf that must get aligned?  Why not just use __attribute__(aligned(4)) on the uint8_t buffer, and +1 there?  more obvious what's going on?

tests/gadget-zero/usb-gadget0.c
```diff
@@ -279,6 +311,7 @@ static void gadget0_set_config(usbd_device *usbd_dev, uint16_t wValue)
 	ER_DPRINTF("set cfg %d\n", wValue);
 	switch (wValue) {
 	case GZ_CFG_SOURCESINK:
+		test_unaligned = 0;
```
karlp: can this just go into the state structure? much like the pattern_counter?

tests/gadget-zero/usb-gadget0.c
```diff
@@ -181,37 +184,60 @@ static struct {
 	.pattern_counter = 0,
 };
 
+/* aligned / unaligned test flag.
+ * Using a bool type for this instead, in some cases, can cause gcc to place other critical buffers
```
karlp: what?

tests/gadget-zero/usb-gadget0.c
fenugrec: didn't cross my mind; of course that line won't be part of the final PR.

tests/gadget-zero/usb-gadget0.c
fenugrec: I was avoiding gcc-specific directives, and uint16_t should guarantee alignment on M0/M0+... but I guess it's not that much clearer this way. I think I'll change it to aligned(2)

tests/gadget-zero/usb-gadget0.c
fenugrec: Certainly, I wasn't sure if I could do that or if the state structure was reserved for the linux-gadget-compatible functionality.

tests/gadget-zero/usb-gadget0.c
fenugrec: That's what caused my problems. I had a global bool, that gcc allocated as a uint8 , and in certain conditions other uint8_t arrays were allocated right after, with no padding, causing an unaligned control buffer and associated issues.

tests/gadget-zero/usb-gadget0.c
karlp: no, not at all, I'm only aiming for functional equiv, not code equiv

tests/gadget-zero/usb-gadget0.c
```diff
@@ -176,42 +179,60 @@ static usbd_device *our_dev;
 static struct {
 	uint8_t pattern;
 	int pattern_counter;
+	int test_unaligned;	/* If 0 (default), use 16-bit aligned buffers. This should not be declared as bool */
```
karlp: comment no longer relevant as you're using attributes on buffer

karlp:
:+1:

karlp:
Thanks for this! merged with some splits and rewords

Issue #654, opened 2016-04-16
stm32f0: ADC calibration never finishes when adc_power_off is called before


DanielG:
The following code will hang forever in `adc_calibrate_wait_finish`:

```
rcc_periph_clock_enable(RCC_ADC);
adc_power_off(ADC);
adc_calibrate_start(ADC);
adc_calibrate_wait_finish(ADC);
```

whereas this works just fine:

```
rcc_periph_clock_enable(RCC_ADC);
adc_calibrate_start(ADC);
adc_calibrate_wait_finish(ADC);
```

RM0091 says to "ensure ADCEN=0" which is implemented in the example code from the appendix as follows:

```
/* (1) Ensure that ADEN = 0 */
/* (2) Clear ADEN */
/* (3) Launch the calibration by setting ADCAL */
/* (4) Wait until ADCAL=0 */
if ((ADC1->CR & ADC_CR_ADEN) != 0) /* (1) */
{
    ADC1->CR &= (uint32_t)(~ADC_CR_ADEN); /* (2) */
}
ADC1->CR |= ADC_CR_ADCAL; /* (3) */
while ((ADC1->CR & ADC_CR_ADCAL) != 0) /* (4) */
{
    /* For robust implementation, add here time-out management */
}
```

Looking at `adc_power_off_async` it seems to me that maybe it should check if ADEN=1 before trying to power down the ADC by setting ADDIS or maybe `adc_calibrate_start` should just force ADEN=0 like the example code does?

For easy reference:
```
void adc_power_off_async(uint32_t adc)
{
	uint32_t checks = ADC_CR_ADSTART;
	uint32_t stops = ADC_CR_ADSTP;
#if defined (ADC_CR_JADSTART)
	checks |= ADC_CR_JADSTART;
	stops |= ADC_CR_JADSTP;
#endif
	if (ADC_CR(adc) & checks) {
		ADC_CR(adc) |= stops;
		while (ADC_CR(adc) & checks);
	}
	ADC_CR(adc) |= ADC_CR_ADDIS;
}
```

(I'm testing this on a STM32F042C6T6)

I'd be happy to submit a patch if someone can tell me what the preferred fix is.

kungjohan:
2016-04-16 17:06 GMT+02:00 Daniel Gröber <notifications@github.com>:

> The following code will hang forever in adc_calibrate_wait_finish:
>
> rcc_periph_clock_enable(RCC_ADC);
> adc_power_off(ADC);
[...]

PR #662, opened 2016-05-26
stm32: adc: Fix adc_power_off_async for already disabled adc

ha-ko:
The ADDIS bit is only allowed to be set by software if the ADC is
already enabled (ADEN=1).

See page 259 of STM32F0x1/STM32F0x2/STM32F0x8 datasheet:
"Software is allowed to set ADDIS only when ADEN=1 and ADSTART=0 (which ensures that no
conversion is ongoing)"

ADSTART=0 is already ensured.

adc_power_off(..) without that fix lead to failing calibration with an already disabled ADC.

The example 

/examples/stm32/f0/stm32f0-discovery/adc/adc.c

is fixed by that as well.

### DIFF
--- a/lib/stm32/common/adc_common_v2.c
+++ b/lib/stm32/common/adc_common_v2.c
@@ -116,7 +116,8 @@ void adc_power_off_async(uint32_t adc)
 		ADC_CR(adc) |= stops;
 		while (ADC_CR(adc) & checks);
 	}
-	ADC_CR(adc) |= ADC_CR_ADDIS;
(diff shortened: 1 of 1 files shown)


ha-ko:
Just saw that this has already been an open issue found by DanielG: 
https://github.com/libopencm3/libopencm3/issues/654

karlp:
Fixed (a different way) in c2e0afd  You can't call stop on it either if it's not enabled.  Just added it before the start.
 thanks for chasing me on this to get it merged

PR #646, opened 2016-04-08
karl-next (devices.data, lm3s, samd)

karlp:
my current work tree..... for those interested....

PR #664, opened 2016-06-21
Make scripts/genlink.awk compatible with mawk by quoting metachars

delamonpansie:
Trying to build project with opencm3 on ubuntu fails on ld sctipt generation because of mawk:
```
$ awk -v PAT="${DEVICE}" -v MODE="DEFS" -f ${OPENCM3_DIR}/scripts/genlink.awk ${OPENCM3_DIR}/ld/devices.data
mawk: libopencm3/scripts/genlink.awk: line 34: regular expression compile failed (missing operand)
?
mawk: libopencm3/scripts/genlink.awk: line 35: regular expression compile failed (missing operand)
*
mawk: libopencm3/scripts/genlink.awk: line 36: regular expression compile failed (missing operand)
+

```

mawk is default awk on ubuntu

### DIFF
--- a/scripts/genlink.awk
+++ b/scripts/genlink.awk
@@ -31,9 +31,9 @@ BEGIN {
 	gsub(/\r$/,"");
 
 	tmp = "^"$1"$";
-	gsub(/?/, ".", tmp);
-	gsub(/*/, ".*", tmp);
-	gsub(/+/, ".+", tmp);
+	gsub(/[?]/, ".", tmp);
+	gsub(/[*]/, ".*", tmp);
+	gsub(/[+]/, ".+", tmp);
 	tolower(tmp);
 
 	if (PAT ~ tmp) {


karlp:
There are already some staged changes and overhauls on this script, please check if the work in https://github.com/libopencm3/libopencm3/pull/646 already handles these cases for you.

delamonpansie:
No, #646 doesn't fix mawk compat.

karlp:
right, ok.  got this here too.  you can also just \ escape the chars.

delamonpansie:
Sure, you can use `\`.
Also, please consider removing `2>/dev/null` from genlink-config.m. Errors will give an idea about that something is gone wrong instead of giving misleading error `libopencm3/mk/genlink-rules.mk:22: *** unknown device stm32f103c8t6 for the linker. Cannot generate ldscript.`
Thank you.

brabo:
I believe a fix was proposed here https://github.com/libopencm3/libopencm3/pull/397 but was lacking more details about the exact problem it intended to fix.

delamonpansie:
Yes, #397 fixes exactly same problem. Also, mawk is default awk on Ubuntu too.

karlp:
fixed in 0cb1db0, thanks for chasing this, and sorry we hadn't handled that earlier

karlp:
removing the silencing of errors is a good idea, and was unecessary anyway, as the $() only captures stdout anyway. I've put that patch in my local staging tree, thanks for the suggestion.

PR #665, opened 2016-06-24
Fix wrong function name in example

DrLuke:
I found this to be wrong while looking at the code.


karlp:
thanks! merged in 18e15c1

PR #667, opened 2016-07-04
stm32/f4: add QUADSPI includes

ChuckM:
This commit adds a qspi.h file for the f4 series and provides #defines for
the registers to configure it. It also updates the rcc clk enum with RCC_QSPI
and memorymap with the base address for the QSPI memory and control registers.


ChuckM:
Just a note on the GET and SET macros, they do two things, one they insure that the shift/mask rule is adhered to (they don't work if it isn't) and 'get' takes the register value and extracts the field, 'set' takes the field and puts in in a place to be inserted. Using SET with all 1's (-1) creates a mask you can use to erase a previous version or test against value.

include/libopencm3/stm32/f4/memorymap.h
```diff
@@ -137,7 +138,21 @@
...
 
 /* AHB3 */
+/* Address: 0x60000000 */
+#define FMC_BANK1			(PERIPH_BASE_AHB3)
+/* Address: 0x70000000 */
+#define FMC_BANK2			(PERIPH_BASE_AHB3 + 0x10000000U)
+/* Address: 0x80000000 */
+#define FMC_BANK3			(PERIPH_BASE_AHB3 + 0x20000000U)
+/* Address: 0x90000000 */
+#define QSPI_BANK2			(PERIPH_BASE_AHB3 + 0x30000000U)
```
karlp: Is there meant to be a QSPI_BANK1 somewhere? or just 2?

include/libopencm3/stm32/f4/qspi.h
```diff
@@ -0,0 +1,165 @@
+/*
+ * vim: set tabstop=8 noexpandtab:
+ * quad SPI defines 
+ */
```
karlp: Drop the vim comments, and include the regular libopencm3 header here.

karlp:
Not sure I agree entirely with the macros reducing code, seems to just add code at the moment :)

Also, not sure if GET_ and SET_ are the right prefix, why not use QSPI_ for the "helper" macros?

include/libopencm3/stm32/f4/memorymap.h
ChuckM: Nope, just the one bank. Fixed.

include/libopencm3/stm32/f4/qspi.h
ChuckM: Done and Done.

ChuckM:
I'm not married to the GET/SET prefixes, I can share with you my reasoning on how I came to use them which may (or may not :-) help here. 

From an English perspective they refer to "GETting the field bits from a register" or "SETting the field bits in the register." From an OOP perspective they are similar to 'getter' and 'setter' functions for changing fields in an object. I've tried a couple of variants and these seem to work well for me in the code. 

As you know, the ST folks in their Standard Peripheral library were using bit fields to do this. Making initializing and changing flags a bit heavy weight (to my way of thinking).  Using the macros was both lighter weight (fewer lines of code to set or read bits) and more comfortable for me to code.

And finally there are a number of bit fields already in LOC3 where the idiomatic way of accessing the bits are to use the _SHIFT and _MASK defines to manipulate the register contents (or a copy of a register) to get to the bit field. In the comments (when they are there) those code segments are often referred to as "getting" the bits, or "setting" the bits of the register.

Given that I felt GET/SET were the prefixes that the most programmers would understand the intent of the macros.

ChuckM:
Karl, I've removed the use of the _GET and _SET macros in the include file. I rebuilt my driver where I switched to using them in my code (so something like `QUADSPI_SET(CCR, INST, 0x95)` rather than `SET_QUADSPI_CCR_INST(0x95)`. I also discovered some typos in the use of MMIO32 that I hadn't caught.

include/libopencm3/stm32/f4/qspi.h
```diff
@@ -118,66 +108,55 @@
...
+/* MODE values */
+#define QUADSPI_CCR_MODE_NONE		0
+#define QUADSPI_CCR_MODE_1LINE		1
+#define QUADSPI_CCR_MODE_2LINE		2
+#define QUADSPI_CCR_MODE_4LINE		3
+
 
 /* QUADSPI address register */
-#define QUADSPI_AR			MMIO32(QUADSPI_BASE, 0x18U)
+#define QUADSPI_AR			MMIO32(QUADSPI_BASE + 0x18U)
```
karlp: nice catch!

karlp:
looks good.  One last thing,  the files are "qspi.h" but the defines are QUADSPI, except for the ones that are QSPI.  Do we want to have QSPI everywhere? or QUADSPI everywhere?

karlp:
RM0386 uses QUADSPI for everything except the rcc enable bit, so I'm leaning towards the files being quadspi as well...

ChuckM:
Per the datasheet and the history of LOC3 I believe QUADSPI is the correct answer.  I can change the RCC constant as well as the QSPI-BANK in the memorymap file.

ChuckM:
And now squashed down to a single commit.

karlp:
The file is still called qspi.h though?

ChuckM:
yes it is.

ChuckM:
It really needed the FMODE defines.

ChuckM:
renamed per the irc chat

karlp:
A quick skim indicates that this is identical to the Quadspi in the L4 family.  Do we want to make ita common file right away? or should we land this first then move it?

ChuckM:
They are the same peripheral, my test/example code works on the L476 disco board as well.

karlp:
merged minus the prerpocessor mangling register helpers.  We can pull it up to a common file for l4 later. thanks for this!

PR #669, opened 2016-07-06
 	clarified automatic linker script makefile requirements.

Jak-o-Shadows:
As is, the example Makefile will not work. Must also define SRCLIBDIR so that the awk scripts actually work.

Not entirely sure how pull requests work...


karlp:
I believe this is already addressed in https://github.com/libopencm3/libopencm3/pull/646 ? which is the "next" branch

PR #671, opened 2016-07-07
[cm3] reset that stm32-f4

H2OBrain:
this fix addresses what was discussed in this old thread [[libopencm3-devel] system reset?](https://sourceforge.net/p/libopencm3/mailman/libopencm3-devel/thread/CAHTjPirViWhJud35%2BM6VyUKxV5TrG1E1Nt24uPT3LLj6EFXUAw%40mail.gmail.com/#msg31887448)

FIXES
- changing the priority grouping in the SCB_AIRCR register along with setting the the SYSRESETREQ flag keeps the cpu from resetting
- the additional dsb instruction synchronises the memory accesses before/during the system reset
- __i did not test this for VECTRESET but assumed it would be the same__

### DIFF
--- a/lib/cm3/scb.c
+++ b/lib/cm3/scb.c
@@ -25,16 +25,26 @@
 #if defined(__ARM_ARCH_7M__) || defined(__ARM_ARCH_7EM__)
 void scb_reset_core(void)
 {
-	SCB_AIRCR = SCB_AIRCR_VECTKEY | SCB_AIRCR_VECTRESET;
-
+	__asm__ volatile ("dsb");
+	SCB_AIRCR = (SCB_AIRCR_VECTKEY
+#if defined(SCB_AIRCR_PRIGROUP_MASK)
+			  | (SCB_AIRCR & SCB_AIRCR_PRIGROUP_MASK)
+#endif
+			  |  SCB_AIRCR_VECTRESET);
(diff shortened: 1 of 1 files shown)


karlp:
Could you try rewording this to explain a little better what this is fixing?

karlp:
Have you got some test case where this isn't working?  I've got a button press to trigger scb_reset_system, and even if I have priority groups set, it still happily resets.  what platforms are you seeing this on?

ChuckM:
Hi @H2OBrain I went back and that code now works with the current scb_system_reset() code. My best guess at what was happening when I wrote that was that the gcc compiler was optimizing out the call to set the SCB register assuming it had no effect.

H2OBrain:
Ah ok. So it is not working for me at the moment because I am optimising the libopencm3 code with some Makefile hacks wit -O3.

karlp:
@H2OBrain but it still shouldn't be optimized out.  can you share your example at all?  have a look at the dissambly and confirm if that's the case, I'm happy to look at this, but I need more evidence of what's going on first.

H2OBrain:
Sorry guys
I could not reproduce the behaviour where the execution got stuck in the while(1); loop in the scb_reset_core/scb_reset_system functions. I must have done something else wrong..
The next pull request I will test thoroughly again. :)

Issue #668, opened 2016-07-04
usb stack doesn't recover from timeouts.


jhoenicke:
Our device sometimes does long computations, in which it currently doesn't call usbd_poll() regularly.  If at this time the usb hid devices are enumerated the OS sends `GET_DESCRIPTOR STRING` requests which timeouts since the usb handler is not called.   Unfortunately afterwards the device will never recover.  Even when usbd_poll is then called regularly, the `GET_DESCRIPTOR STRING` requests are not answered properly.  They also cause long delays in the USB communication.  This causes the device to become almost unusable.

I tried to make some small test that exhibits the problem (you need to adapt it to your hardware though). You need some program on the PC side that causes the STRINGs to be requested while  the device is busy ("usb is off").  These will timeout but also the requests that are sent when the device is not busy anymore ("usb is back") will timeout.

https://gist.github.com/jhoenicke/13dd9d32ee03488993805f122bb840fd

The original bug report is here: trezor/trezor-mcu#98

zyp:
The libopencm3 USB stack itself shouldn't have any requirements on how often usb_poll() needs to be called, but it obviously can't do any work while the device is busy with other stuff.

Most USB control requests have a timeout of five seconds (the exceptions to this have even stricter timeouts). When a requests times out, the host controller gives up waiting for the data and leaves the host software to deal with the situation. A device that let standard control requests time out may be violating the USB specification, and the host stack may judge that such a device is defective and give up trying to talk to it.

Are your long computations part of the startup procedure? If so, you could leave the device in soft detach mode while busy, so the host won't start talking to it until it's ready to answer.

jhoenicke:
The long computation part is when the user entered his password, so it is in the middle of the process.  I suspect that the device isn't properly replying even when we later poll regularly, since according to usbmon the PC sends the requests but they don't get answered.  

I can also try to analyse the raw wire protocol with a logic analyzer...

jhoenicke:
Okay this happens when device is busy:

```
PC: GET DESCRIPTOR STRING 0 --- timeouts after 1 second.
PC: GET DESCRIPTOR STRING 0 --- timeouts after 1 second.
PC: GET DESCRIPTOR STRING 3
Device (which is no longer busy):  Answers with STRING 0!
PC: GET DESCRIPTOR STRING 0 --- timeouts after 1 second (even though usbd_poll is called)
PC: GET DESCRIPTOR STRING 0 --- timeouts after 1 second
[...]

Issue #600, opened 2016-01-17
[USB-OTG] Control FIFO out word loss


H2OBrain:
When receiving control out data in usb-otg (hs and fs) the data is shifted by 4 bytes (1 word) so that the first byte received is the actual fourth byte and the last four bytes are just garbage. 

Example : 

```
PC  CTRL_OUT  size: 11b  msg: 01234567891
MCU CTRL_OUT  size: 11b  msg: 4567891asdf
```

To fix this issue a delay was added in usb_fx07_common.c (see FIXME: Why is a delay needed here?)
This delay causes the whole USB device to crash when I run my adapted (control loopback test)[https://github.com/H2OBrain/libopencm3/commit/e57ca422d92bb7ed49487ffe08f10036c30eda56].
Without this delay, the device 'never' crashes but the first four bytes are always lost.
('never' means `hile [[ 1 ]]; do python test_gadget0.py; done` ran for 10h without stalling/crashing the device. The output was always '.........ss.F............', 'F' being the control-out loopback test failing)

There's an old thread on st's forum about this [short-link](https://my.st.com/b61ac19) or [long-link](https://my.st.com/public/STe2ecommunities/mcu/Lists/cortex_mx_stm32/Flat.aspx?RootFolder=/public/STe2ecommunities/mcu/Lists/cortex_mx_stm32/USB%20FS%20Losing%20word%20in%20RxFIFO%20%28STM42F40x%29&currentviews=1943) (needs login to st forums)
Jason Kotzin from this thread was also asking in the sf.net libopencm3 mailing list [link1](http://sourceforge.net/p/libopencm3/mailman/message/29912341/) [link2](http://sourceforge.net/p/libopencm3/mailman/message/30698153/)

I tried a couple things myself with no success so far.

@karlp : yes, the tests show that we have a big problem in the usb stack here. Maybe some of the who implemented usb in 2012 know more about this bug/quirk of usb-otg in stm32f4 devices..

karlp:
sounds like an _excellent_ case for the tests to try handling.

prusnak:
Is the issue fixed by fixes in https://github.com/libopencm3/libopencm3/pull/672 ?

H2OBrain:
@prusnak Yes it is fixed by #672.

alibabashack:
[...]

Issue #549, opened 2015-10-18
[Feature] STM32 USB OTG fifo-multi-packet-write feature and fifo-flush fix


H2OBrain:
This [example](https://github.com/H2OBrain/libopencm3-examples/tree/master/examples/stm32/f4/stm32f429i-discovery/usb_stresstest) demonstrates a proof-of-concept to enable fifo-multipacket-write feature and fixes for some bugs/missing features that are exposed when using multi-packet write.

It depends on [this](https://github.com/H2OBrain/libopencm3/tree/new_usb) libopencm3 branch.
The following files were changed in libopencm3
- include/libopencm3/stm32/otg_common.h
- lib/usb/usb_fx07_common.c
Commits
- [prepare for multi-packet fifo](https://github.com/H2OBrain/libopencm3/commit/42a538758fdc9e5aa612f8a9fc4f4a3932cd1ebf)
    - basic fifo-multipacket-support implementation which breaks OTG-FS support
    - (and is broken, use this seems to work..)
- [this seems to work..](https://github.com/H2OBrain/libopencm3/commit/143990c54591f6d88386b51063c92608e0e96ce4)
    - fixed definitions in otg_common.h
    - first working implementation
    - adds my own fifo-flush function to ep_reset (but commented out)
- [made maxpacketsize and usb-driver detection variable](https://github.com/H2OBrain/libopencm3/commit/2f8c07e792677474a41bc988699cb9d526058e3f)
    - gets wMaxPacketSize(ep) from the usbd_device in usb_fx07_common.c
        - not fully tested! (I don't know the usb-stack good enough)
-  [replaced usb-fifo ep-flush/disable scheme with one that works for cable-disconnect](https://github.com/H2OBrain/libopencm3/commit/3e849bfc9c1c182206619bc0eab922ce1e9aea54)
    - replaced fifo flush function in endpoint reset

BUGS
- `REBASE(OTG_GRSTCTL) = OTG_GRSTCTL_TXFFLSH | OTG_GRSTCTL_TXFNUM_ALL | OTG_GRSTCTL_RXFFLSH;` is not sufficient to flush the fifo buffers
- OTG_DIEPSIZ0_PKTCNT is misleading (renamed it to OTG_DIEPSIZ_PKTCNT_1 = 1 packet, like OTG_DIEPSIZ0_STUPCNT_1)
[...]

PR #672, opened 2016-07-10
STM32F2: make control usb stack more robust.

jhoenicke:
These patches aim to fix the problems of #668.  There are 3 parts:
1. If we detect there is a packet in the TxFIFO (TSIZ0.pktcnt is not zero) when a SETUP is received, then this packet is discarded by flushing the FIFO.  It doesn't hurt, since the packet was definitely for a different request that is no longer requested.  It is necessary since otherwise the control port doesn't answer the new setup packet.
2. We first check for DIEPINT before checking for RXFLVL.  This is necessary, because the usb_control statemachine requires that we first call usb_control_in on setup_read transactions before calling usb_control_out.
3. We keep better track of rxbcnt (I didn't have the problem but we should fix it nonetheless).  I also removed the FIXME delay loop as I think it is no longer necessary.

I didn't implement the idea of not answering a SETUP packet if there is another SETUP in the queue.  The idea of @zyp (to poll but not pop GRXSTSR) does not work.  There is still the SETUP_DONE packet in the RXFIFO.  Also, I don't want to change to much.  So the current implementation will still answer the wrong SETUP packet in the case where usbd_poll is called after a timeout.

### DIFF
--- a/lib/usb/usb_fx07_common.c
+++ b/lib/usb/usb_fx07_common.c
@@ -247,11 +247,6 @@ uint16_t stm32fx07_ep_read_packet(usbd_device *usbd_dev, uint8_t addr,
 		memcpy(buf32, &extra, i);
 	}
 
-	REBASE(OTG_DOEPTSIZ(addr)) = usbd_dev->doeptsiz[addr];
-	REBASE(OTG_DOEPCTL(addr)) |= OTG_DOEPCTL0_EPENA |
-	    (usbd_dev->force_nak[addr] ?
-	     OTG_DOEPCTL0_SNAK : OTG_DOEPCTL0_CNAK);
-
 	return len;
 }
 
@@ -315,12 +310,21 @@ void stm32fx07_poll(usbd_device *usbd_dev)
 		/* Receive FIFO non-empty. */
 		uint32_t rxstsp = REBASE(OTG_GRXSTSP);
 		uint32_t pktsts = rxstsp & OTG_GRXSTSP_PKTSTS_MASK;
+		uint8_t ep = rxstsp & OTG_GRXSTSP_EPNUM_MASK;
+		if (pktsts == OTG_GRXSTSP_PKTSTS_OUT_COMP
+			|| pktsts == OTG_GRXSTSP_PKTSTS_SETUP_COMP)  {
+			REBASE(OTG_DOEPTSIZ(ep)) = usbd_dev->doeptsiz[ep];
+			REBASE(OTG_DOEPCTL(ep)) |= OTG_DOEPCTL0_EPENA |
+				(usbd_dev->force_nak[ep] ?
+				 OTG_DOEPCTL0_SNAK : OTG_DOEPCTL0_CNAK);
+			return;
+		}
+
 		if ((pktsts != OTG_GRXSTSP_PKTSTS_OUT) &&
 		    (pktsts != OTG_GRXSTSP_PKTSTS_SETUP)) {
 			return;
 		}
 
-		uint8_t ep = rxstsp & OTG_GRXSTSP_EPNUM_MASK;
 		uint8_t type;
 		if (pktsts == OTG_GRXSTSP_PKTSTS_SETUP) {
 			type = USB_TRANSACTION_SETUP;
@@ -338,15 +342,6 @@ void stm32fx07_poll(usbd_device *usbd_dev)
 
 		/* Save packet size for stm32f107_ep_read_packet(). */
 		usbd_dev->rxbcnt = (rxstsp & OTG_GRXSTSP_BCNT_MASK) >> 4;
-
-		/*
-		 * FIXME: Why is a delay needed here?
-		 * This appears to fix a problem where the first 4 bytes
-		 * of the DATA OUT stage of a control transaction are lost.
-		 */
-		for (i = 0; i < 1000; i++) {
-			__asm__("nop");
-		}
 
 		if (usbd_dev->user_callback_ctr[ep][type]) {
 			usbd_dev->user_callback_ctr[ep][type] (usbd_dev, ep);
--- a/lib/usb/usb_fx07_common.c
+++ b/lib/usb/usb_fx07_common.c
@@ -262,6 +262,23 @@ void stm32fx07_poll(usbd_device *usbd_dev)
 		return;
 	}
 
+	/*
+	 * There is no global interrupt flag for transmit complete.
+	 * The XFRC bit must be checked in each OTG_DIEPINT(x).
+	 */
+	for (i = 0; i < 4; i++) { /* Iterate over endpoints. */
+		if (REBASE(OTG_DIEPINT(i)) & OTG_DIEPINTX_XFRC) {
+			/* Transfer complete. */
+			if (usbd_dev->user_callback_ctr[i]
+						       [USB_TRANSACTION_IN]) {
+				usbd_dev->user_callback_ctr[i]
+					[USB_TRANSACTION_IN](usbd_dev, i);
+			}
+
+			REBASE(OTG_DIEPINT(i)) = OTG_DIEPINTX_XFRC;
+		}
+	}
+
 	/* Note: RX and TX handled differently in this device. */
 	if (intsts & OTG_GINTSTS_RXFLVL) {
 		/* Receive FIFO non-empty. */
@@ -302,23 +319,6 @@ void stm32fx07_poll(usbd_device *usbd_dev)
 		}
 
 		usbd_dev->rxbcnt = 0;
-	}
-
-	/*
-	 * There is no global interrupt flag for transmit complete.
-	 * The XFRC bit must be checked in each OTG_DIEPINT(x).
-	 */
-	for (i = 0; i < 4; i++) { /* Iterate over endpoints. */
-		if (REBASE(OTG_DIEPINT(i)) & OTG_DIEPINTX_XFRC) {
-			/* Transfer complete. */
-			if (usbd_dev->user_callback_ctr[i]
-						       [USB_TRANSACTION_IN]) {
-				usbd_dev->user_callback_ctr[i]
-					[USB_TRANSACTION_IN](usbd_dev, i);
-			}
-
-			REBASE(OTG_DIEPINT(i)) = OTG_DIEPINTX_XFRC;
-		}
 	}
 
 	if (intsts & OTG_GINTSTS_USBSUSP) {
--- a/lib/usb/usb_fx07_common.c
+++ b/lib/usb/usb_fx07_common.c
@@ -248,6 +248,30 @@ uint16_t stm32fx07_ep_read_packet(usbd_device *usbd_dev, uint8_t addr,
 	return len;
 }
 
+static void stm32fx07_flush_txfifo(usbd_device *usbd_dev, int ep)
+{
+	uint32_t fifo;
+	/* set IN endpoint NAK */
+	REBASE(OTG_DIEPCTL(ep)) |= OTG_DIEPCTL0_SNAK;
+	/* wait for core to respond */
+	while (!(REBASE(OTG_DIEPINT(ep)) & OTG_DIEPINTX_INEPNE)) {
+		/* idle */
+	}
+	/* get fifo for this endpoint */
+	fifo = (REBASE(OTG_DIEPCTL(ep)) & OTG_DIEPCTL0_TXFNUM_MASK) >> 22;
+	/* wait for core to idle */
+	while (!(REBASE(OTG_GRSTCTL) & OTG_GRSTCTL_AHBIDL)) {
+		/* idle */
+	}
+	/* flush tx fifo */
+	REBASE(OTG_GRSTCTL) = (fifo << 6) | OTG_GRSTCTL_TXFFLSH;
+	/* reset packet counter */
+	REBASE(OTG_DIEPTSIZ(ep)) = 0;
+	while ((REBASE(OTG_GRSTCTL) & OTG_GRSTCTL_TXFFLSH)) {
+		/* idle */
+	}
+}
+
 void stm32fx07_poll(usbd_device *usbd_dev)
 {
 	/* Read interrupt status register. */
@@ -295,6 +319,14 @@ void stm32fx07_poll(usbd_device *usbd_dev)
 			type = USB_TRANSACTION_SETUP;
 		} else {
 			type = USB_TRANSACTION_OUT;
+		}
+
+		if (type == USB_TRANSACTION_SETUP
+			&& (REBASE(OTG_DIEPTSIZ(ep)) & OTG_DIEPSIZ0_PKTCNT)) {
+			/* SETUP received but there is still something stuck
+			 * in the transmit fifo.  Flush it.
+			 */
+			stm32fx07_flush_txfifo(usbd_dev, ep);
 		}
 
 		/* Save packet size for stm32f107_ep_read_packet(). */
--- a/lib/usb/usb_fx07_common.c
+++ b/lib/usb/usb_fx07_common.c
@@ -228,15 +228,22 @@ uint16_t stm32fx07_ep_read_packet(usbd_device *usbd_dev, uint8_t addr,
 	uint32_t extra;
 
 	len = MIN(len, usbd_dev->rxbcnt);
-	usbd_dev->rxbcnt -= len;
 
 	volatile uint32_t *fifo = REBASE_FIFO(addr);
 	for (i = len; i >= 4; i -= 4) {
 		*buf32++ = *fifo++;
+		usbd_dev->rxbcnt -= 4;
 	}
 
 	if (i) {
 		extra = *fifo++;
+		/* we read 4 bytes from the fifo, so update rxbcnt */
+		if (usbd_dev->rxbcnt < 4) {
+			/* Be careful not to underflow (rxbcnt is unsigned) */
+			usbd_dev->rxbcnt = 0;
+		} else {
+			usbd_dev->rxbcnt -= 4;
+		}
 		memcpy(buf32, &extra, i);
 	}
 
--- a/lib/usb/usb_fx07_common.c
+++ b/lib/usb/usb_fx07_common.c
@@ -31,7 +31,6 @@
  * according to the selected cores base address. */
 #define dev_base_address (usbd_dev->driver->base_address)
 #define REBASE(x)        MMIO32((x) + (dev_base_address))
-#define REBASE_FIFO(x)   (&MMIO32((dev_base_address) + (OTG_FIFO(x))))
 
 void stm32fx07_set_address(usbd_device *usbd_dev, uint8_t addr)
 {
@@ -210,11 +209,10 @@ uint16_t stm32fx07_ep_write_packet(usbd_device *usbd_dev, uint8_t addr,
 	REBASE(OTG_DIEPTSIZ(addr)) = OTG_DIEPSIZ0_PKTCNT | len;
 	REBASE(OTG_DIEPCTL(addr)) |= OTG_DIEPCTL0_EPENA |
 				     OTG_DIEPCTL0_CNAK;
-	volatile uint32_t *fifo = REBASE_FIFO(addr);
 
 	/* Copy buffer to endpoint FIFO, note - memcpy does not work */
 	for (i = len; i > 0; i -= 4) {
-		*fifo++ = *buf32++;
+		REBASE(OTG_FIFO(addr)) = *buf32++;
 	}
 
 	return len;
@@ -227,16 +225,19 @@ uint16_t stm32fx07_ep_read_packet(usbd_device *usbd_dev, uint8_t addr,
 	uint32_t *buf32 = buf;
 	uint32_t extra;
 
+	/* We do not need to know the endpoint address since there is only one
+	 * receive FIFO for all endpoints.
+	 */
+	(void) addr;
 	len = MIN(len, usbd_dev->rxbcnt);
 
-	volatile uint32_t *fifo = REBASE_FIFO(addr);
 	for (i = len; i >= 4; i -= 4) {
-		*buf32++ = *fifo++;
+		*buf32++ = REBASE(OTG_FIFO(0));
 		usbd_dev->rxbcnt -= 4;
 	}
 
 	if (i) {
-		extra = *fifo++;
+		extra = REBASE(OTG_FIFO(0));
 		/* we read 4 bytes from the fifo, so update rxbcnt */
 		if (usbd_dev->rxbcnt < 4) {
 			/* Be careful not to underflow (rxbcnt is unsigned) */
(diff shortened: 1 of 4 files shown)


jhoenicke:
BTW, I made some experiments and it never happened that the busy-wait loop in flush_txfifo were executed.  I'm not sure, if one needs to add some counters to these loops, to avoid freezing the device under strange circumstances.  `stm32f207_usbd_init` has similar unguarded while loops, though.

ChuckM:
Nice work. I've pulled this and run it on my usb code and it works well.
:+1:

H2OBrain:
Nice to see that someone made progress with USB :)

I have not really tested your code, but Issue #600 sadly still persists.
Here is the gadget0 test i hacked together, [fix1_testcase](/H2OBrain/libopencm3/tree/fix1_testcase).
It now always fails like before when ran this test without the busy-wait loop
```
- buf_out : b'4567890123456789012345678901234567890123456789012345678901234567'
- buf_in  : b'890123456789012345678901234567890123456789012345678901234567\x00\x00g\x00'
```
As you can see, the buf_in is 4 bytes behind buf_out. In my version this happens about 95% of the time.

In my own code I removed the busy wait loop because it makes USB unstable. On the PC side I have to send 4 preceding bytes in every control-out transfer so I rarely use any control-out transfers at all.

jhoenicke:
@H2OBrain I'll try if I can reproduce this on the STM32F2. 
But can you provide some logs, preferably also an USB dump (e.g. taken with wireshark)?
It looks like someone is stealing one word from the FIFO.  If I understand correctly this must be between the GRXSTSP access reading the OUT packet and the usbd_ep_read_packet call, and there is not much code in between.  What happens if you replace GRXSTSP by GRXSTSR  in `usb_fx07_common-debug.c` line 312 (this shouldn't work, but it would be interesting to know what happens)?

jhoenicke:
I got the test program running and could reproduce the problem.  No need to send me any dumps.

H2OBrain:
thanks, i'm a bit busy at the moment, sorry :)

zyp:
I made some comments about @H2OBrain's comment in IRC earlier today. I think I might be onto something so I'm attaching it here too for reference.
```
11:19:55 < zyp> h2obrain, hmm, that's a weird bug
11:21:04 < zyp> I haven't seen that in my stack
11:28:34 < zyp> oh, I remember I was having some problems, and I can see I am doing something different from libopencm3
11:37:04 < zyp> hmm, changing libopencm3 looks like it'll require some work, I should replicate the test case with my stack and try reproduce the problem
12:11:40 < h2obrain> zyp, it's still possible that my control-transfer is completely wrong..
12:12:15 < zyp> I think this is possibly related to some stuff I saw when I was writing my own stack
12:12:35 < zyp> at the time I didn't know much about usb, so I just tried stuff until I found something that works
12:12:56 < h2obrain> hehe, i am still in that stage :)
12:15:07 < zyp> anyway, the difference is which events from GRXSTSP triggers what
12:15:57 < zyp> libopencm3 ignores all other events than OUT packet received and SETUP packet received
12:17:19 < zyp> in my stack, when I get SETUP packet received, I only buffer the packet, but don't process it (and then CNAK) until I get SETUP transaction completed
12:17:40 < zyp> ref. http://cgit.jvnv.net/laks/tree/usb/dwc_otg.h#n20
12:17:56 < zyp> 0x6 is setup packet received, 0x4 is setup transaction complete
12:18:35 < zyp> I remember changing that around due to some weird behavior until I landed on something that worked
12:19:13 < zyp> so I suspect that there's a race condition between setting CNAK (and consequently receiving a new in packet) and reading out the setup transaction done token
12:20:19 < zyp> in libopencm3 there's more layers between the rxfifo handler and the code setting CNAK, so it's going to be a bit more work to split apart
12:22:59 < zyp> receiving a new OUT packet*
13:13:59 < h2obrain> thanks for the hint, i will look into that  the next time i work on the usb stack
```
It'll be a bit of a hassle to rewrite the libopencm3 stack to work this way, but not processing the control request until «SETUP transaction completed» is popped will also allow you to peek at GRXSTSR and drop the request if there's another in the FIFO.

I'd try to reproduce this issue with my stack to check that the event timing really makes a difference, but unfortunately I don't have access to any hardware to test on this week.

jhoenicke:
It's definitely setting CNAK on the DOEPCTL that caused this.  Putting the delay loop before this statement also works, putting it after it breaks.  

I now tried setting the CNAK only when OUT_COMP/SETUP_COMP is received (see above).  I'm currently not sure if there is an OUT_COMP after every OUT packet or only after the final one.

zyp:
When DOEPTSIZx.PKTCNT is always set to 1, every packet is always the last one, so that shouldn't really matter.

I also believe that when it's configured to expect multiple packets, hardware doesn't assert NAK until the transfer is complete, so only setting CNAK after OUT_COMP should be fine in either case.

H2OBrain:
in your latest patch (c692e86) in stm32fx07_ep_read_packet() you have to replace REBASE_FIFO(0) with REBASE_FIFO(addr) twice (usb_fx07_common.c, lines 233/238)

jhoenicke:
The REBASE_FIFO(0) was intentional, as there is only one receive fifo shared between all endpoints.  But I can change it back to REBASE_FIFO(ep/addr) if that is more readable.  Changing the last `*fifo++` looks probably better.

I'm not sure why the chip designers made the fifo a region; maybe to support memcpy but that doesn't work reliably because byte-wise access doesn't work.

H2OBrain:
I didn't notice that. Then I would leave it REBASE_FIFO(0).
I managed to merge your fixes in my own branch and it seems to work. Now I can test it.

jhoenicke:
From my side the patch set is ready to be merged.  I just rebased it to the current master.  The problem with the lost 4 bytes is fixed in https://github.com/libopencm3/libopencm3/pull/672/commits/faf1709d0cf63215ac74b7829cb449c03be01186

Do you want me to change something else in these patches?

ChuckM:
:+1:

karlp:
just fyi, I'm still on holidays for another week or two, and not near my desk.  Did someone setup a test case that fails until this code is merged? or does it still require manual testing with some blocking?

H2OBrain:
my testcase above covers only the word-loss part..
[fix1_testcase](/H2OBrain/libopencm3/tree/fix1_testcase)

PS: as far as i can tell, this also fixes the fifo-flush problems i had in my multi-packet proposal (issue #549). I undid the changes to stm32fx07_endpoints_reset in commit 'made maxpacketsize and usb-driver detection variable' f3bf69d in my libopencm3-fork and usb still works

jhoenicke:
Test cases added (finally).  The first test  is from @H2OBrain (I had to renumber the commands since they conflicted with align).  The second patch just names magic numbers in the python test script.  The third patch is the one that tests my  issue: interrupt transfers that take a while to process.

I rebased everything on current master.  Tested with the stm32f4disco configuration (I now have some hardware that is compatible to that).

tests/gadget-zero/usb-gadget0.c
```diff
@@ -297,6 +300,37 @@ static int gadget0_control_request(usbd_device *usbd_dev,
...
 		}
 		return USBD_REQ_HANDLED;
+	case GZ_REQ_WRITE_LOOPBACK_BUFFER:
+		if (req->wValue > sizeof(usbd_control_buffer)) {
+			ER_DPRINTF("Can't store more than out control buffer! %d > %d\n",
+				req->wValue, sizeof(usbd_control_buffer));
+			return USBD_REQ_NOTSUPP;
+		}
+		/* Don't store more than asked for! */
+		if (req->wValue > req->wLength) {
```
karlp: I wouldn't have thought this was a necessary check, as long as you didn't overflow, but I guess it doesn't hurt.

tests/gadget-zero/test_gadget0.py
```diff
@@ -43,7 +46,7 @@ def tearDown(self):
         uu.dispose_resources(self.dev)
 
     def test_sanity(self):
-        self.assertEqual(2, self.dev.bNumConfigurations, "Should have 2 configs")
+        self.assertEqual(3, self.dev.bNumConfigurations, "Should have 2 configs")
```
karlp: Should have 3 configs?

tests/gadget-zero/test_gadget0.py
```diff
@@ -3,8 +3,11 @@
 import usb.core
 import usb.util as uu
 import logging
+import struct
+import time
 
 import unittest
+import pytest
```
karlp: What are trying here? mixing different styles of tests in the same file is pretty confusing to say the least.

tests/gadget-zero/test_gadget0.py
```diff
@@ -191,7 +194,7 @@ def setUp(self):
 
         self.cfg = uu.find_descriptor(self.dev, bConfigurationValue=2)
         self.assertIsNotNone(self.cfg, "Config 2 should exist")
-        self.dev.set_configuration(self.cfg);
+        self.dev.set_configuration(self.cfg)
```
karlp: valid, but noise for this commit

tests/gadget-zero/test_gadget0.py
```diff
@@ -242,7 +245,7 @@ def setUp(self):
...
```
karlp: valid, but noise for this commit

tests/gadget-zero/test_gadget0.py
```diff
@@ -294,12 +297,7 @@ def inner(x):
                              buf_out.tostring(), buf_in.tostring()))
         
         ep0_size = self.dev.bMaxPacketSize0
-        #for i in range(4, ep0_size) : inner(i)
-        # some random sizes to write/readback..
-        inner(ep0_size)
-        inner(ep0_size+1)
-        inner(4)
-        inner(57)
+        for i in range(4, ep0_size) : inner(i)
```
karlp: back to theoriginal h2obrain code :) seems out of scope for this commit.

tests/gadget-zero/test_gadget0.py
```diff
@@ -406,5 +404,74 @@ def test_unaligned(self):
...
+        Test if usb stack recovers correctly after a timeout.
+        The long delay will cause a timeout on the next ctrl_transfer
+        and we check that this gives an exception.  The next ctrl_transfer
+        should go through without problems.
+        """
+        x = 16
+        buf_out = array.array('b')
+        for i in range(0,x) : buf_out.append(48 + (x+i) % 10)
+        self.dev.ctrl_transfer(self.req_out, 3, x, 0, buf_out)
+        delaycmd = struct.pack('<BL', 0, 50000000)
```
karlp: given you pulled up constants earlier, I'd like to see these at least broken into 1000s.  is this 500 ms or what?

tests/gadget-zero/usb-gadget0.c
```diff
@@ -254,6 +307,28 @@ static void gadget0_tx_cb_loopback(usbd_device *usbd_dev, uint8_t ep)
...
 }
 
+static void gadget0_rx_cb_interrupt(usbd_device *usbd_dev, uint8_t ep)
+{
+	static uint8_t buf[64] __attribute__ ((aligned(4)));
+	int len = usbd_ep_read_packet(usbd_dev, ep, buf, 64);
+
+	ER_DPRINTF("interrupt rx %x %d\n", ep, len);
+
+	if (len == 0)
```
karlp: always use {}

tests/gadget-zero/usb-gadget0.c
```diff
@@ -254,6 +307,28 @@ static void gadget0_tx_cb_loopback(usbd_device *usbd_dev, uint8_t ep)
...
+		return;
+	switch (buf[0]) {
+	case 0: /* delay */ {
```
karlp: your love of constants didn't shine through here :)

tests/gadget-zero/usb-gadget0.c
```diff
@@ -254,6 +307,28 @@ static void gadget0_tx_cb_loopback(usbd_device *usbd_dev, uint8_t ep)
...
+		if (len < 5)
+			return;
+		uint32_t delay = buf[1] | ((uint32_t) buf[2] << 8) | ((uint32_t) buf[3] << 16) | ((uint32_t) buf[4] << 24);
```
karlp: given you have control over the byte order when you made the packet, why not make it so you can just cast this here?

tests/gadget-zero/usb-gadget0.c
```diff
@@ -254,6 +307,28 @@ static void gadget0_tx_cb_loopback(usbd_device *usbd_dev, uint8_t ep)
...
+		while (delay--) {
+			__asm__ ("nop");
+		}
+		ER_DPRINTF("delayed\n");
+	}
+	}
```
karlp: something looks wrong style wise here.

lib/usb/usb_fx07_common.c
```diff
@@ -262,6 +262,23 @@ void stm32fx07_poll(usbd_device *usbd_dev)
 		return;
 	}
 
+	/*
+	 * There is no global interrupt flag for transmit complete.
+	 * The XFRC bit must be checked in each OTG_DIEPINT(x).
+	 */
+	for (i = 0; i < 4; i++) { /* Iterate over endpoints. */
```
karlp: This number is too small for the OTG_HS device, which also uses this code isn't it?

lib/usb/usb_fx07_common.c
```diff
@@ -304,23 +321,6 @@ void stm32fx07_poll(usbd_device *usbd_dev)
 		usbd_dev->rxbcnt = 0;
 	}
 
-	/*
-	 * There is no global interrupt flag for transmit complete.
-	 * The XFRC bit must be checked in each OTG_DIEPINT(x).
-	 */
-	for (i = 0; i < 4; i++) { /* Iterate over endpoints. */
```
karlp: ok, existing bug then :)  think it might have already been filed elsewhere, but fine for now then.

lib/usb/usb_fx07_common.c
```diff
@@ -228,15 +228,22 @@ uint16_t stm32fx07_ep_read_packet(usbd_device *usbd_dev, uint8_t addr,
 	uint32_t extra;
```
karlp: So, is comment about reading seveeral small chunks, what's the impact of that? is that a change from existing?

lib/usb/usb_fx07_common.c
```diff
@@ -262,6 +262,23 @@ void stm32fx07_poll(usbd_device *usbd_dev)
...
 
+	/*
+	 * There is no global interrupt flag for transmit complete.
+	 * The XFRC bit must be checked in each OTG_DIEPINT(x).
+	 */
+	for (i = 0; i < 4; i++) { /* Iterate over endpoints. */
+		if (REBASE(OTG_DIEPINT(i)) & OTG_DIEPINTX_XFRC) {
+			/* Transfer complete. */
+			if (usbd_dev->user_callback_ctr[i]
+						       [USB_TRANSACTION_IN]) {
```
jhoenicke: I'd rather not change this, since it is only moving the code from below without any changes.

tests/gadget-zero/test_gadget0.py
jhoenicke: I moved this to a separate commit

tests/gadget-zero/test_gadget0.py
jhoenicke: fixed.

tests/gadget-zero/usb-gadget0.c
jhoenicke: I rewrote h2obrain's loopback code to not use wValue at all.

tests/gadget-zero/test_gadget0.py
jhoenicke: done.

lib/usb/usb_fx07_common.c
jhoenicke: Previous: if one reads a unaligned small chunk, fifo may get out of sync and some data bytes omitted
Now: fifo stays in sync, but data bytes still omitted for second read.

If fifo gets out of sync, then even the basic functionality is broken.

davids5:
@jhoenicke 

RE:The missing word in the FIFO - I would love if we could collaborate and share our findings.
I have not found a good reference for the proper interaction with the STM32[F|L][4|7] USB core. That describes the states and requirements to service them. Do you (or anyone) have one? I feel as if I am pushing a race condition around.

We first noticed issues on the F4 in PX4 on NuttX where the value in the EPNUM had bad values read from the OTGFS_GRXSTSP register. We were seeing 0xb4e48168 or 0xa80c9367 or 267E781c. All of which provide out of range indexes for  EPNUM.

It was a random occurrence and I failed to find the root cause, at the time. We added a workaround for it. Here is an eaxample of the same code on the L4 https://bitbucket.org/nuttx/nuttx/src/d4a8585d6f2819f4c4b2c92262542219801ee652/arch/arm/src/stm32l4/stm32l4_otgfsdev.c?fileviewer=file-view-default#stm32l4_otgfsdev.c-3252 

Below are some links to the discussion and fixes for the missing word in the FIFO. The F7 link has more debug detail. See "This is what it look like when it is bad" on that page. and "I finally got the STM32L476G-DISCO" and "For the L4 i am stumped" for what I was seeing on the L4

https://github.com/libopencm3/libopencm3/pull/681#issuecomment-251841801 herein discussing the FIFO and USB core issues.

F4: https://bitbucket.org/nuttx/nuttx/pull-requests/125/bugfix-lost-first-word-from-fifo/diff
F7:https://bitbucket.org/nuttx/nuttx/pull-requests/129/upstream_to_greg/diff
L4:https://bitbucket.org/nuttx/nuttx/pull-requests/140/master-l4-usbfix/diff

jhoenicke:
@davids5 The FIFO fix is https://github.com/libopencm3/libopencm3/pull/672/commits/fd2cd09b1396dc11c22bd3181a2e249be0901c7c
The documentation is not very clear here.  But with this patch the application program sequence to read OUT packets is only restarted (i.e. DOEPTSIZ and DOEPCTL are written) when the "OUT data transfer completed" message is popped from the FIFO.  Previously it did it already on the "OUT data" message. This probably lead to race conditions, but it is hard to say without knowing what the hardware internally does.

@karlp: Since your reviewed the previous patch, I also have at https://github.com/jhoenicke/libopencm3/commits/fix2 the commits in the original order followed by the commits that fix the issues you found.  fix2 and fix1 should lead to an identical image.  fix1 has the commits in a clean order.  I'm not sure why github shows the commits in a random order (it probably sorts it by first commit date).

prusnak:
Any update on this?

prusnak:
@BuFran @esden @gsmcmullin What do you think? Or who can decide whether this should go in, i.e. who is the maintainer of USB stack?

karlp:
It's probably all good, now that earlier comments have been addressed, but I've not looked at it since, and want to explore how the test cases work.  I want to make sure that I'm seeing the failure before, and not after.  (I've not been seeing the failure before at this stage)  However, I'm currently dedicating virtually all of my spare development time to a personal christmas project, and simple haven't had time to look at anything meaningful in locm3 for the last month or two, and don't expect to be able to again until at least after christmas.

prusnak:
Hey @karlp, how did your Christmas project go? :)

karlp:
went ok ish I guess.  Not nearly as complete as i'd hoped, but it got the bare minimum in the end, and it can be field upgradeable if I ever get around to it.  Needs a good blog post to wrap it up.  Basic video preview at https://www.youtube.com/watch?v=3-j7mIYNssI

Back looking at basic stuff here again, will try and chew down the PRs a bit.  USB's turn has come again.

martin-lizner:
Cant wait for this PR to be merged already. Thanks.

ChuckM:
bump 👍

tests/gadget-zero/test_gadget0.py
```diff
@@ -369,3 +408,75 @@ def test_aligned(self):
...
+
+        self.cfg = uu.find_descriptor(self.dev, bConfigurationValue=4)
+        self.assertIsNotNone(self.cfg, "Config 4 should exist")
+        self.dev.set_configuration(self.cfg)
+        self.req     = uu.CTRL_IN  | uu.CTRL_TYPE_VENDOR | uu.CTRL_RECIPIENT_INTERFACE
+        self.req_out = uu.CTRL_OUT | uu.CTRL_TYPE_VENDOR | uu.CTRL_RECIPIENT_INTERFACE
+        self.intr_out = 1
+        # send dummy packages on interrupt port to get in sync
+        # (alternating bit protocol)
+        # FIXME: Why is this necessary?
```
karlp: Any theories?

karlp: suggested that this is a work around for another existing data toggle bug in the stack.  sounds good to me.

tests/gadget-zero/test_gadget0.py
```diff
@@ -401,5 +410,73 @@ def test_unaligned(self):
...
+        """
+        x = 16
+        buf_out = array.array('b')
+        for i in range(0,x) : buf_out.append(48 + (x+i) % 10)
+        self.dev.ctrl_transfer(self.req_out, GZ_REQ_WRITE_LOOPBACK_BUFFER, 0, 0, buf_out)
+        delaycmd = struct.pack('<BL', GZ_INTR_CMD_DELAY, 1500 * DELAY_MSEC)
+        self.dev.write(self.intr_out, delaycmd)
+        # this is expected to timeout
+        self.assertRaises(usb.USBError, self.dev.ctrl_transfer, self.req, GZ_REQ_READ_LOOPBACK_BUFFER, 0, 0, x)
+        time.sleep(2)
```
karlp: sleep 2 is the 1500msec from the command, plus what?  What timeout will libusb use?  these seems far longer than it should need to be, if it's needed at all?

tests/gadget-zero/test_gadget0.py
jhoenicke: Something on the hardware level.  The alternating bit protocol is not visible by the software but implemented in the STMF2 in hardware.  Due to the timeout one of the parties doesn't toggle the bit (I don't remember which), so the next message is just ignored.

tests/gadget-zero/test_gadget0.py
jhoenicke: Yeah, I wanted to be safe.  Something like 500 ms may be enough here (1500 ms delay minus 1 second waiting for timeout).

prusnak:
I've seen that some of the commits appeared in the master. Is this pull request fully merged?

prusnak:
I rebased both `fix1`/`fix2` branches from jhoenicke's repo on top of current master to see that they indeed produce the same source tree. Also I confirm that all fixes made in files `lib/usb/usb_control.c and `lib/usb/usb_fx07_common.c` are completely merged. What has not been fully merged are changes in `tests/gadget-zero/usb-gadget0.c` and `tests/gadget-zero/test_gadget0.py` as there are still some pending changes in `fix1`/`fix2` branches.

karlp:
I believe this is complete/superseded by recent code merges, and covered by the tests added.  If anyone is still having issues, or believes there is extra code that should still be added, please re-open or comment further.  (The interrupt endpoint code is completely out of scope of this, that should be filed separately)

PR #673, opened 2016-07-14
STM32F1: RTC: rtc_awake_from_off() clears CR register incorrectly.

ksarkies:
It sets bit 5 (manual says "read only") and clears bit 4 which takes the RTC
out of config mode. The RTC registers are not cleared as a result.
Probably not necessary to clear any registers, but this at least will do it.


karlp:
this has now landed in 808cd44 thanks!

PR #676, opened 2016-07-26
Added Support for stm32F446 stm32f469 usb

davids5:
@ChuckM, @nschodet-sevenhugs

This has been tested on STM32F469 and STM32F427


lib/usb/usb_f107.c
```diff
@@ -78,11 +88,15 @@ static usbd_device *stm32f107_usbd_init(void)
...
 	OTG_FS_GAHBCFG |= OTG_GAHBCFG_GINT;
-	OTG_FS_GINTMSK = OTG_GINTMSK_ENUMDNEM |
+	uint32_t reg32 = OTG_GINTMSK_ENUMDNEM |
 			 OTG_GINTMSK_RXFLVLM |
 			 OTG_GINTMSK_IEPINT |
 			 OTG_GINTMSK_USBSUSPM |
 			 OTG_GINTMSK_WUIM;
+
+	if (id == OTG_CID_HAS_VBDEN) {
+	}
```
karlp: this looks kinda irrelevant?

lib/usb/usb_fx07_common.c
```diff
@@ -336,11 +336,11 @@ void stm32fx07_poll(usbd_device *usbd_dev)
...
 	if (intsts & OTG_GINTSTS_SOF) {
-		if (usbd_dev->user_callback_sof) {
-			usbd_dev->user_callback_sof();
+			if (usbd_dev->user_callback_sof) {
+				usbd_dev->user_callback_sof();
+			}
+			REBASE(OTG_GINTSTS) = OTG_GINTSTS_SOF;
 		}
-		REBASE(OTG_GINTSTS) = OTG_GINTSTS_SOF;
-	}
```
karlp: this looks like unrelated whitespace changes?

karlp:
while this works, I kinda think this would still be best addressed by having driver private data, and having them as two different drivers.  Even the existing hard turn on of vbus detection is dubious, it should be an optional api call, and splitting the drivers up allows this and more.

davids5:
@karlp 

Thanks for the review. Sorry about the ws change and empty if - 

I am not really familiar with the library architecture or naming: If you can refer me to doc that explains the driver usage and layering then I can munge this into a new driver.

ChuckM:
Some additional comments, I would add the `RCC_PLLSAICFGR_PLLSAIP_SHIFT` and `RCC_PLLSAICFGR_PLLSAIP_MASK `defines to the file. Also update the pllsai helper macro to take that p value as a parameter. Also note that the define for the 'n' parameter is wrong the shift should be '6' not '14' (it ends at bit 14, the shift should relocate bits 6 - 14 down to 0 or a constant up to bit position 6.

davids5:
@ChuckM - LOL the inline had magic numbers not defines, that were correct. - it is fixed and the inline has been expanded to support 'p' and preserve the reserved bits. P of 0 is the reset state.

@karlp - the unneeded changes were backed out. As far as splitting this off to a new "driver". 
The more I think about it from a runtime, code size, and maintenance perspective:
 The code is called once - so the check for core 0x2000 is not a runtime issue
 From a maintenance standpoint why dup the code - as it is I fixed the init order to match the guidelines lines. - Why do that in 2 places?
Code sizewise - it is a nit - all the rcc tables waste far more space.

For these reasons I would leave it as is. But it is your project so it is up to you,

ChuckM:
Interesting I also need to add two clocks to the rcc_osc enum `RCC_PLLSAI` and `RCC_PLLI2S`. With commensurate fix ups in `rcc_osc_on`, `rcc_osc_off` and `rcc_wait_for_osc_ready`

ChuckM:
@davids5 I'd like to co-ordinate on this see #677 which is just the rcc changes (although i think I missed the code change on clock setting. I'll fix that.

davids5:
@ChuckM  - sure! - I would say to use your cleaned up file  and my inline the ensures the reserved bits are respected. Please merge how you see fit.

As far are the usb changes that can be cherry picked.

@karlp - will you be able to get the RCC and USB changes merged to master?

karlp:
I expect, when I've got my own computer back, that any basic rcc and bit definitions changes I should be able to merge easily, if the commits are clear, contained, and well labelled.  I don't expect to get my computer back for any meaningful work for another week or so, I have too many house guests this summer :(

ChuckM:
@karlp The basic rcc and bit definition changes are in #677 and they supercede this PR (for RCC changes)

davids5:
@karlp - Any chance of getting this in this week? The reason I ask is [here] (https://github.com/libopencm3/libopencm3/pull/677#issuecomment-236681349)

lib/usb/usb_f107.c
```diff
@@ -52,21 +52,31 @@ const struct _usbd_driver stm32f107_usb_driver = {
...
+	} else {
+		/* Enable VBUS sensing in device mode and power up the PHY. */
+		OTG_FS_GCCFG |= OTG_GCCFG_VBUSBSEN | OTG_GCCFG_PWRDWN;
+	}
+
+
 	/* Force peripheral only mode. */
 	OTG_FS_GUSBCFG |= OTG_GUSBCFG_FDMOD | OTG_GUSBCFG_TRDT_MASK;
 
+	OTG_FS_GINTSTS = OTG_GINTSTS_MMIS;
```
kuldeepdhaka: any specific reason to move this here?

lib/usb/usb_f107.c
```diff
@@ -52,21 +52,31 @@ const struct _usbd_driver stm32f107_usb_driver = {
...
-	OTG_FS_GCCFG |= OTG_GCCFG_VBUSBSEN | OTG_GCCFG_PWRDWN;
 
 	/* Wait for AHB idle. */
 	while (!(OTG_FS_GRSTCTL & OTG_GRSTCTL_AHBIDL));
 	/* Do core soft reset. */
 	OTG_FS_GRSTCTL |= OTG_GRSTCTL_CSRST;
 	while (OTG_FS_GRSTCTL & OTG_GRSTCTL_CSRST);
 
+	uint32_t id = OTG_FS_CID;
+	if (id == OTG_CID_HAS_VBDEN) {
```
kuldeepdhaka: I do not think relying on CID register is a good idea.
afaiu, It is just a value that can change between part with same capabilities.

lib/usb/usb_f107.c
```diff
@@ -52,21 +52,31 @@ const struct _usbd_driver stm32f107_usb_driver = {
...
+		/* Enable VBUS detection in device mode and power up the PHY. */
+		OTG_FS_GCCFG |= OTG_GCCFG_VBDEN | OTG_GCCFG_PWRDWN;
+		/* Set the Soft Connect (STMF32446, STMF32469 comes up disconnected) */
+		OTG_FS_DCTL &= ~OTG_DCTL_SDIS;
```
kuldeepdhaka: on other target, this value is low.
on the target you are having - it has this bit high.
so i think, this line can be used for both.

lib/usb/usb_f107.c
davids5: What do you understand the usage of CID? - I read it as the version of microcode running on the USB IP core. Therefore it it is what should be queried to determine the "Differences" Chip to Chip.

lib/usb/usb_f107.c
davids5: The this line of code makes up for the microcode change to the Poweron state of this bit. With this change the F469 and F446 have the Soft Disconnect state the same as the F427 at power up. Which is connected. It would 
NOT be appropriate for the F427

lib/usb/usb_f107.c
davids5: Yes.  Once the mode is selected it is appropriate to clear any pending error.

karlp:
commit headlines for all of these should be something like "stm32f4: xxxx" (see the history for reference)

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -103,6 +103,8 @@
 
 /* --- RCC_PLLCFGR values -------------------------------------------------- */
 
+/* PLLQ: [30:28] */
```
karlp: should be pllr right?

lib/stm32/f4/rcc.c
```diff
@@ -622,7 +627,7 @@ void rcc_clock_setup_hse_3v3(const struct rcc_clock_scale *clock)
 	rcc_set_ppre2(clock->ppre2);
 
 	rcc_set_main_pll_hse(clock->pllm, clock->plln,
-			clock->pllp, clock->pllq);
+			clock->pllp, clock->pllq,clock->pllr);
```
karlp: add a space after the comma please

lib/stm32/f4/rcc.c
```diff
@@ -578,13 +578,18 @@ void rcc_set_main_pll_hsi(uint32_t pllm, uint32_t plln, uint32_t pllp,
...
 
 void rcc_set_main_pll_hse(uint32_t pllm, uint32_t plln, uint32_t pllp,
-			  uint32_t pllq)
+			  uint32_t pllq, uint32_t pllr)
 {
-	RCC_PLLCFGR = (pllm << RCC_PLLCFGR_PLLM_SHIFT) |
+	uint32_t reg32 = RCC_PLLCFGR & RCC_PLLCFGR_NO_R_RESERVED;
+	if (pllr != 0) {
+		reg32 &= RCC_PLLCFGR_R_RESERVED;
+	}
```
karlp: given that you mask it with both anyway, it seems unecessary to have the full masks.  Could you add documentation to this function to explain the special use of the pllr param too please.  Also, is  there any particular reason you didn't add it to the hsi function?

lib/usb/usb_f107.c
karlp: My understanding is the same as kuldeeps, it's probably _functional_ and it's unlikely to be a problem, but it's the identifier of the core IP version, not what vendor customizations have been applied.  Regardless, why do you need the pointless variable?

lib/usb/usb_f107.c
```diff
@@ -83,6 +93,7 @@ static usbd_device *stm32f107_usbd_init(void)
 			 OTG_GINTMSK_IEPINT |
 			 OTG_GINTMSK_USBSUSPM |
 			 OTG_GINTMSK_WUIM;
+
```
karlp: please don't make unrelated white space changes.

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -115,6 +115,10 @@
 /* PLLM: [5:0] */
 #define RCC_PLLCFGR_PLLM_SHIFT			0
 
+#define RCC_PLLCFGR_NO_R_RESERVED 0xf0bc8000
+#define RCC_PLLCFGR_R_RESERVED    0x80bc8000
```
karlp: even without my earlier comments on these, (and the same applies to the new ones you're adding here) moving these defines is out of scope of this commit.

davids5: @ChuckM code will do that.

davids5:
@karlp - Let me know if you need anything else.

lib/usb/usb_f107.c
davids5: @karlp  it had been removed.

karlp:
Yes, we still only want single issue comits, not commits with review fixes in them please.

lib/usb/usb_f107.c
```diff
@@ -60,8 +60,7 @@ static usbd_device *stm32f107_usbd_init(void)
 	OTG_FS_GRSTCTL |= OTG_GRSTCTL_CSRST;
 	while (OTG_FS_GRSTCTL & OTG_GRSTCTL_CSRST);
 
-	uint32_t id = OTG_FS_CID;
-	if (id == OTG_CID_HAS_VBDEN) {
+	if (OTG_FS_CID == OTG_CID_HAS_VBDEN) {
```
kuldeepdhaka: the body of if statement can do the purpose, and the whole check `CID == HAS_VBDEN` can be removed.
and,
VBDEN and VBUSSEN maps to same bit.
After all, if the SDIS bit is already in clear state (on older uC), cleared on older uC would result in no change.

davids5: Perhaps you are looking at a different set of docs?  My understanding is that Bit 21 polarity has changed.

## STM32F427
DocID018909 Rev 8 1273/1718
RM0090 USB on-the-go full-speed (OTG_FS)
1362
image

## STM32F469

DocID026744 Rev 2 1311/1599 
image

Good point about SDIS - but I prefer the deliberate setting with the comment because it highlights the difference

davids5:
@karlp you have it here https://github.com/libopencm3/libopencm3/pull/681

PR #677, opened 2016-07-29
469 rcc update

ChuckM:
This is a combined PR with both the QUADSPI.h file and changes I've had to make to the rcc/rcc.h files to work with the DSI component. I went through and audited the rcc.h file and found a number of discrepancies (you can see in the diff) I've built the examples repo with this version but haven't run the examples to test them. Nothing major changed.

### DIFF
--- a/include/libopencm3/stm32/f4/rcc.h
+++ b/include/libopencm3/stm32/f4/rcc.h
@@ -98,25 +98,35 @@
 #define RCC_CR_HSEON				(1 << 16)
 /* HSICAL: [15:8] */
 /* HSITRIM: [7:3] */
+#define RCC_CR_HSITRIM_SHIFT			3
+#define RCC_CR_HSITRIM_MASK			0x1f
 #define RCC_CR_HSIRDY				(1 << 1)
 #define RCC_CR_HSION				(1 << 0)
 
 /* --- RCC_PLLCFGR values -------------------------------------------------- */
 
+/* PLLR: [30:28] */
+#define RCC_PLLCFGR_PLLR_SHIFT			28
+#define RCC_PLLCFGR_PLLR_MASK			0x7
 /* PLLQ: [27:24] */
 #define RCC_PLLCFGR_PLLQ_SHIFT			24
+#define RCC_PLLCFGR_PLLQ_MASK			0xf
 #define RCC_PLLCFGR_PLLSRC			(1 << 22)
 /* PLLP: [17:16] */
 #define RCC_PLLCFGR_PLLP_SHIFT			16
+#define RCC_PLLCFGR_PLLP_MASK			0x3
 /* PLLN: [14:6] */
 #define RCC_PLLCFGR_PLLN_SHIFT			6
+#define RCC_PLLCFGR_PLLN_MASK			0x1ff
 /* PLLM: [5:0] */
 #define RCC_PLLCFGR_PLLM_SHIFT			0
+#define RCC_PLLCFGR_PLLM_MASK			0x3f
 
 /* --- RCC_CFGR values ----------------------------------------------------- */
 
 /* MCO2: Microcontroller clock output 2 */
 #define RCC_CFGR_MCO2_SHIFT			30
+#define RCC_CFGR_MC02_MASK			0x3
 #define RCC_CFGR_MCO2_SYSCLK			0x0
 #define RCC_CFGR_MCO2_PLLI2S			0x1
 #define RCC_CFGR_MCO2_HSE			0x2
@@ -124,7 +134,9 @@
 
 /* MCO1/2PRE: MCO Prescalers */
 #define RCC_CFGR_MCO2PRE_SHIFT			27
+#define RCC_CFGR_MCO2PRE_MASK			0x7
 #define RCC_CFGR_MCO1PRE_SHIFT			24
+#define RCC_CFGR_MCO1PRE_MASK			0x7
 #define RCC_CFGR_MCOPRE_DIV_NONE		0x0
 #define RCC_CFGR_MCOPRE_DIV_2			0x4
 #define RCC_CFGR_MCOPRE_DIV_3			0x5
@@ -190,6 +202,7 @@
 #define RCC_CIR_CSSC				(1 << 23)
 
 /* OSC ready interrupt clear bits */
+#define RCC_CIR_PLLSAIRDYC			(1 << 22)
 #define RCC_CIR_PLLI2SRDYC			(1 << 21)
 #define RCC_CIR_PLLRDYC				(1 << 20)
 #define RCC_CIR_HSERDYC				(1 << 19)
@@ -198,6 +211,7 @@
 #define RCC_CIR_LSIRDYC				(1 << 16)
 
 /* OSC ready interrupt enable bits */
+#define RCC_CIR_PLLSAIRDYIE			(1 << 14)
 #define RCC_CIR_PLLI2SRDYIE			(1 << 13)
 #define RCC_CIR_PLLRDYIE			(1 << 12)
 #define RCC_CIR_HSERDYIE			(1 << 11)
@@ -209,6 +223,7 @@
 #define RCC_CIR_CSSF				(1 << 7)
 
 /* OSC ready interrupt flag bits */
+#define RCC_CIR_PLLSAIRDYF			(1 << 6)
 #define RCC_CIR_PLLI2SRDYF			(1 << 5)
 #define RCC_CIR_PLLRDYF				(1 << 4)
 #define RCC_CIR_HSERDYF				(1 << 3)
@@ -220,9 +235,12 @@
 
 #define RCC_AHB1RSTR_OTGHSRST			(1 << 29)
 #define RCC_AHB1RSTR_ETHMACRST			(1 << 25)
+#define RCC_AHB1RSTR_DMA2DRST			(1 << 23)
 #define RCC_AHB1RSTR_DMA2RST			(1 << 22)
 #define RCC_AHB1RSTR_DMA1RST			(1 << 21)
 #define RCC_AHB1RSTR_CRCRST			(1 << 12)
+#define RCC_AHB1RSTR_IOPKRST			(1 << 10)
+#define RCC_AHB1RSTR_IOPJRST			(1 << 9)
 #define RCC_AHB1RSTR_IOPIRST			(1 << 8)
 #define RCC_AHB1RSTR_IOPHRST			(1 << 7)
 #define RCC_AHB1RSTR_IOPGRST			(1 << 6)
@@ -243,10 +261,13 @@
 
 /* --- RCC_AHB3RSTR values ------------------------------------------------- */
 
+#define RCC_AHB3RSTR_QSPIRST			(1 << 1)
 #define RCC_AHB3RSTR_FSMCRST			(1 << 0)
 
 /* --- RCC_APB1RSTR values ------------------------------------------------- */
 
+#define RCC_APB1RSTR_UART8RST			(1 << 31)
+#define RCC_APB1RSTR_UART7RST			(1 << 30)
 #define RCC_APB1RSTR_DACRST			(1 << 29)
 #define RCC_APB1RSTR_PWRRST			(1 << 28)
 #define RCC_APB1RSTR_CAN2RST			(1 << 26)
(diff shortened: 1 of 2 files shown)


ChuckM:
Sorry about that, I didn't want to confuse the QUADSPI and the RCC updates, so I split out the RCC changes and squashed it down to a single commit. Thanks for having a look at it @karlp

ChuckM:
Ok, another pair of eyes on this would be good. I don't think I've made any change to existing functionality that wasn't a bug, I've added some missing peripherals from the bigger chips into the various rcc enums.

ChuckM:
@davids5 if you could pull this version of rcc.h/rcc.c into your repo an have a look it would be helpful. One of the things I discovered after looking into it is that you can set the PLLR value after the PLL is running the system clock, it is only the n,q, and p values that are locked once the PLL is started. So the set function zeros n, p, q, and r, then replaces them with the values passed in. In my DSI code I just set R when when I'm setting up the DSI clock (but I also have my own clock setup since the existing rcc clock setup doesn't know anything about using the PLLSAI PLL to provide clocks for SDIO and USB.

lib/stm32/f4/rcc.c
```diff
@@ -569,22 +621,48 @@ void rcc_set_rtcpre(uint32_t rtcpre)
...
+	uint32_t reg32 = RCC_PLLCFGR;
+
+	/* mask out any previous values */
+	reg32 &= ~(
+		 RCC_PLLCFGR_PLLSRC |	/* 0 = HSI */
+		(RCC_PLLCFGR_PLLM_MASK << RCC_PLLCFGR_PLLM_SHIFT) |
+		(RCC_PLLCFGR_PLLN_MASK << RCC_PLLCFGR_PLLN_SHIFT) |
+		(RCC_PLLCFGR_PLLP_MASK << RCC_PLLCFGR_PLLP_SHIFT) |
+		(RCC_PLLCFGR_PLLQ_MASK << RCC_PLLCFGR_PLLQ_SHIFT) |
+		(RCC_PLLCFGR_PLLR_MASK << RCC_PLLCFGR_PLLR_SHIFT));
```
karlp: old behavior was to jsut replace, any particular reason to not do the same thing now?

lib/stm32/f4/rcc.c
```diff
@@ -569,22 +621,48 @@ void rcc_set_rtcpre(uint32_t rtcpre)
...
+	RCC_PLLCFGR = reg32 | (
+		(pllm << RCC_PLLCFGR_PLLM_SHIFT) |
 		(plln << RCC_PLLCFGR_PLLN_SHIFT) |
 		(((pllp >> 1) - 1) << RCC_PLLCFGR_PLLP_SHIFT) |
-		(pllq << RCC_PLLCFGR_PLLQ_SHIFT);
+		(pllq << RCC_PLLCFGR_PLLQ_SHIFT) |
+		(pllr << RCC_PLLCFGR_PLLR_SHIFT));
 }
 
+/* Note it adjusts PLLP */
```
karlp: just pllp? what?

ChuckM: PLLP has a "relationship" with its value, the value '0' means "divide by 2" the value "1" means divide by 4. So the code takes what you want pllp to do (2, 4, 6, 8) and adjusts it to the value that makes it do that. That code was already in the file, the only change to the API is to include the new pll "r" parameter.

lib/stm32/f4/rcc.c
ChuckM: No, but there are reserved bits. And the previous "just replace" code smashed the value of pll "R" which was a new use of the upper bits because it did not preserve the reserved bits.

davids5: What about SOIC without an R. Please look at the PR I did - it has 2 reserved values and if R is passed as 0 (a non legal value) it will not change the default value - this code will, and if passed 0 it will use the illegal value.

ChuckM: Good point @davids5 I'll update it.

davids5:
> @davids5 if you could pull this version of rcc.h/rcc.c into your repo an have a look it would be helpful. One of the things I discovered after looking into it is that you can set the PLLR value after the PLL is running the system clock, it is only the n,q, and p values that are locked once the PLL is started. So the set function zeros n, p, q, and r, then replaces them with the values passed in. In my DSI code I just set R when when I'm setting up the DSI clock (but I also have my own clock setup since the existing rcc clock setup doesn't know anything about using the PLLSAI PLL to provide clocks for SDIO and USB.

I am a fan of following the reference manual with it come to reserved bits and and changing locked values.  For all our applications the PLLs are set once and left that way. 

I am on a deadline to get a bootloader finished that uses the simple change I added. @ChuckM if you can vet your changes and ensure they do not break F4's without the PLL R - that's great, if not this code was tested on a STM32F427 and STMF469 (without / with PLLR) and it known working.

Therefore @karlp I would prefer the gets merged into master as is. Then @ChuckM is welcome to vet any changes he makes and submit his PR.

ChuckM:
Thanks to @davids5's eagle eye I've updated the code that updates the `PLLCFGR` register to be smart about updating R.

karlp:
you guys have two competing prs now on this right? can you guys sort out which of them you're trying to merge?

davids5:
see https://github.com/libopencm3/libopencm3/pull/677#issuecomment-236681349

karlp:
"Therefore @karlp I would prefer the gets merged into master as is. Then @ChuckM is welcome to vet any changes he makes and submit his PR." isn't clear to me what you're meaning I'm afraid. I feel there's words missing maybe?

ChuckM:
I believe (and @davids5 can correct me if I misspeak) that you can merge this PR first (#677), then David will pull it into his repo, and push just the USB changes in #676 which should be merged second (after updating). It separates the two into "RCC" fixes, and "USB" fixes.

davids5:
Oh Sorry!

s/"I would prefer the gets merged"/"I would prefer that [this] (https://github.com/libopencm3/libopencm3/pull/676) gets merged"/g

No actually - the opposite - I vetted [my PR] (https://github.com/libopencm3/libopencm3/pull/676) on both F427 and F469.

It doe just what is needed. To get the 446 and 469 booted with USB. @ChuckM RCC improvements can them be added on top of that.

davids5:
@ChuckM are you OK with that?

ChuckM:
@davids5 @karlp David, you are in the driver's seat here with respect to testing. I do not have a USB test case. I know that my DSI code does not work with your version of rcc.h. But I don't know if your USB code works with my version of rcc.h. My proposal is that you pull over this copy of rcc.h, put it into your environment, and test your USB code. If it still works, then you can update your PR with just the USB fixes and the final result with work with both your code and my code. With your proposal, merging your RCC and USB changes will get your code to work, then merging my RCC changes will get my code to work, but it may break your code. That was why I suggested up above that you pull this version of rcc.h and test it in your environment.

davids5:
@ChuckM - The makes sense.  I will test it in the AM

davids5:
@karlp, @ChuckM  - This has been tested with Just my [USB changes] (https://github.com/libopencm3/libopencm3/pull/681) and is good to go. I validated the logic in void rcc_set_main_pll_hse and it does the correct thing wrt pllr.

Please merge this PR

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -665,6 +760,7 @@ enum rcc_periph_clken {
 	RCC_SPI6	= _REG_BIT(0x44, 21),/* F2xx, F3xx */
 	RCC_SAI1EN	= _REG_BIT(0x44, 22),/* F42x, F43x */
 	RCC_LTDC	= _REG_BIT(0x44, 26),/* F42x, F43x */
+	RCC_DSI		= _REG_BIT(0x44, 27),/* F4x9, F4x9 */
```
karlp: You added some more ioports earlier, were there more to add here too? (port J and K)

lib/stm32/f4/rcc.c
```diff
@@ -624,6 +624,11 @@ void rcc_set_main_pll_hsi(uint32_t pllm, uint32_t plln, uint32_t pllp,
 			  uint32_t pllq, uint32_t pllr)
 {
 	uint32_t reg32 = RCC_PLLCFGR;
+	uint8_t		new_r;
+
+	/* If passed in value is legal, use it, else use version from register */
+	new_r = (pllr > 1) ? pllr :
```
karlp: greater than 1 or greater than 0?

lib/stm32/f4/rcc.c
```diff
@@ -512,10 +512,10 @@ void rcc_osc_off(enum rcc_osc osc)
 		RCC_CSR &= ~RCC_CSR_LSION;
 		break;
 	case RCC_PLLSAI:
-		RCC_CSR &= ~RCC_CR_PLLSAION;
+		RCC_CR &= ~RCC_CR_PLLSAION;
 		break;
 	case RCC_PLLI2S:
-		RCC_CSR &= ~RCC_CR_PLLI2SON;
+		RCC_CR &= ~RCC_CR_PLLI2SON;
```
karlp: Can you fix these in the first commit in the series? they were only just introduced weren't they?

lib/stm32/f4/rcc.c
davids5: No 0 and 1 are ileagal

davids5:
@karlp are you familiar with this https://github.com/blog/2141-squash-your-commits  ?

see "Only allow squash commits" I think it will ensure you get what you want and the PR make it in.

include/libopencm3/stm32/f4/rcc.h
ChuckM: GPIO J and K were already in the file as RCC_xxx defines but that integration was incomplete apparently. I fixed the missing bits.

ChuckM:
I have squashed it into a single commit for you're convenience @karlp

karlp:
@davids5 I am familiar with github's "squash everything" option, but that wasn't what I was after here.  I was quite happy with two commits in this series, but the third one was simply fixing the first one,  that was all I intended to be squished.  It's now squished all of it, which is now bigger than they needed to be, but ok, so be it.  Ideally I want single, self contained commits.  I don't need to squish everything in a feature branch into a single commit, and I honestly think github did a bit of a disservice there.  I'm sure it helps with some PRs, that have a single issue that keeps getting amended, but that wasn't the case here, and is rarely the case in this repo.

davids5:
@karlp - Thank you for clarifying that. Will this be merged?

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -477,17 +530,24 @@
 
 #define RCC_SSCGR_SSCGEN			(1 << 31)
 #define RCC_SSCGR_SPREADSEL			(1 << 30)
-/* RCC_SSCGR[27:16]: INCSTEP */
-#define RCC_SSCGR_INCSTEP_SHIFT			16
+/* RCC_SSCGR[27:13]: INCSTEP */
+#define RCC_SSCGR_INCSTEP_SHIFT			13
+#define RCC_SSCGR_INCSTEP_MASK			0xffff
```
karlp: shouldn't this be 0x7fff? just reviewing before pulling in here locally.

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -477,17 +530,24 @@
...
 /* RCC_SSCGR[15:0]: MODPER */
-#define RCC_SSCGR_MODPER_SHIFT			15
+#define RCC_SSCGR_MODPER_SHIFT			0
+#define RCC_SSCGR_MODPER_MASK			0xfff
```
karlp: and this should be 0x1fff from RM0386rev2?

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -499,17 +559,45 @@
...
+
+#define RCC_DCKCFGR_PLLI2SDIVQ_SHIFT		0
+#define RCC_DCKCFGR_PLLI2SDIVQ_MASK		0x1f
+
+#define RCC_DCKCFGR_PLLSAIDIVR_SHIFT		16
+#define RCC_DCKCFGR_PLLSAIDIVR_MASK		0x3
+#define RCC_DCKCFGR_PLLSAIDIVR_DIVR_2		0x0
+#define RCC_DCKCFGR_PLLSAIDIVR_DIVR_4		0x1
+#define RCC_DCKCFGR_PLLSAIDIVR_DIVR_8		0x2
+#define RCC_DCKCFGR_PLLSAIDIVR_DIVR_16		0x3
```
karlp: I'll probably move this back up above PLLSAIDIVQ to keep the bits in order if that's ok with you

karlp: Where did you even get this definitions from? the PLLSAIDIVR isn't in the latest refman for f46x/f47x?

include/libopencm3/stm32/f4/rcc.h
ChuckM: yup its 15 bits, not 16. I miscounted. `0x7fff` is the correct mask.

karlp:
I've merged in all the rcc bit updates and the PLLSAI/I2S helpers.  Still need to look over the PLLR stuff sorry

karlp:
I believe all the remaining code here (the pllr routine additions) are addressed in https://github.com/libopencm3/libopencm3/pull/690 where I rebased this.  Please correct me if that's wrong.

PR #678, opened 2016-07-31
Simplify TOG_SET_REG_BIT_MSK_AND_SET

urjaman:
A xor B with B=0 is A, thus this should be correct and faster.

I found this in my tree as I resumed hacking, so I kinda hope that I missed something here and this simplification is wrong :P, but ... wtf?


karlp:
merged in 62b02c515fc8195eae2155921ee283eddd1cac9e thanks!

PR #680, opened 2016-08-05
stm32/f4: dsi: DSI Peripheral include file

ChuckM:
Adds the dsi.h file to the f4 path of the library.

### DIFF
--- a/include/libopencm3/stm32/dsi.h
+++ b/include/libopencm3/stm32/dsi.h
@@ -1,0 +1,25 @@
+/* This provides unification of code over STM32 subfamilies */
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#if defined(STM32F4)
+#       include <libopencm3/stm32/f4/dsi.h>
+#else
+#       error "dsi.h not available for this family."
+#endif
+
--- a/include/libopencm3/stm32/f4/dsi.h
+++ b/include/libopencm3/stm32/f4/dsi.h
@@ -1,0 +1,814 @@
+/*
+ * STM32F4 DSI Host Defines
+ *
+ * Copyright (C) 2016, Chuck McManis <cmcmanis@mcmanis.com>
+ *
+ * This file is part of the libopencm3 project.
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/stm32/memorymap.h>
+
+#ifndef DSI_H
+#define DSI_H
+
+/*
+ * A couple of helper macros to avoid copy pasta code
+ * For multi-bit fields we can use these macros to create
+ * a unified SET_<reg>_<field> version and a GET_<reg_<field> version.
+ */
+#define DSI_GET(reg, field, x) \
+    (((x) >> DSI_ ## reg ## _ ## field ##_SHIFT) & DSI_ ## reg ##_ ## field ##_MASK)
+
+#define DSI_SET(reg, field, x) \
+    (((x) & DSI_ ## reg ##_## field ##_MASK) << DSI_## reg ##_## field ##_SHIFT)
+
+#define DSI_MASK(reg, field) \
+    ((DSI_ ## reg ##_## field ##_MASK) << DSI_## reg ##_## field ##_SHIFT)
+
+/*
+ * DSI Host Version Register
+ */
+#define DSI_VR				MMIO32(DSI_BASE + 0x0U)	
+
+/*
+ * DSI Host Control Register
+ */
+#define DSI_CR				MMIO32(DSI_BASE + 0x4U)
+#define DSI_CR_EN			(1 << 0)
+
+/*
+ * DSI Host Clock Control Register
+ */
+#define DSI_CCR				MMIO32(DSI_BASE + 0x8U)
+#define DSI_CCR_TOCKDIV_SHIFT		8
+#define DSI_CCR_TOCKDIV_MASK		0xff
+#define DSI_CCR_TXECKDIV_SHIFT		0
+#define DSI_CCR_TXECKDIV_MASK		0xff
+
+/*
+ * DSI Host LTDC VCID Register
+ */
+#define DSI_LVCIDR			MMIO32(DSI_BASE + 0xcU)
+#define DSI_LVCIDR_VCID_SHIFT		0
+#define DSI_LVCIDR_VCID_MASK		0x3
+
+/*
+ * DSI Host LTDC Color Coding Register
+ */
+#define DSI_LCOLCR			MMIO32(DSI_BASE + 0x10U)
+#define DSI_LCOLCR_LPE			(1 << 8)
+#define DSI_LCOLCR_COLC_SHIFT		0
+#define DSI_LCOLCR_COLC_MASK		0xf
+
+/*
+ * DSI Host LTDC Polarity Configuration Register
+ */
+#define DSI_LPCR			MMIO32(DSI_BASE + 0x14U)
+#define DSI_LPCR_HSP			(1 << 2)
+#define DSI_LPCR_VSP			(1 << 1)
+#define DSI_LPCR_DEP			(1 << 0)
+
+/*
+ * DSI Host Low-power Configuration Register
+ */
+#define DSI_LPMCR			MMIO32(DSI_BASE + 0x18U)
+#define DSI_LPMCR_LPSIZE_SHIFT		16
+#define DSI_LPMCR_LPSIZE_MASK		0xff
+#define DSI_LPMCR_VLPSIZE_SHIFT		0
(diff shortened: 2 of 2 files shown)


include/libopencm3/stm32/dsi.h
```diff
@@ -0,0 +1,25 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#if defined(STM32F4)
+#       include <libopencm3/stm32/f4/dsi.h>
+#else
+#       error "quadspi.h not available for this family."
```
karlp: looks like the wrong wording in the error?

include/libopencm3/stm32/dsi.h
```diff
@@ -0,0 +1,25 @@
+/* This provides unification of code over STM32F subfamilies */
```
karlp: I know it's all over the place, but we should try and remember the L devices too one day :)  there's also T and there was W ;)

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,814 @@
...
+ * a unified SET_<reg>_<field> version and a GET_<reg_<field> version.
+ */
+#define DSI_GET(reg, field, x) \
+    (((x) >> DSI_ ## reg ## _ ## field ##_SHIFT) & DSI_ ## reg ##_ ## field ##_MASK)
+
+#define DSI_SET(reg, field, x) \
+    (((x) & DSI_ ## reg ##_## field ##_MASK) << DSI_## reg ##_## field ##_SHIFT)
+
+#define DSI_MASK(reg, field) \
+    ((DSI_ ## reg ##_## field ##_MASK) << DSI_## reg ##_## field ##_SHIFT)
```
karlp: I would greatly prefer to see this sort of preprocessor magic left in peoples examples or drivers, not in the common headers.  They don't exist anywhere else in the library, and I don't personally want to start spreading them around.  You're not using them anywhere else in this file, and honestly, the usage of the macros is not obvious and there's no documentation.

include/libopencm3/stm32/dsi.h
ChuckM: Yes, I have fixed both errors (the STM32{F} and the quadspi in the error string.

include/libopencm3/stm32/f4/dsi.h
ChuckM: Except that if you take the examples which would potentially follow this commit into the examples repo there would be number of files showing how they are used. I'm inclined to leave them in rather than repeat them in example after example.

ChuckM:
Updated and squashed, I think this one is ready to go as well.

ChuckM:
@karlp I thought about your question about doxygen comments and you make a good point. I've added doxygen comments to the macros and the file correctly slots itself in group stm32fx_includes. I would appreciate it if you would take a look and tell me if that works for you. If it does I'll add similar comments to the other include files that have get/set macros in them.

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+#define DSI_VMCR_VMT_NON_BURST_PULSE	0x0
+#define DSI_VMCR_VMT_NON_BURSE_EVENT	0x1
+#define DSI_VMCR_VMT_BURST		0x2
+
+/*
+ * DSI Host Video Packet Configuration Register
+ */
+#define DSI_VPCR			MMIO32(DSI_BASE + 0x3CU)
+#define DSI_VPCR_VPSIZE_SHIFT		0
+#define DSI_VPCR_VPSIZE_MASK		0x3fff
```
karlp: I think this is 0x7fff?

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+
+/*
+ * DSI Host Video Chunks Configuration Register
+ */
+#define DSI_VCCR			MMIO32(DSI_BASE + 0x40U)
+#define DSI_VCCR_NUMC_SHIFT		0
+#define DSI_VCCR_NUMC_MASK		0x1fff
```
karlp: 0x3ff?

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+ */
+#define DSI_GHCR			MMIO32(DSI_BASE + 0x6CU)
+#define DSI_GHCR_WCMSB_SHIFT		16
+#define DSI_GHCR_WCMSB_MASK		0xff
+#define DSI_GHCR_WCLSB_SHIFT		8
+#define DSI_GHCR_WCLSB_MASK		0xff
+#define DSI_GHCR_DATA1_SHIFT		16	/* data 1 in 'short' mode */
+#define DSI_GHCR_DATA1_MASK		0xff
+#define DSI_GHCR_DATA0_SHIFT		8	/* data 0 in 'short' mode */
+#define DSI_GHCR_DATA0_MASK		0xff
```
karlp: Would it be more clear to do
```
#define DSI_GHCR_DATA0_SHIFT DSI_GHCR_WCLSB_SHIFT
etc
```

instead?  I was looking at this with no idea where you'd gotten these DATA0/1 names from at all! :)

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+ */
+#define DSI_GPDR			MMIO32(DSI_BASE + 0x70U)
+#define DSI_GPDR_BYTE4_SHIFT		24
+#define DSI_GPDR_BYTE4_MASK		0xff
+#define DSI_GPDR_BYTE3_SHIFT		16
+#define DSI_GPDR_BYTE3_MASK		0xff
+#define DSI_GPDR_BYTE2_SHIFT		8
+#define DSI_GPDR_BYTE2_MASK		0xff
+#define DSI_GPDR_BYTE1_SHIFT		0
+#define DSI_GPDR_BYTE1_MASK		0xff
```
karlp: Will people actually use them like this?  Or would this be more useful?
```
#define DSI_GPDR_BYTEx_SHIFT ((x) * 8)
#define DSI_GPDR_BYTEx_MASK 0xff
```
Not real concerned either way personally.

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+#define DSI_GPSR_RCB			(1 << 6)
+#define DSI_GPSR_PRDFF			(1 << 5)
+#define DSI_GPSR_PRDFE			(1 << 4)
+#define DSI_GPSR_PWRFF			(1 << 3)
+#define DSI_GPSR_PWRFE			(1 << 2)
+#define DSI_GPSR_CMDFF			(1 << 1)
+#define DSI_GPSR_CMDFE			(1 << 0)
+
+/*
+ * DSI Host Tineout Counter Configuration Register
```
karlp: typo in timeout

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+
+/*
+ * DSI Host LTDC Current VCID Register
+ */
+#define DSI_LCVCIDR			MMIO32(DSI_BASE + 0x10CU)
+#define DSI_LCVCIDR_VCID_SHIFT		0
+#define DSI_LCVCIDR_VCID_MASK		0x3
+
+/*
+ * DSI Host LTCD Curren Color Coding Register
```
karlp: typo in current

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+ */
+#define DSI_WPCR3			MMIO32(DSI_BASE + 0x424U)
+#define DSI_WPCR3_TLPXC_SHIFT		24
+#define DSI_WPCR3_TLPXC_MASK		0xff
+#define DSI_WPCR3_THSEXIT_SHIFT		16
+#define DSI_WPCR3_THSEXIT_MASK		0xff
+#define DSI_WPCR3_TLPXD_SHIFT		8
+#define DSI_WPCR3_TLPXD_MASK		0xff
+#define DSI_WPCR3_THSZERO_SHIFT		0
+#define DSI_WPCR3_THSZERO_MASK		0xff
```
karlp: It's somewhat tempting to define some of these registers as MMIO8 with the right offsets.   Or are they always written in one go anyway?  Just a silly idea?

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+
+/*
+ * DSI Wrapper PHY Configuration Register 4
+ */
+#define DSI_WPCR4			MMIO32(DSI_BASE + 0x428U)
+#define DSI_WPCR4_TCLKPOST_SHIFT	0
+#define DSI_WPCR4_TCLKPOST_MASK		0xff
```
karlp: If you can tell your FAE that the register map is wrong here too, it has a copy of THSZERO here, but you rightly have TCLKPOST as the register descriptions have.

include/libopencm3/stm32/f4/dsi.h
```diff
@@ -0,0 +1,897 @@
...
+#define DSI_WRPCR_ODF_DIV_8		3
+#define DSI_WRPCR_IDF_SHIFT		11
+#define DSI_WRPCR_IDF_MASK		0xf
+#define DSI_WRPCR_IDF_DIV_1		0
+#define DSI_WRPCR_IDF_DIV_2		2
+#define DSI_WRPCR_IDF_DIV_3		3
+#define DSI_WRPCR_IDF_DIV_4		4
+#define DSI_WRPCR_IDF_DIV_5		5
+#define DSI_WRPCR_IDF_DIV_6		6
+#define DSI_WRPCR_IDF_DIV_7		7
```
karlp: Is it really worth defining these?  It's just the native number, and they've "protected" you from divide by zero.

include/libopencm3/stm32/f4/dsi.h
ChuckM: This register is "dual use", when you send a "short" generic packet (2 or fewer parameters), you include DATA0 and DATA1 in this register, when you send a "long" generic packet (3 or more parameters) then you write the length of the packet in here, and push the parameters into GPDR.

karlp: Yeah, I figured that out, was just curious whether it made more sense to use the same defines, rather than repeating them, to hammer home that they are aliases for the same field. no biggy

include/libopencm3/stm32/f4/dsi.h
ChuckM: I added them because when debugging my code it was hard to remember they were "off by one" divides. So I'd see the constant '2' there and think "oh its dividing by 2" but really it was dividing by 3. I went around a couple of frustrating loops that @schodet got be out of by pointing out this error. When it is written out "DIV_2" I don't get it mixed up.

karlp: but it's not off by one.... not the IDF divs?  it just does div1 for 0 and div1 for 1, to protect you from divide by zero, but all the others are just "as is" but ok, I'll leave it.  I'll change div_1 to be 1 though, to be consistent with the rest, and also the manual which allows both

karlp:
I've queued this locally now

karlp:
merged via rebase with tweaks in 32f2b50

PR #682, opened 2016-08-09
[stm32f4-ltdc] this fixes ltdc_get_rgb888_from_rgb565 return value type

h2obrain:
... and adds ltdc_get_rgb565_from_rgb888

This small fix renames the misleading rgb888 argument to rgb565 and changes the return value type to uint32_t since 3*8 bits do not fit into 16 bits.

Additionally I added the opposite function.

(This bug was probably introduced during the conversion from macro to static inline..)

### DIFF
--- a/include/libopencm3/stm32/f4/ltdc.h
+++ b/include/libopencm3/stm32/f4/ltdc.h
@@ -487,15 +487,36 @@ static inline bool LTDC_SRCR_IS_RELOADING(void)
 }
 
 /**
- * color conversion helper function
+ * RGB565-to-RGB888 color conversion helper function
  * (simulate the ltdc color conversion)
+ * NOTE: always use this function to calculate rgb888 for color key
+ *       from rgb565 color!
  */
-
-static inline uint16_t ltdc_get_rgb888_from_rgb565(uint16_t rgb888)
+static inline uint32_t ltdc_get_rgb888_from_rgb565(uint16_t rgb565)
 {
-	return ((((rgb888) & 0xF800)   >> (11-8))/31)<<16
-	       | ((((rgb888) & 0x07E0) <<  (8-5))/63)<<8
-	       | ((((rgb888) & 0x001F) <<  (8-0))/31)<<0;
+	uint32_t r,g,b;
+	
+	r = ((((rgb565) & 0xF800) >> (11-8))/31);
+	g = ((((rgb565) & 0x07E0) <<  (8-5))/63);
+	b = ((((rgb565) & 0x001F) <<  (8-0))/31);
+
+	if (r==256) r = 255;
+	if (g==256) g = 255;
+	if (b==256) b = 255;
+
+	return (r<<16) | (g<<8) | b;
+}
+
+/**
+ * RGB888-to-RGB565 color conversion helper function
+ */
+static inline uint16_t ltdc_get_rgb565_from_rgb888(uint32_t rgb888)
+{
+	return (uint16_t)(
+		((rgb888 & 0xf80000) >> (16-11 + 8-5)) /* grab upper 5 bits of red   */
+	  | ((rgb888 & 0x00fc00) >> (8-5   + 8-6)) /* grab upper 6 bits of green */
+	  | ((rgb888 & 0x0000f8) >> (0-0   + 8-5)) /* grab upper 5 bits of blue  */
+	);
 }
 
 


schodet:
It does not seems to work, for example if blue is 31: (31 << 8) / 31 = 256, that's too much by 1.
In the other function if blue is 255 : (255 * 31) >> 8 => 30, that's too low by 1.
Did I miss something?

include/libopencm3/stm32/f4/ltdc.h
```diff
@@ -490,12 +490,20 @@ static inline bool LTDC_SRCR_IS_RELOADING(void)
...
-static inline uint16_t ltdc_get_rgb888_from_rgb565(uint16_t rgb888)
+static inline uint16_t ltdc_get_rgb565_from_rgb888(uint32_t rgb888)
 {
-	return ((((rgb888) & 0xF800)   >> (11-8))/31)<<16
-	       | ((((rgb888) & 0x07E0) <<  (8-5))/63)<<8
-	       | ((((rgb888) & 0x001F) <<  (8-0))/31)<<0;
+	return (uint16_t) (
+			 (((((rgb888 & 0xFF0000) >> 16)*31) << (11-8)) & 0xF800)
+	       | (((((rgb888 & 0x00FF00) >>  8)*63) >> (8-5) ) & 0x07E0)
+	       | (((((rgb888 & 0x0000FF) >>  0)*31) >> (8-0) ) & 0x001F)
```
ChuckM: Seems a bit like you are moving the bits more than needed, have you tried something like
```
    return (uint16_t)(
            ((rgb888 & 0xf80000) >> 16)  | /* grab upper 5 bits of red)
            ((rgb888 & 0xfc00) >> 5) | /* grab 6 bits of green */
            ((rgb888 & 0xf8) >> 3); /* grab 5 bits of blue */
    );
```

H2OBrain:
okay, I did some simulation for possible conversion functions in python
```python
def rgb888_from_rgb565_old(rgb565) :
    return (((((rgb565) & 0xF800) >> (11-8))/31)<<16) \
         | (((((rgb565) & 0x07E0) <<  (8-5))/63)<<8)  \
         | (((((rgb565) & 0x001F) <<  (8-0))/31)<<0)

def rgb888_from_rgb565_standard(rgb565) :
    return (((((rgb565 & 0xF800) >> 11) * 255 + 15) / 31) << 16) \
         | (((((rgb565 & 0x07E0) >>  5) * 255 + 31) / 63) << 8)  \
         | (((((rgb565 & 0x001F) >>  0) * 255 + 15) / 31) << 0)

def rgb888_from_rgb565_simplest(rgb565) :
    return ((rgb565 & 0xF800) << (16 - 11 + 8 - 5)) \
         | ((rgb565 & 0x07E0) << (8  -  5 + 8 - 6)) \
         | ((rgb565 & 0x001F) << (0  -  0 + 8 - 5))

def rgb888_from_rgb565_old_fixed(rgb565) :
    r = ((((rgb565) & 0xF800) >> (11-8))/31)
    g = ((((rgb565) & 0x07E0) <<  (8-5))/63)
    b = ((((rgb565) & 0x001F) <<  (8-0))/31)
    
    if (r==256) : r = 255;
    if (g==256) : g = 255;
    if (b==256) : b = 255;
    
    return (r<<16) | (g<<8) | b

def rgb565_from_rgb888_simplest(rgb888) :
    return  ((rgb888 & 0xf80000) >> (16 - 11 + (8-5))) \
          | ((rgb888 & 0x00fc00) >> (8  -  5 + (8-6))) \
          | ((rgb888 & 0x0000f8) >> (0  -  0 + (8-5)))

def test_conversion(rgb565) :
    print "##################################################"
    print "conversion of 0x%04X" %rgb565
    print "0x%06X old"           %rgb888_from_rgb565_old(rgb565)
    print "0x%06X standard"      %rgb888_from_rgb565_standard(rgb565)
    print "0x%06X simplest"      %rgb888_from_rgb565_simplest(rgb565)
    print "0x%06X old fixed"     %rgb888_from_rgb565_old_fixed(rgb565)
    print
    rgb888 = rgb888_from_rgb565_standard(rgb565)
    print "conversion of 0x%06X" %rgb888
    print "0x%04X simplest"      %rgb565_from_rgb888_simplest(rgb888)
    print

# run tests
test_conversion(0x01ad)
test_conversion(0)
test_conversion(0xffff)
```
results :
```
##################################################
conversion of 0x01AD
0x00346B old
0x00356B standard
0x003468 simplest
0x00346B old fixed

conversion of 0x00356B
0x01AD simplest

##################################################
conversion of 0x0000
0x000000 old
0x000000 standard
0x000000 simplest
0x000000 old fixed

conversion of 0x000000
0x0000 simplest

##################################################
conversion of 0xFFFF
0x1010100 old
0xFFFFFF standard
0xF8FCF8 simplest
0xFFFFFF old fixed

conversion of 0xFFFFFF
0xFFFF simplest
```

the fixed PR uses `rgb888_from_rgb565_old_fixed` as rgb565-to-rgb888 function and `rgb565_from_rgb888_simplest` as rgb888-to-rgb565 function. to avoid future 'fixes' i also added the more standard rgb565-to-rgb888 conversion function `rgb888_from_rgb565_standard`

include/libopencm3/stm32/f4/ltdc.h
```diff
@@ -489,13 +489,43 @@ static inline bool LTDC_SRCR_IS_RELOADING(void)
...
+
+	if (r==256) r = 255;
+	if (g==256) g = 255;
+	if (b==256) b = 255;
 
-static inline uint16_t ltdc_get_rgb888_from_rgb565(uint16_t rgb888)
+	return (r<<16) | (g<<8) | b;
+}
+/**
+ * widely used conversion formula for rgb888-to-rgb565 conversion
```
karlp: to or from?

karlp:
Is it _immediately_ clear to the user what they should use in all cases?  Would it perhaps make more sense to drop these from the headers and include them in some example code instead?  Where it can be more clear what version you're using where and why?

include/libopencm3/stm32/f4/ltdc.h
H2OBrain: ups, i meant from
this function could be dropped to avoid confusion

H2OBrain:
Set the color key

Use case 1 - __Select a color-key in rgb565__
- Get rgb888 using __ltdc_get_rgb888_from_rgb565__ to set color key. (The hardware will convert this automatically to rgb565 (or active color format))
- Fill the buffer with the selected color-key

Use case 2 - __Select a color-key in rgb888__
- Set color key (The hardware will convert this automatically to rgb565 (or active color format))
- Get rgb565 using ltdc_get_rgb565_from_rgb888
- Fill the buffer with the converted color-key

karlp:
was anything wrong with this? I realise it fell through the floor, I've never had any devices with screens to even try this, but it doesn't look like anythign was actually wrong with this?

h2obrain:
Ups.. I did some spring cleaning and did not realise this was still open..
I think this function is useful to have, since you need to set the color keys in rgb888 format.
The current function is broken, so I would fix or delete it.

PR #683, opened 2016-08-10
stm32: dma2d: Add include files for DMA2D

ChuckM:
Adds include files and updates memorymap.h file to
include base address of DMA2D


ChuckM:
Ok, this is the final version I'm using with my example code. One question came up, which was if the settings for the Foreground and Background PFC registers should be just one set of `#defines`. Some would say yes others no. I'm going back and forth.

include/libopencm3/stm32/f4/dma2d.h
```diff
@@ -0,0 +1,222 @@
...
+#define DMA2D_FGPFCCR_CM_RGB888		1
+#define DMA2D_FGPFCCR_CM_RGB565		2
+#define DMA2D_FGPFCCR_CM_ARGB1555	3
+#define DMA2D_FGPFCCR_CM_ARGB4444	4
+#define DMA2D_FGPFCCR_CM_L8		5
+#define DMA2D_FGPFCCR_CM_AL44		6
+#define DMA2D_FGPFCCR_CM_AL88		7
+#define DMA2D_FGPFCCR_CM_L4		8
+#define DMA2D_FGPFCCR_CM_A8		9
+#define DMA2D_FGPFCCR_CM_A4		10
```
ChuckM: The color mode (CM), CLUT color mode (CM), Alpha, and Alpha Mode (AM) fields are identical in the Foreground PFCCR and the Background PFCCR.  It could be argued they would be the same flags so maybe `DMA2D_xPFCCR_START` sort of like the layer code in the LTDC. Thoughts?

schodet: Personally, I prefer one macro. With two macros, after some copy pasting you risk to use the "wrong" one anyway.

davids5:
@ChuckM Are you on skype?

ChuckM:
Generally I prefer email, never did like Skype or facetime. @davids5 that is why I put my email address in my copyright comment :smile:

include/libopencm3/stm32/f4/dma2d.h
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <libopencm3/stm32/memorymap.h>
+#ifndef uint32_t
+#include <stdint.h>
+#endif
```
karlp: nope, we've never done this anywhere else, just toss this sort of thing and include stdint.h like we do elsewhere

include/libopencm3/stm32/f4/dma2d.h
```diff
@@ -0,0 +1,222 @@
...
+#define DMA2D_ISR_TEIF			(1 << 0)
+
+/* DMA2D Interrupt Flag Clear Register */
+#define DMA2D_IFCR			MMIO32(DMA2D_BASE + 0x8U)
+#define DMA2D_IFCR_CCEIF		(1 << 5)
+#define DMA2D_IFCR_CCTCIF		(1 << 4)
+/* this is what the reference manual specifies */
+#define DMA2D_IFCR_CAECIF		(1 << 3)
+/* this is what it "should be" */
+#define DMA2D_IFCR_CCAEIF		(1 << 3)
```
karlp: cute :)  What is the bet that an updated refman will fix it?

include/libopencm3/stm32/f4/dma2d.h
karlp: DMA2X_xPFCCR_ looks good to me.

include/libopencm3/stm32/f4/dma2d.h
ChuckM: Actually I know for a fact that several will be fixed, as I was doing my code I was feeding documentation bugs to one of ST's field engineers who assured me they would get to the right technical pubs people.

karlp:
So, this all looks good, but again, I'm not going to merge those preprocessor string conversion macros.  A single mistyped definition and you get a collapsed jungle of errors inside preprocessor expansions that you can't grep for, we've never needed them before, and they're the sort of optimization thinking that we've never done here in the past.  At _most_ they belong in some example code using the definitions, but not in the common permanent export header.

karlp:
Can you drop the dupe of the foreground/background stuff and I'll merge all the rest.

ChuckM:
will do

ChuckM:
dupes dropped

karlp:
Thanks for these, I also dropped the prepreocessor macros as before however. merged in b897397

PR #688, opened 2016-08-15
L1 rcc power staging2

karlp:
A second staging branch.  Have another niggling power issue on l1 to solve and test further.  This just pulls in some of the simplest and unquestionable bits of header updates.  Chuck, this uses the different definitions of INCSTEP and MODPER I mentioned.

Supersedes https://github.com/libopencm3/libopencm3/pull/634


karlp:
merged with updates from other PRs

PR #689, opened 2016-08-16
Drafts/l1 rcc power3

karlp:
minor cleanup while working on some rcc bugs.  includes a breaking change for f2 users, that should have hit them months ago


karlp:
original contents of this branch have merged, the remainder is a separete issue.

PR #690, opened 2016-08-16
stm32f4: rcc: support new plls for new f4 parts

karlp:
Revise the PLL inits to support new and old PLL configurations,
particularly to support F4x9 devices.

(This is a rebased and slightly reformatted version of the remainder of https://github.com/libopencm3/libopencm3/pull/677  I cherry picked out the earlier parts, so I've rebased it for you here, I hope to get to this in the next few days)

PR #691, opened 2016-08-19
Drafts/f469 rcc pllr alternate

karlp:
Alternate update tweak for review, see also https://github.com/libopencm3/libopencm3/issues/690

### DIFF
--- a/lib/stm32/f4/rcc.c
+++ b/lib/stm32/f4/rcc.c
@@ -562,6 +562,43 @@ void rcc_osc_bypass_disable(enum rcc_osc osc)
 	}
 }
 
+
+/**
+ * Set the dividers for the PLLSAI clock outputs
+ * divider p is only available on F4x9 parts, pass 0 for other parts.
+ * @param n valid range is 49..432
+ * @param p 0 if unused, @ref rcc_pllsaicfgr_pllsaip
+ * @param q valid range is 2..15
+ * @param r valid range is 2..7
+ * @sa rcc_pllsai_postscalers
+ */
+void rcc_pllsai_config(uint16_t n, uint16_t p, uint16_t q, uint16_t r)
+{
+	RCC_PLLSAICFGR = (
+	  ((n & RCC_PLLSAICFGR_PLLSAIN_MASK) << RCC_PLLSAICFGR_PLLSAIN_SHIFT) |
+	  ((p & RCC_PLLSAICFGR_PLLSAIP_MASK) << RCC_PLLSAICFGR_PLLSAIP_SHIFT) |
+	  ((q & RCC_PLLSAICFGR_PLLSAIQ_MASK) << RCC_PLLSAICFGR_PLLSAIQ_SHIFT) |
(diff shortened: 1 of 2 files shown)


ChuckM:
:+1:

lib/stm32/f4/rcc.c
```diff
@@ -653,60 +653,52 @@ void rcc_set_rtcpre(uint32_t rtcpre)
...
-		(RCC_PLLCFGR_PLLR_MASK << RCC_PLLCFGR_PLLR_SHIFT));
-
-	/* add back new values, PLLSRC is HSI */
-	RCC_PLLCFGR = reg32 | (
+	/* Use reset value if not legal, for parts without pllr */
+	if (pllr < 2) {
+		pllr = 2;
+	}
+	RCC_PLLCFGR = 0 | /* HSI */
 		(pllm << RCC_PLLCFGR_PLLM_SHIFT) |
```
ChuckM: One of the things I appreciate about my "SET" macros is that they take the passed in value and mask it against the known width of the bitfield. So in this case the code would be:
```
    RCC_PLLCFGR = 0 | 
           RCC_SET(PLLCFGR, PLLM, pllm) |
           RCC_SET(PLLCFGR, PLLN, plln) |
           RCC_SET(PLLCFGR, PLLP, ((pllp >> 1) - 1)) |
           RCC_SET(PLLCFGR, PLLR, pllr);
```
And those macros would protect against two things, one sending in a value that was "too large" for the divisor which you might do accidently, or sending in 0 for pllp which results in pllp being -1 and stomping all the bits above the PLLP bitfield. 

At a minimum I would suggest you change this to:
```
                ((pllm & RCC_PLLCFGR_PLLM_MASK) << RCC_PLLCFGR_PLLM_SHIFT)
```
for each of the values to avoid smashing bits you don't want to smash.

karlp:
Adding the mask is valid, sure, that was just existing code, from both your and davids change, I didn't touch that bit.

lib/stm32/f4/rcc.c
```diff
@@ -546,6 +562,43 @@ void rcc_osc_bypass_disable(enum rcc_osc osc)
...
+ * @sa rcc_pllsai_postscalers
+ */
+void rcc_pllsai_config(uint16_t n, uint16_t p, uint16_t q, uint16_t r)
+{
+	RCC_PLLSAICFGR = (
+	  ((n & RCC_PLLSAICFGR_PLLSAIN_MASK) << RCC_PLLSAICFGR_PLLSAIN_SHIFT) |
+	  ((p & RCC_PLLSAICFGR_PLLSAIP_MASK) << RCC_PLLSAICFGR_PLLSAIP_SHIFT) |
+	  ((q & RCC_PLLSAICFGR_PLLSAIQ_MASK) << RCC_PLLSAICFGR_PLLSAIQ_SHIFT) |
+	  ((r & RCC_PLLSAICFGR_PLLSAIR_MASK) << RCC_PLLSAICFGR_PLLSAIR_SHIFT));
+}
```
ChuckM: This one has the masks

PR #692, opened 2016-08-19
stm32l1: rcc: reliably clock up/down

karlp:
When changing the system clock, you must take care to not exceed the
legal ranges based on voltage and flash wait states.

Existing code made it possible to provide a valid clock structure, that
would run out of bounds temporarily.  Some boards would crash with
various Usage faults / Bus errors due to this.

https://github.com/karlp/libopencm3-tests/tree/master/tests/rcc-legal-ranges provides a test case that fails on most stm32l boards without this change.

Ideally, this needs to be extended to cover the PLL clocks, and then further out to other families.  For L1 at least, MSI is valid for all speeds in all ranges, only wait states might be questionable.

### DIFF
--- a/lib/stm32/l1/rcc.c
+++ b/lib/stm32/l1/rcc.c
@@ -497,31 +497,46 @@ void rcc_clock_setup_msi(const struct rcc_clock_scale *clock)
 	rcc_apb2_frequency = clock->apb2_frequency;
 }
 
+
+/**
+ * Switch sysclock to HSI with the given parameters.
+ * This should be usable from any point in time, but only if you have used
+ * library functions to manage clocks.  It relies on the global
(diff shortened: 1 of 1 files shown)


ChuckM:
This looks like the right fix to me. I've tried it on the 469, 429, and 411 processors.

karlp:
did you port that change locally? it's not code used on any f4 parts, so it shouldn't have changed anything there anway?

ChuckM:
On the F4's is the flash wait states that screws them up. (not as problem going down through, but speeding up it hangs). Although on the 446 (which I didn't try) also has a wider voltage scaling range.

karlp:
Still need to apply more of this to other clock setups, but more improvements more time

PR #695, opened 2016-09-02
stm32f4: rcc: fix setup of main system clock with pllp

marcushv:
Fixes: 57c2b00a69f97205313e1c7ab8116ee1893b231e

There was an issue with the pllp value calculation where the masking was done
in the wrong place. The pllp value was always equivalent to 2 (the bits were
always set to 0b00) which could result in an undesired system clock frequency.


karlp:
Thanks for catching this.  The masks were added last minute after some IRC discussions. :(

marcushv:
I understand. It's a good addition though ;)

PR #696, opened 2016-09-04
Add convenience function for hsi pll config on stm32f4

kliment:
Same format as the hse pll setup function, and added 96MHz as a frequency

### DIFF
--- a/lib/stm32/f4/rcc.c
+++ b/lib/stm32/f4/rcc.c
@@ -49,7 +49,6 @@ uint32_t rcc_ahb_frequency = 16000000;
 uint32_t rcc_apb1_frequency = 16000000;
 uint32_t rcc_apb2_frequency = 16000000;
 
-#define rcc_hsi_3v3 rcc_hse_16mhz_3v3
 
 const struct rcc_clock_scale rcc_hse_8mhz_3v3[RCC_CLOCK_3V3_END] = {
 	{ /* 48MHz */
--- a/lib/stm32/f4/rcc.c
+++ b/lib/stm32/f4/rcc.c
@@ -49,6 +49,8 @@ uint32_t rcc_ahb_frequency = 16000000;
 uint32_t rcc_apb1_frequency = 16000000;
 uint32_t rcc_apb2_frequency = 16000000;
 
+#define rcc_hsi_3v3 rcc_hse_16mhz_3v3
+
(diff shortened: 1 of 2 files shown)


lib/stm32/f4/rcc.c
```diff
@@ -49,6 +49,8 @@ uint32_t rcc_ahb_frequency = 16000000;
 uint32_t rcc_apb1_frequency = 16000000;
 uint32_t rcc_apb2_frequency = 16000000;
 
+#define rcc_hsi_3v3 rcc_hse_16mhz_3v3
```
karlp: what's this for?

kliment: Just commented on this in the IRC channel, but here it is again for the
record:

The f4 HSI always runs at 16MHz, so instead of replicating the 16MHz
presets for rcc_hsi_3v3, I aliased rcc_hsi_3v3 to rcc_hse_16mhz_3v3

lib/stm32/f4/rcc.c
```diff
@@ -307,6 +357,22 @@ const struct rcc_clock_scale rcc_hse_25mhz_3v3[RCC_CLOCK_3V3_END] = {
 		.apb1_frequency = 42000000,
 		.apb2_frequency = 84000000,
 	},
+    { /* 96MHz */
+		.pllm = 25,
+		.plln = 192,
```
karlp: What's the idea here? HSI is at 16MHz, VCO input must be between 1 and 2Mhz, and is recommended to be 2Mhz to reduce jitter.  16/25 is < 1 MHz.  It looks sort of like you copied this from a 25MHz crystal config?

kliment: Sorry, I should have made this two separate commits. So, I did two
distinct things:

One - add an HSI PLL config convenience function and alias the 16MHz HSE
presets to rcc_hsi_3v3

Two - add a 96MHz preset to ALL the HSE preset sets, including the 25MHz
one you're looking at here.

karlp: hang on, is this just more clock configs for 96mhz, mixed in with a commit that is for enabling convenience functions for HSI?  that's confusing.  Also, what's the particular motivation for 96mhz configs anyway?  Is that the max speed of any parts?  Custom configs like this can just be created in user code.

kliment: The motivation for 96MHz is that it's a fairly common frequency that
seems to be missing from the list. And yes, it should have been a
separate commit.

PR #697, opened 2016-09-09
BREAKING: stm32: adc: extract common calibration code

karlp:
@DanielG and @kungjohan This will affect both of you.
Down the road, I'd like to update the f1 code to use this api too.

Extract the calibration code from the f0, and share it with the other
adc-v2 peripheral users (f0,l0,f3,l4)

Uses the same naming set of is/async naming conventions requested by the
RTOS guys instead of having blocking only calls.

Old code:
	adc_calibrate_start(ADC);
	adc_calibrate_wait_finish(ADC);

New code (blocking):
	adc_calibrate(ADC);

New code (asynch):
	adc_calibrate_asynch(ADC);
	// do stuff
	adc_is_calibrate(ADC);  // will be false when it's finished.


karlp:
Actually, should I just include the f1 calibration change in this too?

lib/stm32/common/adc_common_v2.c
```diff
@@ -147,6 +147,37 @@ void adc_power_off(uint32_t adc)
...
+{
+	ADC_CR(adc) = ADC_CR_ADCAL;
+}
+
+/**
+ * Is the ADC Calibrating?
+ * @param adc ADC Block register address base @ref adc_reg_base
+ * @return true if the adc is currently calibrating
+ */
+bool adc_is_calibrate(uint32_t adc)
```
DanielG: That doesn't really sound like English, wouldn't `adc_is_calibrating()` be better?

karlp: absolutely, I had that first.  I was just thinking of leaving it with the same word "calibrate" all the time, but yeah, I think my first choice, "is_calibrating" is indeed better now :)

karlp:
Added F1 to the mix.  Would it be majorly important to mark the old ones as deprecated in the short term? or are breaking api changes still ok?

PR #699, opened 2016-09-11
stm32 adc calibration api standardization

karlp:
Version of https://github.com/libopencm3/libopencm3/pull/697 with the old functions preserved and marked deprecated instead.  The deprecated versions can be removed "in the future"  Seems better to me?


ChuckM:
:+1:  This looks like a good compromise to me.

PR #702, opened 2016-09-21
stm32: add dma_get_number_of_data

DanielG:
Just something that's convenient when you need to know how much has been transferred so far.

### DIFF
--- a/include/libopencm3/stm32/common/dma_common_l1f013.h
+++ b/include/libopencm3/stm32/common/dma_common_l1f013.h
@@ -410,6 +410,7 @@ void dma_disable_channel(uint32_t dma, uint8_t channel);
 void dma_set_peripheral_address(uint32_t dma, uint8_t channel,
 				uint32_t address);
 void dma_set_memory_address(uint32_t dma, uint8_t channel, uint32_t address);
+uint16_t dma_get_number_of_data(uint32_t dma, uint8_t channel);
 void dma_set_number_of_data(uint32_t dma, uint8_t channel, uint16_t number);
 
 END_DECLS
--- a/lib/stm32/common/dma_common_f24.c
+++ b/lib/stm32/common/dma_common_f24.c
@@ -775,6 +775,20 @@ void dma_set_memory_address_1(uint32_t dma, uint8_t stream, uint32_t address)
 }
 
 /*---------------------------------------------------------------------------*/
+/** @brief DMA Stream Get the Transfer Block Size
+
+@param[in] dma unsigned int32. DMA controller base address: DMA1 or DMA2
+@param[in] stream unsigned int8. Stream number: @ref dma_st_number
+@returns unsigned int16. Number of remaining data words to transfer (65535
+maximum).
+*/
+
+uint16_t dma_get_number_of_data(uint32_t dma, uint8_t stream)
+{
+	return DMA_SNDTR(dma, stream);
+}
+
+/*---------------------------------------------------------------------------*/
 /** @brief DMA Stream Set the Transfer Block Size
 
 @note The DMA stream must be disabled before setting this count value. The count
--- a/include/libopencm3/stm32/common/dma_common_f24.h
+++ b/include/libopencm3/stm32/common/dma_common_f24.h
@@ -613,6 +613,7 @@ void dma_disable_stream(uint32_t dma, uint8_t stream);
 void dma_set_peripheral_address(uint32_t dma, uint8_t stream, uint32_t address);
 void dma_set_memory_address(uint32_t dma, uint8_t stream, uint32_t address);
 void dma_set_memory_address_1(uint32_t dma, uint8_t stream, uint32_t address);
+uint16_t dma_get_number_of_data(uint32_t dma, uint8_t stream);
 void dma_set_number_of_data(uint32_t dma, uint8_t stream, uint16_t number);
 
 END_DECLS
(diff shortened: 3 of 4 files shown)


DanielG:
No response after three months. I guess I'm starting to see why that whole unicore-mx fork happened. Given how relatively small this patch is I'm really disappointed it's just getting ignored :/

karlp:
It's not being ignored, but when the entire contribution consists of ```return DMA_CNDTR(dma,  channel)``` but you felt the need to go and add copyright claims all over the code it wasn't something I was super excited about. Is it a useful addition? sure! Is it a bug that needed immediate attention, or functionality that wasn't available without heavy work on the user side? not really.

DanielG:
I was under the impression adding copyright headers is good form, especially since the GPL relies on copyright to implement it's protections. Naturally given the triviality of the patch I'm not against removing them.

Anyways I wasn't trying to attack anyone, I was just trying to point out that ignoring PRs outright looks really bad. I'm also a project maintainer, I know that sometimes you just don't have time to look at irrelevant things like this right away. The problem is I would actually like to contribute more than just this trivial patch in the future but if something this small gets ignored why should I bother working on anything more substantial?

karlp:
I'm sorry I didn't give more immediate feedback.  I do always try.  But the copyright additions gave me a hard time.  I understand people like them, I know some projects (most apache) actively get _rid_ of them, I know that people get very heated on whether they're actually even required or not for any sort of enforcement.  My general approach is that it's noise.  The commit log tracks authorship, having a line for every edit isn't scaleable, and having a list of "owners" in a file discourages other people from editing.  Trying to find a nice way of saying, "I think the commit part looks fine, but I want you to throw out the attribution you wanted to add" was what stopped me at the time.  You also never came back to push on it, and yes, some requests do fall through the cracks.    I've put some more comments inline, and will try and get this into my queue, as it is indeed a small patch.

include/libopencm3/stm32/common/dma_common_l1f013.h
```diff
@@ -422,4 +424,3 @@ END_DECLS
 /** @endcond */
 
 /**@}*/
-
```
karlp: unrelated whitespace changes are discouraged.

lib/stm32/common/dma_common_f24.c
```diff
@@ -775,6 +776,20 @@ void dma_set_memory_address_1(uint32_t dma, uint8_t stream, uint32_t address)
 }
 
 /*---------------------------------------------------------------------------*/
+/** @brief DMA Stream Get the Transfer Block Size
+
+@param[in] dma unsigned int32. DMA controller base address: DMA1 or DMA2
+@param[in] stream unsigned int8. Stream number: @ref dma_st_number
+@returns unsigned int16. Number of remaining data words to transfer (65535
+maximum).
+*/
```
karlp: There's no need for the [in] and the "unsigned int32" stuff.  Doxygen figures that out itself from the code.  Try and use the full * on each line like you should see in most other comments.

You can even drop the "@brief" and simply put a "." (full stop) at the end of the first line if you like.

lib/stm32/common/dma_common_f24.c
```diff
@@ -791,4 +806,3 @@ void dma_set_number_of_data(uint32_t dma, uint8_t stream, uint16_t number)
 	DMA_SNDTR(dma, stream) = number;
 }
 /**@}*/
-
```
karlp: whitespace again

lib/stm32/common/dma_common_f24.c
karlp: (I see that this existing code isn't using the full * comment style so props for at least keeping it consistent) :)

karlp:
re: copyright headers, I can't find a better source right now, but as I remember it, the copyright headers were only useful for american source pre 1992, when the us finally adopted the berne convention.  http://geotoolsnews.blogspot.is/2016/07/copyright-headers-in-source-code-files.html has some, as does https://en.wikipedia.org/wiki/Copyright_notice but it's still a heated debate in many communities :)

DanielG:
I think GNU basically strongly encourages putting the license and copyright notices in the source files but from reading the GPL it does sound like it's not required, though it does say:

> The work must carry prominent notices stating that you modified it, and giving a relevant date.

So whether a git commit is sufficient here is debatable. (See also https://www.gnu.org/licenses/gpl-faq.en.html#NoticeInSourceFile)

Of course you are right that authorship information already lives in git but the problem with that is end-user doesn't always get the whole git repo but rather just a tarball or something. 

I think at the very least an `AUTHORS` or `CONTRIBUTORS` file that just collects that info in a central place so it's not noise in the source could be useful.

lib/stm32/common/dma_common_f24.c
DanielG: I just based the description on the `_set_` call as closely as I could since I don't actually know the proper doxygen syntax ;)

So do you want me to fix the style of the rest of the stuff in the file or just keep it as is?

include/libopencm3/stm32/common/dma_common_l1f013.h
DanielG: Sorry, my editor does that. I usually remove those changes again when submitting something but I though it doesn't really hurt at the end of the file since even a git blame wouldn't really show that anyways.

lib/stm32/common/dma_common_f24.c
karlp: don't worry about it, it's the same style as the rest of the file, even if the file is not matching the rest of the project.  At some point I'll go and clean them up.

include/libopencm3/stm32/common/dma_common_l1f013.h
karlp: there's a "make stylecheck" target that is a bit noisy, but can be helpful.

karlp:
thanks for this, it's now merged in https://github.com/libopencm3/libopencm3/commit/743513a4b1b15901fd6ff35f9b3c8812dc96dcaa

PR #705, opened 2016-10-04
usb: dwc_otg: enable clocks automatically.

karlp:
ST usbfs-v1, v2, EFM32 USB, lm4f usb all automatically enable the clock
in their _init routine.  Do the same for the dwc_otg drivers to be
consistent.


karlp:
merged in 422d708

PR #706, opened 2016-10-12
fix CAN on STM32F3

jschall:
These were the changes required to use CAN on STM32F3.


lib/stm32/can.c
```diff
@@ -75,8 +77,10 @@ void can_reset(uint32_t canport)
 		rcc_peripheral_reset(&RCC_APB1RSTR, RCC_APB1RSTR_CAN1RST);
 		rcc_peripheral_clear_reset(&RCC_APB1RSTR, RCC_APB1RSTR_CAN1RST);
 	} else {
+#if !defined(STM32F3)
 		rcc_peripheral_reset(&RCC_APB1RSTR, RCC_APB1RSTR_CAN2RST);
 		rcc_peripheral_clear_reset(&RCC_APB1RSTR, RCC_APB1RSTR_CAN2RST);
+#endif
```
karlp: Can you find a way to do this based on, say, the existance of the CAN2 definition instead?  it will make it more portable going forward for other devices.

jschall: Amended.

karlp:
Looks good to me.  Any can users out there?

karlp:
This has now landed.  still would be nice to hear from more can users :)

PR #708, opened 2016-11-05
cm3/fpb: Fix overly strict ARMv7 check

devanlai:
The check to see if the architecture is ARMv7 was negated incorrectly, reporting an error if either of `__ARM_ARCH_7M__` or `__ARM_ARCH_7EM__` is not defined instead of only triggering when both are undefined.

AFAICT,  `-mcpu=cortex-m3` only defines `__ARM_ARCH_7M__` and `-mcpu=cortex-m4` only defines `__ARM_ARCH_7EM__`, so this always errors out.

This commit makes the check consistent with the versions in `itm.h` and `tpiu.h`.


karlp:
Thanks for this!

PR #710, opened 2016-11-27
Add stm32 function for retrieving DFU serial number.

dhylands:


karlp:
m,erged thanks!  (I renamed it to be like the others and moved and reformatted the commentary though)

PR #711, opened 2016-11-28
STM32: RTC: Adds RTC initialize function for setting dividers, date/time regs

fishr:
Uses process described in AN3371 to configure RTCCLK dividers and load a BCD encoded date and time into the respective RTC registers

Note, the RTC domain is initially asynchronous wrt the core.  The app note simply says to wait until the synchronization flag is set (RTC_ISR_INTF) to configure it.  This is accomplished via a (blocking) while loop.  Those needing to set the RTC from a real-time context will probably want to run this via interrupt.

Also adds some one-liner getter functions to return the current BCD date/time

### DIFF
--- a/include/libopencm3/stm32/common/rtc_common_l1f024.h
+++ b/include/libopencm3/stm32/common/rtc_common_l1f024.h
@@ -327,6 +327,8 @@ specific memorymap.h header before including this header file.*/
 
 BEGIN_DECLS
 
+uint32_t rtc_get_bcd_date(void);
+uint32_t rtc_get_bcd_time(void);
 void rtc_set_prescaler(uint32_t sync, uint32_t async);
 void rtc_wait_for_synchro(void);
 void rtc_lock(void);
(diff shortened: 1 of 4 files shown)


karlp:
This all seems reasonable, but is there any particular reason this is only on f4?

fishr:
main reason: That's my target, I don't have others to test it on.  I haven't read the ref manuals for any breed other than F4.

The app note says the process should work for L1 and F0,2,3,4.  There isn't currently an rtc_common_l1f0234, but I can kill this pull request and put rtc_initialize  and the date/time getters in a new rtc_common_l1f0234.c/.h if that's desired. 
Alternatively, I could check if rtc_common_l1f024 actually applies to F3 as well, and rename it as such and change the relevant includes.

karlp:
rtc_common_l1f024.c should actually be named rtc-v2 or somthing to reduce confusion.  It's the "v2/BCD" RTC peripheral, as opposed to the "v1/seconds counter" peripheral the f1 had.  The variations within v2 are subsecond support, and number of alarms.

PR #713, opened 2016-11-29
stm32: cleanup MCO definitions

karlp:


karlp:
This will _BREAK_ some people's code, but it's in the right direction long term.  Anyone have any strong concerns on this?

ChuckM:
You probably can guess my vote here, cleaner is better and consistency across chips should be a priority. :shipit:

karlp:
Always good to get more eyes though :) thanks for the comments

PR #714, opened 2016-11-30
!stm32:usb_fx07_common - fix compiler error

alexrayne:
keil compiler throws type incompatible in assignment error, gcc should rise at least warning


karlp:
This is valid, we started some work on this in https://github.com/libopencm3/libopencm3/pull/497 but it needs to be expanded.  However, the correct fix is not simply inserting another cast.  The function arguments should just be changed to the right type.

alexrayne:
i`ve rewrite above fix
and provide additional fixes on arithmetics integer sign change and overload char shifted with long shifts

PR #715, opened 2016-11-30
stm32:CMSIS:gpio, rcc - compatibiluty vs CMSIS library 

alexrayne:
this patch allows to mix code use HAL and opencm3 includes.
                !!! HAL library MUST be included BEFORE libopencm3, to avoid conflicts
                !!! prefer use GPIOx_BASE for opencm3 routines parameters instead of GPIOx to avoid type conflicts

### DIFF
--- a/include/libopencm3/cm3/memorymap.h
+++ b/include/libopencm3/cm3/memorymap.h
@@ -21,7 +21,11 @@
 #define LIBOPENCM3_CM3_MEMORYMAP_H
 
 #ifndef __CMSIS_USE
-#if defined(__CM3_CMSIS_VERSION_MAIN) || defined(__CM4_CMSIS_VERSION_MAIN)
+#if defined(__STM32F0xx_CMSIS_DEVICE_VERSION) \
+            || defined(__STM32F1xx_CMSIS_DEVICE_VERSION)\
+            || defined(__STM32F2xx_CMSIS_DEVICE_VERSION)\
+            || defined(__STM32F3xx_CMSIS_DEVICE_VERSION)\
+            || defined(__STM32F4xx_CMSIS_DEVICE_VERSION)
 #define __CMSIS_USE	1
 #else
 #define __CMSIS_USE	0
--- a/include/libopencm3/cm3/memorymap.h
+++ b/include/libopencm3/cm3/memorymap.h
@@ -20,6 +20,14 @@
 #ifndef LIBOPENCM3_CM3_MEMORYMAP_H
 #define LIBOPENCM3_CM3_MEMORYMAP_H
 
+#ifndef __CMSIS_USE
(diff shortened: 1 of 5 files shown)


karlp:
I'm really not sure there's an awful lot of interest in trying to achieve code mixing.  It's going to make a complete mess of the libopencm3 code to do this properly at all, and I seriously question the benefits and motivation of trying to mix the two codebases in the same project.

alexrayne:
i`ve got project written on keil with RTX and TM with SPL device support code. i need to add usb device functionality. without such patch, need stricktly demarkate code used opencm3 and SPL - this leed to generate headers that allow isolate one domains and a few additional code.
so - this path provide some kind of handy for mxed projects, allow easy use SPL,HAL codebase in projects for opencm3. 
for projects without CMSIS, as you can see - no changes on opencm3 API injected.

karlp:
Why aren't you using ST's SPL usb device code? it's designed to operate with the rest of their code.

alexrayne:
because one code not designed to implement multiple interfaces, i need to provide VCP + MSC devices. imho,  their code requres many refactoring work to allow this feature.
any way, if you reject this patch, i`m forced to rebuild project code to isolate CMSIS code from opencm3

include/libopencm3/cm3/memorymap.h
```diff
@@ -21,7 +21,11 @@
 #define LIBOPENCM3_CM3_MEMORYMAP_H
 
 #ifndef __CMSIS_USE
-#if defined(__CM3_CMSIS_VERSION_MAIN) || defined(__CM4_CMSIS_VERSION_MAIN)
+#if defined(__STM32F0xx_CMSIS_DEVICE_VERSION) \
+            || defined(__STM32F1xx_CMSIS_DEVICE_VERSION)\
+            || defined(__STM32F2xx_CMSIS_DEVICE_VERSION)\
+            || defined(__STM32F3xx_CMSIS_DEVICE_VERSION)\
+            || defined(__STM32F4xx_CMSIS_DEVICE_VERSION)
```
karlp: This is even worse.  What about non ST parts? What about STM32L1 and L4?!  This is a bad path.

alexrayne: hm. i thought, you wanted reject this idea anyway. 
it`s "even work" on that noted platforms... as for other targets  - somebody who need this compilance, should append their identificators. 
if you realy, interested in that patch, i append other identificators, that i can found in cmsis for stm32
this patch is not complete solution, cause i have compile it vs only rcc, gpio, and usb parths. other devices not tested vs macro definitions conflicts.

karlp: I'm not interested in it, but I'll still try and help improve things you submit :)

karlp:
No-one else has every requested this, so don't see any real reason to keep this PR around open.  Closing it.

PR #716, opened 2016-11-30
stm32(x07) .usbd - add otg_FS device driver

alexrayne:
this patch introduce stm32f207_usbd_hs_driver and stm32f207_usbd_fs_driver, allow to use both OTG cores.
also it provideOTG_xxx_io structures like CMSIS ones, and provides for FS and HS devices io structures OTG_xS_GLOBAL/DEVICE/HOST/IEPn/OEPn/HCn

### DIFF
--- a/include/libopencm3/stm32/otg_common.h
+++ b/include/libopencm3/stm32/otg_common.h
@@ -490,4 +490,137 @@
 
 
 
+
+//*******************************************************************
+//          this OTG IO structures imported from CMSIS
+/**
+ * @brief __USB_OTG_Core_register
+ */
+typedef struct {
+    uint32_t GOTGCTL;       /*!<  USB_OTG Control and Status Register    000h*/
+    uint32_t GOTGINT;       /*!<  USB_OTG Interrupt Register             004h*/
+    uint32_t GAHBCFG;       /*!<  Core AHB Configuration Register    008h*/
+    uint32_t GUSBCFG;       /*!<  Core USB Configuration Register    00Ch*/
+    uint32_t GRSTCTL;       /*!<  Core Reset Register                010h*/
+    uint32_t GINTSTS;       /*!<  Core Interrupt Register            014h*/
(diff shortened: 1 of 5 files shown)


karlp:
I'm afraid I don't see any motivation for this patch.  Aside from the completely different code style and the redundant defininitions, what are you actually trying to add here? Which target doesn't have what module enabled?

alexrayne:
current implementation of usb/usb_f207.c provides only stm32f207_usb_driver - that works on HS core.
i use stm32f4_discovery that have micro-usb port routed on OTG_FS core - it is another core, and attached on different pins than HS.  Such FS core have all stm32f4, stm32f2 targets.
Therefore to work with this FS i need to provide yet another driver - this one declared as usbd.h:stm32f207_usbd_fs_driver. and old stm32f207_usb_driver - now wraped to stm32f207_usbd_hs_driver

To work with similar OTG device usb_fx07_common havily used REBASE macro ( even hard for compiler optimisator, for example debug lesser optimisation may not optimise such deep dereferencing, and i saw compilers that eve can`d do it on strong optimisations -O2  ). but this macro is some local hardcode - (usbd_dev->driver->base_address) , wich must be always localy reimplements.  
Should i do this?
(imho SPL strategy- defining io structures more perspective - it is more native and glance.)

karlp:
the stm32f107 is the fs core.  This is already available on f4 discovery.  See https://github.com/libopencm3/libopencm3-examples/blob/master/examples/stm32/f4/stm32f4-discovery/usb_cdcacm/cdcacm.c#L238

For the other F4 discovery, that has the HS core connected: see https://github.com/libopencm3/libopencm3-examples/blob/master/examples/stm32/f4/stm32f429i-discovery/usb_cdcacm/cdcacm.c#L239

Here you can see that both modules are included for the F4 already: https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/f4/Makefile#L53-L54

PR #719, opened 2016-12-07
usb: Functions to manipulate odd/even flag for iso endpoints

WGH-:
Disclaimer: I'm by no means expert in STM32 programming. This pull request is the result of amateur experiments with my microcontroller. Some conclusions might be wrong.

On STM32, USB isochronous IN endpoints have rather weird feature: a built-in "filter" that, depending on its settings, accepts packets only on odd/even frames.

This basically means that when streaming data, unless you flip the "filter" setting accordingly each frame, you'll get 50% packet loss. You can find a few mentions of people battling with this feature over the internet, e.g. here: http://www.chibios.com/forum/viewtopic.php?f=16&t=926 

Unfortunately, libopencm3 doesn't expose corresponding bits of the control register that manipulate the "filter".

This pull requests adds two new functions to generic driver interface: `usbd_ep_set_eonum` (changes filter setting for specified endpoint) and `usbd_get_fnsof` (retrieves current frame number from status registers), and also their implementations for `stm32f107_usb_driver`.

I have tested these new functions on my NUCLEO-F411RE board, by calling `   usbd_ep_set_eonum(usbd_dev, 0x01, usbd_get_fnsof(usbd_dev));` in SOF callback, and eliminated packet loss altogether.


ChuckM:
This isn't a "filter" this is part of the USB spec, it is up to the host and device to correctly synch their data toggle bits. The role of the data toggle is to give an indication in bulk or iso data transfers to indicate that the client and host are out of sync and to take corrective action. Normally the data toggles 1/0/1/0/1/0/1/0 and the host will ignore packets that have the toggle bit set differently than it is expecting.

WGH-:
Could you quote the spec about this?

I know that USB uses data toggle bits (DATA0/DATA1) everything (?) except isochronous transfers (and I had no problems with control and bulk on STM32 at all), and also for high-speed isochronous transfers (my STM32 supports only full-speed, so it doesn't matter).

> The data transmitter and receiver must have their sequence bits synchronized at the start of a transaction. The synchronization mechanism used varies with the transaction type. Data toggle synchronization is not supported for isochronous transfers.

> High-speed, high-bandwidth isochronous and interrupt endpoints support a similar but different data
synchronization technique called data PID sequencing. That technique is used instead of data toggle
synchronization. Section 5.9.1 defines data PID sequencing.

page-tek:
#604 also looks to be solving the same problem

PR #721, opened 2016-12-09
*usb:usbd_init:aligns - notice  and aserted about parameters align

alexrayne:
*Desribe :since xxx_descriptor structures are packed compiler may place it unaligned.
        at least keil ARM compiler do thus. therefore this check helps find it early
        at debuging stage.

### DIFF
--- a/include/libopencm3/usb/usbd.h
+++ b/include/libopencm3/usb/usbd.h
@@ -84,10 +84,10 @@ extern const usbd_driver efm32lg_usb_driver;
  * @return the usb device initialized for use. (currently cannot fail).
  */
 extern usbd_device * usbd_init(const usbd_driver *driver,
-			       const struct usb_device_descriptor *dev,
-			       const struct usb_config_descriptor *conf,
+			       const struct usb_device_descriptor *dev,     //MUST alined(4)
+			       const struct usb_config_descriptor *conf,    //MUST alined(4)
 			       const char **strings, int num_strings,
-			       uint8_t *control_buffer,
+			       uint8_t *control_buffer,                     //MUST alined(4)
 			       uint16_t control_buffer_size);
 
 /** Registers a reset callback */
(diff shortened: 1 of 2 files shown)


karlp:
Only _required_ on CM0, _can_ provide performance gains on cm3+, but not required.  Correct place for this is in user documentation, never by an assert.

alexrayne:
i`ve got hardware failts on stm32f4, because ARM compiler plased descriptor constants in .text segment misaligned. so, imho, it is not only CM0 problem

karlp:
If you've got faults from unaligned on f4, please fix the tests in https://github.com/libopencm3/libopencm3/blob/master/tests/gadget-zero/test_gadget0.py#L321

alexrayne:
i can rewrite it, and even can check that it compiles. but can`t debug this test - i`m winXP user.
anyway, this misalign bug - is generic for ARM and MIPS cores. it`s funny that bug was classifyed only for M0. imho it is caused by features of gcc - tryes to align even packed constats.

karlp:
there's nothing linux specific about the testing.  You can run the python tests perfectly well on windows.

alexrayne:
imho, this is not a test, for my fix. hardfaults caused by unaligned device descriptor access occurs at usb setup stage - even before device initiated.
i guess, that failures may be fixed by pull #723 "*stm32:unaligned buffer endpints read/write". on request USB_DT_DEVICE , usb_standard_get_descriptor try to post device descriptor direct from descriptor constant. so, if one misaligned, we got hardfault by misaligned buffer read to fifo. 
other requests are builds in ctrl_buf with memcopy, so looks like them works fine.
imho even better fix  - memcopy device descriptor to ctrl_buf in usb_standard_get_descriptor, just like other descriptors builds.

karlp:
no update on fixing the tests, and noupdates on what they were going to try later.  will be, or already has been, handled by other users with alignment concerns. not keeping this around any more

PR #722, opened 2016-12-09
!usbd_register_set_config_callback - no handler copy 

alexrayne:
just a small fix protects register one callback multiple times, that will loose callbacl slots and may prevent executing configuration multiple times at activation

### DIFF
--- a/lib/usb/usb_standard.c
+++ b/lib/usb/usb_standard.c
@@ -46,7 +46,10 @@ int usbd_register_set_config_callback(usbd_device *usbd_dev,
 
 	for (i = 0; i < MAX_USER_SET_CONFIG_CALLBACK; i++) {
 		if (usbd_dev->user_callback_set_config[i]) {
-			continue;
+			if (usbd_dev->user_callback_set_config[i] != callback)
+				continue;
+			else
+				return 0;
 		}
 
 		usbd_dev->user_callback_set_config[i] = callback;


karlp:
So to clarify, this is intended to allow registering the same callback repeatedly?

alexrayne:
yep, it is intended for implementing composit devices with multiple functions - if i create multile MSC interfaces, they can try to register ones callback per every interface. Without this patch, i must perform registration of functional class, outside of class  functionality-implementer. 
Maybe this aproach more careful, but this patch gives more fredom for library users.

lib/usb/usb_standard.c
```diff
@@ -46,7 +46,10 @@ int usbd_register_set_config_callback(usbd_device *usbd_dev,
 
 	for (i = 0; i < MAX_USER_SET_CONFIG_CALLBACK; i++) {
 		if (usbd_dev->user_callback_set_config[i]) {
-			continue;
+			if (usbd_dev->user_callback_set_config[i] != callback)
+				continue;
+			else
+				return 0;
```
ChuckM: I would write this differently
```
if ((usbd_dev->user_callback_set_config[i] != NULL) &&
   (usbd_dev->user_callback_set_config[i] == callback)) {
    return 0;
}
```
The multiple continues in there are a bit weird.

karlp: it's so that it continues properly with the right return codes though.

ChuckM: You are right @karlp here is (for me) the more readable and understandable code:
```c
int usbd_register_set_config_callback(usbd_device *usbd_dev,
				       usbd_set_config_callback callback)
{
	int i;
	for (i = 0; i < MAX_USER_SET_CONFIG_CALLBACK; i++) {
		if ((usbd_dev->user_callback_set_config[i] == NULL) ||
                    (usbd_dev->user_callback_set_config[i] == callback)) {
			usbd_dev->user_callback_set_config[i] = callback;
                        return 0;
               }
        }
       return -1;
}
```

karlp: hehe :)  I'd worked up this version myself locally: 
```
	for (i = 0; i < MAX_USER_SET_CONFIG_CALLBACK; i++) {
		if (usbd_dev->user_callback_set_config[i]) {
			if (usbd_dev->user_callback_set_config[i] == callback) {
				return 0;
			}
			continue;
		}

		usbd_dev->user_callback_set_config[i] = callback;
		return 0;
	}
```

karlp:
an implementation of this has now merged, see  https://github.com/libopencm3/libopencm3/commit/f4bbe7c5bb39086a748810e83220e4e5a336bf50

PR #723, opened 2016-12-09
*stm32:unaligned buffer endpints read/write 

alexrayne:
provide handling of unaligned buffer io for stm32fx07_ep_write/write_packet.

        old behaviour demands that buf MUST be aligned at uint32_t boundary. that leads to difficulties with
        unaligned data. now for unaligned bufer used a bit slower FIFO load/store, that should give a negligible
        penalty vs aligned data on system with cache enabled.

### DIFF
--- a/lib/usb/usb_fx07_common.c
+++ b/lib/usb/usb_fx07_common.c
@@ -197,6 +197,8 @@ uint16_t stm32fx07_ep_write_packet(usbd_device *usbd_dev, uint8_t addr,
 			      const void *buf, uint16_t len)
 {
 	const uint32_t *buf32 = buf;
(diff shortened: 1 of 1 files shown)


karlp:
You'll need to provide solid test results verifying gains here to even consider adding duplicate paths like this.

alexrayne:
my english not enough to understan you, can you describe more?
i`ve tested this patch on my implementation of CDC device implements VCP:
 my vcp have in, out fifo buffers where endpoints read\writes their data. since comm data are byte-aligned, read\write positions are byte-aligned too. this patch surely helps, without it i must implement double-bufering - transfer from endpoint to aligned buffer, and unaligned byte copy to fifo. double-buffering is realy cpu impact on memory, code and speed.

PR #725, opened 2016-12-10
!usb: fix loosing XFRC flag after slow IN endpoint callback

alexrayne:
*Descibe: if IN callback is slow enough, that usb core completes send transfer before callback
        return, OTG_DIEPINTX_XFRC will dorp by poller, therefore completion of sended frame
        cant be noifyed. that s why need to clear OTG_DIEPINTX_XFRC before callback


karlp:
Please setup tests for this, it should be quite straightforward in the gadget0 framework.

karlp:
Further, since this was filed, there have been some important fixes to dwc_otg, please confirm that there's still a bug at all.

karlp:
No reply, no tests, no updates. 

Please feel free to re-open this if you have some tests or further results, but I'm closing this for now.

PR #726, opened 2016-12-13
Fixed SCB_SHCSR and SCB_DFSR on M0/M0+ and masked FPU registers from M3

HenningJW:
Making SCB_SHCSR and SCB_DFSR available on M0/M0+ including the bitmasks according to the [ARMv6-M Architecture Reference Manual](https://imagecraft.com/pub/ARM/DDI0419C.pdf). Also masked the not available FPU registers on M3. There are optional on ARMv7E-M and I don't know if there is a define for an included FPU to mask the registers further. I also don't think this is necessary.

### DIFF
--- a/include/libopencm3/cm3/scb.h
+++ b/include/libopencm3/cm3/scb.h
@@ -51,19 +51,19 @@
 #define SCB_SHPR2				MMIO32(SCB_BASE + 0x1C)
 #define SCB_SHPR3				MMIO32(SCB_BASE + 0x20)
 
-/* Those defined only on ARMv7 and above */
-#if defined(__ARM_ARCH_7M__) || defined(__ARM_ARCH_7EM__)
 /* SHCSR: System Handler Control and State Register */
 #define SCB_SHCSR				MMIO32(SCB_BASE + 0x24)
 
+/* DFSR: Debug Fault Status Register */
+#define SCB_DFSR				MMIO32(SCB_BASE + 0x30)
+
+/* Those defined only on ARMv7 and above */
+#if defined(__ARM_ARCH_7M__) || defined(__ARM_ARCH_7EM__)
 /* CFSR: Configurable Fault Status Registers */
 #define SCB_CFSR				MMIO32(SCB_BASE + 0x28)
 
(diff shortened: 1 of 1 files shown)


karlp:
For future reference, these would be easier to consider as two commits, as they're separate issues.

karlp:
Confirm DFSR and SHCSR.  Need to check the provided defines for FPU stuff

karlp:
Not entirely sure about the FPU stuff.  according to my tests with the guidelines from https://launchpadlibrarian.net/287100883/readme.txt

Compiling for m3 defines __ARM_ARCH_7M__ and for m4 (NO FPU!) defines __ARM_ARCH_7EM__ and even __ARM_FEATURE_DSP.

Compiling for m4 with fpu provides some extra things, but nothing easy or obvious.

I'm inclined to simple leave access to the FPU defines for all 7EM.?

HenningJW:
In my commit I only masked the FPU from ARM_ARCH_7M. Even though their [reference manual for 7M](https://web.eecs.umich.edu/~prabal/teaching/resources/eecs373/ARMv7-M_ARM.pdf) on page 737 states "If the FP extension is not implemented, the FPCCR register location is reserved.", there is no M3 with FP extension supported by libopencm3 from my knowledge and I think there is even no FPU enabled M3 available anywhere. M4 (without FPU) I did not touch, partly because I don't know if it would be a smart choice to check for fpu symbols ("fpu=fpv4xxx" etc. or maybe negated "fpu=none") and partly because I don't think it is really that important.

If you prefer a split or different formatting I can update this PR (or close and make new ones).

karlp:
merged the SCHSR and DFSR changes.  Ran out of time pulling apart the rest.  Not super sure it's entirely necessary to try and protect M3 users from accessing the FPU regs, and the restriction you've got would still allow M4(nofpu) users access, so not entirely sure of the benefit.  Atmel SAM4n (at least) is m4(nofpu) so that certainly exists.  Anyway, resubmit that portion if you like, it's certainly technically valid :)

karlp:
closed as the main portions have merged in f974861

PR #727, opened 2016-12-16
stm32l1: flash: avoid duplicate calls to unlock_pecr

karlp:
Attempts to helpfully unlock PECR when required in unlock_progmem and
unlock_option_bytes actually cause a bus error due to repeated unlocks,
as per ref manual and tedious experience.  The better tested eeprom helper
routines unlock/lock in chunks, but that's not applicable for flash
writing.

Fixes: cf5fb002f6016242fb23b81fcbe98ee022bb84e9

### DIFF
--- a/lib/stm32/l1/flash.c
+++ b/lib/stm32/l1/flash.c
@@ -121,6 +121,12 @@ void flash_set_ws(uint32_t ws)
 	FLASH_ACR = reg32;
 }
 
+/**
+ * Unlock primary access to the flash control/erase block
+ * You must call this before using any of the low level routines
+ * yourself.
+ * @sa flash_unlock
+ */
 void flash_unlock_pecr(void)
 {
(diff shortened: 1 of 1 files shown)


karlp:
As an alternative, the lower level unlock routines _could_ check whether the bits are locked first, but that seems like more code for little gain.  Possibly the entire apis are poor, and should be unified across all the families, but it's a little used area, and at least making it not busfault seems like the priorirty

ChuckM:
The changes look good so :+1: to those,  to your API question a unified 'write option bytes' or 'write flash' API that would unlock, write, and re-lock would minimize errors here.

karlp:
indeed.  There's an example bootloader for f1, expanding it to be a "DFU bootloader for all platforms" would be a great exercise of both USB and the flash APIs.  Another one to chalk up as a "great project" if someone's bored. Long lists never seem to get much shorter!

karlp:
merged in 950e064

PR #729, opened 2016-12-19
stm32/ethernet: fix broken include

fxkr:


karlp:
Actually, I think the correct path here is to simply delete that file.  You should just 

```
#include <libopencm3/ethernet/mac.h>
```

I've merged that in 1963f32 thanks for reporting this!

PR #733, opened 2017-01-02
stm32/f0: implement RTC helpers

geomatsi:
Hi all,
This PR includes two simple patches implementing various RTC alarm and clock helpers.
Tested on a custom board based on stm32f030f4p6 chip.

Regards,
Sergey


lib/stm32/common/rtc_common_l1f024.c
```diff
@@ -120,4 +120,14 @@ void rtc_clear_wakeup_flag(void)
 	RTC_ISR &= ~RTC_ISR_WUTF;
 }
 
+/*---------------------------------------------------------------------------*/
+/** @brief Clears the Alarm A flag
+
+*/
+void rtc_clear_alarm_flag(void)
+{
+	/* clear flag: not write protected */
```
karlp: that sort of comment is probably more useful above perhaps? where it makes it into the doxygen?

lib/stm32/f0/rcc.c
```diff
@@ -439,6 +439,47 @@ void rcc_set_usbclk_source(enum rcc_osc clk)
...
+	switch (clk) {
+	case RCC_HSE:
+		RCC_BDCR = (RCC_BDCR & ~RCC_BDCR_RTCSEL) | RCC_BDCR_RTCSEL_HSE;
+	case RCC_LSE:
+		RCC_BDCR = (RCC_BDCR & ~RCC_BDCR_RTCSEL) | RCC_BDCR_RTCSEL_LSE;
+	case RCC_LSI:
+		RCC_BDCR = (RCC_BDCR & ~RCC_BDCR_RTCSEL) | RCC_BDCR_RTCSEL_LSI;
+	default:
+		/* incorrect clock source: turn off clock sources to RTC */
+		RCC_BDCR = (RCC_BDCR & ~RCC_BDCR_RTCSEL) | RCC_BDCR_RTCSEL_NOCLK;
```
karlp: in most other cases, the "deffault" does nothing, not changes the behaviour, I think that's the right approach to keep.

geomatsi: Ok, that sounds reasonable, will be fixed.

lib/stm32/common/rtc_common_l1f024.c
geomatsi: Ok, I will add this note as `@details` into doxygen header.

geomatsi:
v1 -> v2: changes according to review comments

- `rtc_clear_alarm_flag`: move comment on register protection to `@details` in doxygen header
- `rcc_set_rtc_clock_source`: fix behavior for incorrect input clock source - in this case do nothing keeping current clock settings
- `rcc_set_rtc_clock_source`: add missing `break` statements

lib/stm32/common/rtc_common_l1f024.c
```diff
@@ -120,4 +120,14 @@ void rtc_clear_wakeup_flag(void)
...
 }
 
+/*---------------------------------------------------------------------------*/
+/** @brief Clears the Alarm A flag
+
+@details RTC_ISR[13:8] bits are not write protected: no need to unlock
+*/
+void rtc_clear_alarm_flag(void)
+{
+	RTC_ISR &= ~RTC_ISR_ALRAF;
```
karlp: You've added this to the common code.  Either, 

1. The common code should clear bits 8 and 9 (Alarm A and B)  (On f0, it's "must be kept at reset value (0)" so that's "safe") (and docs updated to indicate a bit more clearly what it's doing)
2. Or, the function prototype stays in common code, and the implementation is moved to just f0, where it only clears Alarm A, and implementations added to L1, F2, F4 (and later eventually probably  l0, f3, f7.....)

I'm inclined to go with option 1,  but as it stands, this would be misleading for the  non-f0 parts.  Opinions? Thoughts?

 I guess there's also 

3. Separate functions for clear alarm A and clear alarm B
4. Use a function more like the F1 API, which uses a single "clear flag" function and takes the flag to clear as an arg.

(I would still lean towards option 1 personally)

karlp:
If you can separate the alarm stuff from the clock source stuff, I can pull in the clock source stuff, it looks good to go.

lib/stm32/common/rtc_common_l1f024.c
geomatsi: Well, it is probably better to keep it in the common code in the case if the same functionality is available on all the platforms in question. Anyway thanks for pointing at the alarm B thing, I will take a look through all those specs and get back with comments and updates. On the first glance, the answer is going to depend on whether alarm B is available on the same platforms as A.

karlp: As far as I can tell from the F0/l1/f4 and f2 manuals, (and my memory of the "intro to the RTC peripheral" app note that explains some of the platform differences) everythign has alarmA, and some have alarmB, and as far as I can tell,  all the bits for alarm B are always "reserved" on devices that don't have the  alarm.

geomatsi:
Sure, will do it later this evening. I will force-update this PR keeping only HSE clock stuff. RTC will go in a separate PR after I figure out alarm B details.

karlp:
ok, sounds reasonable. thanks for sticking  with this. I know the lag is high :|

geomatsi:
np, I don't mind lags in the case if polished and not-so-controversial stuff can finally make its way into master.

geomatsi:
Hi all,
v1 -> v2: keep only commit implementing RTC clock helpers in this PR. Alarm change need to be revisited in terms of alarm A and alarm B support. It will be submitted later in a separate PR.

Regards,
Sergey

karlp:
this has now landed in 43736cf thanks!

Issue #734, opened 2017-01-08
STM32F0: usart_get_interrupt_source() doesn't work properly


zyp:
The current code assumes that ORE is gated by CTSIE. According to the reference manual, ORE flag is gated by RXNEIE or EIE and has nothing to do with CTSIE.

The function will also ignore any other enabled interrupts except IDLE, RXNE, TC and TXE.

I'm also not sure checking the interrupt bits against the enable bits are useful at all. I assume the thought behind it is to ignore interrupt bits that can't be the source of the currently active interrupt, but why would users call this function for disabled interrupts in the first place?

I propose replacing the entire function body with ` return (USART_ISR(usart) & flag) != 0;`.

karlp:
I concur.  Only, with other wording.  Simply drop the usart_get_interrupt_source() routine.  Use the usart_get_flag()

I'm not sure of the usefulness of such explicit internal checking of both enabled and status bits either.

PR #735, opened 2017-01-08
stm32: usart: drop usart_get_interrupt_source()

karlp:
It was never complete, even for F1 family code, and went on to be even
less complete for f0 and f3.  The usefulness of a library function to
check for both the irq being enabled _and_ the status flag is highly
questionable, and caused known user confusion.

The existing, much simpler, and fully functional usart_get_flag() is
a good replacement in almost all sane use cases.

Fixes https://github.com/libopencm3/libopencm3/issues/734


karlp:
this has now landed

PR #743, opened 2017-02-05
st_usbfs_v1: Allows user to assign endpoint address by themselves

rgwan:
I 'm not sure that every usb ip core allows user to assign endpoint by manual, and if I modify the usbd part, it will causes a lot of work: To modify old codes.

So, I realized it in a ugly way just for st_usb driver by a new core function which allows user to assign endpoint address manually.

This function is "st_usbfs_ep_setup_core" in st_usb_core.c. Hopefully somebody will figure out a MUCH better idea for usbd modification.


karlp:
Leaving aside the issue of having a stable API across cores, what's even the use case for this?

PR #744, opened 2017-02-08
added functions for setting rtc-relevant registers in the rcc domain,…

fishr:
… mainly clock sources and backup domain resetting

Looks to be relevant to L1, F0,2,3,4 based on [http://www.st.com/resource/en/application_note/dm00025071.pdf](url) Currently the only files modified are for F4, but that could be expanded.


karlp:
Needs to go to https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/common/rtc_common_l1f024.c and have documentation added, but otherwise looks fine.  commit subject line should be reworded to be consistent with history too, but that's minor.

fishr:
I did think it was strange as implemented, but the registers involved are RCC registers even though they only affect RTC function.  I just want to double check that putting these functions in the RTC file is the desired structure.  I'll have to add rcc.h as an include to rtc_common.  Also, should I keep the function names prepended with "rcc_" ?

PR #750, opened 2017-03-04
stm32f4: pwr: Fix broken PWR_SCALE2 mode

0xdec:
It looks as though someone was trying to clear the VOS bit by ANDing it, but forgot to NOT it first.

### DIFF
--- a/lib/stm32/f4/pwr.c
+++ b/lib/stm32/f4/pwr.c
@@ -38,9 +38,14 @@
 
 void pwr_set_vos_scale(enum pwr_vos_scale scale)
 {
+	/*
+	 * Scale 1 mode: max 168MHz (high performance, default at reset).
+	 * Scale 2 mode: max 144MHz (power saving).
+	 * Scale 3 mode: max 120MHz (STM32F42xxx/3xxx only).
+	 */
 	if (scale == PWR_SCALE1) {
 		PWR_CR |= PWR_CR_VOS;
 	} else if (scale == PWR_SCALE2) {
-		PWR_CR &= PWR_CR_VOS;
+		PWR_CR &= ~PWR_CR_VOS;
 	}
 }


0xdec:
After looking a bit further, I'm not completely confident that PWR_SCALE1 is low power and PWR_SCALE2 is high performance. That's what it seems it should be based on the data sheet and manual, but rcc_clock_setup_hse_3v3 sets VOS to PWR_SCALE2 if clock->power_save is 1, and PWR_SCALE1 otherwise. This behavior seems to be opposite?

lib/stm32/f4/pwr.c
```diff
@@ -38,9 +38,15 @@
 
 void pwr_set_vos_scale(enum pwr_vos_scale scale)
 {
+	/*
+	 * This function only works fully for F405xx/07xx and F415xx/17xx.
+	 * Scale 1 mode is power saving and is the default at reset.
+	 * Scale 2 mode is high performance.
```
0xdec: I can change these to read simply "Scale 1 mode is the default at reset." if needed, as I am unclear as to which is which.

karlp:
Some of this probably comes from the earlier parts only having a single bit, and new parts adding extra bits and making a range of power scales.  I'll look at it more closely later.

lib/stm32/f4/pwr.c
```diff
@@ -39,10 +39,9 @@
 void pwr_set_vos_scale(enum pwr_vos_scale scale)
 {
 	/*
-	 * This function only works fully for F405xx/07xx and F415xx/17xx.
-	 * Scale 1 mode is power saving and is the default at reset.
-	 * Scale 2 mode is high performance.
-	 * TODO: add support for Scale 3 mode
+	 * Scale 1 mode: max 168MHz (high performance, default at reset).
+	 * Scale 2 mode: max 144MHz (power saving).
+	 * Scale 3 mode: max 120MHz (STM32F42xxx/3xxx only).
```
karlp: You seem to dropped the "TODO: add support for scale 3 mode" ?

karlp: Also, seeing as this comment was only introduced by your first commit, simply squish them together into a single commit.  you can push -f to the same branch and github will autoupdate the PR

0xdec: I tried to squash the two commits earlier but it seems not to have worked. I think I fixed it now. As for Mode 3 support, I know how to go about adding it, but I don't have F42xxx hardware to test it on.

0xdec:
The main issue is just the missing tilde that would cause it to actually clear the bit. Comments should be squashed now.

karlp:
superseded by cc8c8a2

PR #752, opened 2017-03-06
stm32f4: flash: Refactor FLASH_ACR code

0xdec:
Instead of enabling data and instruction caches through flash_set_ws(), use their dedicated functions. This seemed hacky to be using flash_set_ws() to set bits other than the wait state bits. I changed it to instead use the flash_dcache_enable() and flash_icache_enable() functions, which should be more robust.

### DIFF
--- a/lib/stm32/f4/rcc.c
+++ b/lib/stm32/f4/rcc.c
@@ -60,8 +60,7 @@ const struct rcc_clock_scale rcc_hse_8mhz_3v3[RCC_CLOCK_3V3_END] = {
 		.ppre1 = RCC_CFGR_PPRE_DIV_4,
 		.ppre2 = RCC_CFGR_PPRE_DIV_2,
 		.power_save = 1,
-		.flash_config = FLASH_ACR_ICE | FLASH_ACR_DCE |
-				FLASH_ACR_LATENCY_3WS,
+		.flash_config = FLASH_ACR_DCEN | FLASH_ACR_ICEN | FLASH_ACR_LATENCY_3WS,
 		.ahb_frequency  = 48000000,
 		.apb1_frequency = 12000000,
 		.apb2_frequency = 24000000,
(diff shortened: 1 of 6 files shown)


lib/stm32/f4/rcc.c
```diff
@@ -709,14 +693,14 @@ uint32_t rcc_system_clock_source(void)
 
 void rcc_clock_setup_hse_3v3(const struct rcc_clock_scale *clock)
 {
-	/* Enable internal high-speed oscillator. */
+	/* Enable internal high-speed oscillator (HSI). */
```
karlp: out of scope.  please avoid making changes like this.

lib/stm32/f4/rcc.c
```diff
@@ -743,6 +727,8 @@ void rcc_clock_setup_hse_3v3(const struct rcc_clock_scale *clock)
 	rcc_wait_for_osc_ready(RCC_PLL);
 
 	/* Configure flash settings. */
+	flash_dcache_enable();
+	flash_icache_enable();
```
karlp: hacky as it may have been, you've removed the option of running with these disabled.

lib/stm32/f4/rcc.c
0xdec: Okay, I was just trying to improve the comments. Should I edit the commit to remove those, and submit a separate PR? Or just nix them entirely?

lib/stm32/f4/rcc.c
0xdec: Good point. Would it be acceptable to have a check for those bits there and then call the associated flash_Xcache_enable() function if needed? That way the config options could still be passed in the struct but would not rely on flash_set_ws().

lib/stm32/f4/rcc.c
karlp: I'd edit the commit to remove those, and honestly, right now, don't submit a new PR for those just now :|

lib/stm32/f4/rcc.c
karlp: up to you.  You could also add bools to the struct.  You could also simply document that it's always turned on now, and see if anyone actually complains :)

lib/stm32/f4/rcc.c
0xdec: Sounds good, I'm on it.

lib/stm32/f4/rcc.c
0xdec: Haha, thanks!

karlp:
this has now landed in 6798cee..90d8bd6 thanks!

PR #756, opened 2017-03-11
stm32: timer: Add missing TIMx_CCER bit definitions

0xdec:
These are needed if you want to use channel 4 with a different polarity


karlp:
What part are using?  Bit 15 exists in the f4, but I don't see bit 14 anywhere?  Just curious, extra bit definitions don't hurt anyone

0xdec:
You're right, I don't see bit 14 either, but I just assumed a different part might have it. I can remove it if you think that's better.

0xdec:
Removed bit 14 (TIM_CCER_CC4NE)

karlp:
merged in 26ca4cc thanks

PR #737, opened 2017-01-12
Karlp staging 2017.1

karlp:
Starting a new staging branch. (I rebase this and do local testing for a while, and eventually, without comments, this whole tree lands)

Supersedes #418 
Contains #735 
Contains #706 
Contains #673
Contains #651
Contains #552 
Contains #733 
Contains #728
Contains #470 (expanded)
Contains #752
Contains #399

Questions on the names of the two rng_get_random functions.  Should the "blocking, easy, simple" version be "rng_get_random" and the non-blocking have the suffix?  what's the nicest in user code?

PR #757, opened 2017-03-12
stm32 PWR renaming

matthewlai:
Renamed pwr_common_all to pwr_common_v1, and pwr_common_l01 to pwr_common_v2.

F7's pwr definitions are very different and not a superset of any existing series, so I'll make them into pwr_common_v3 (in anticipation that H7 will be similar if not the same) in a future PR if there are no objections?

Thanks


karlp:
looks good to me, v3 sounds good.

.gitignore
```diff
@@ -20,6 +20,7 @@ latex/
 *.pdf
 *.tag
 .DS_Store
+locm3.sublime-workspace
```
karlp: perhaps just *.sublime-workspace?

karlp:
Pulled the pwr bit into https://github.com/libopencm3/libopencm3/pull/737

.gitignore
matthewlai: That works too, though I believe the file name here is always the same as the project name, and the project file is in the repo.

karlp: oh, it is too!  I hadn't noticed someone had added that.  If it's actually part of the project though, it shouldn't be gitignored.  It should either be dropped, and added to gitignore, or you need to coordinate your changes with other people.

matthewlai: According to the documentation here (https://www.sublimetext.com/docs/3/projects.html), .sublime-project should be in version control (since it contains things like paths and tab size), and .sublime-workspace shouldn't (since it contains things like which files are currently open in the editor).

karlp: ahh, gotcha.  I missed the -workspace vs -project distinction.  my bad.  I'll pull this bit in too.

matthewlai:
Cool thanks!
I accidentally used my master branch for this. Not very familiar with GitHub. Should I close the PR so new commits won't show up here?

karlp:
Just leave it.  I've marked it for my staging, but that hasn't landed yet.  I'll close this when I land all of that. And don't worry about your own master getting new commits, (for this PR)  Just use a branch for the PRs going forward.

karlp:
This has now landed

PR #758, opened 2017-03-14
stm32: rcc: added delays after periph clock enable

matthewlai:


karlp:
The big TODO doesn't say anything?  There was discussion on this elsewhere, on IRC I guess, that this was ~never an issue on non f7, as they were never fast enough, so it was in the "correct, but really?"   And the F7 code isn't ready either.  (And people would be rather surprised if periph enable resulted in division calculations)

@OpenUAS have you actually hit this?

PR #764, opened 2017-03-28
Rename some f4 register definitions

jhalmen:
I recently started using an f401 and was very confused when i couldn't find them at first, here i fixed the ones i found


karlp:
Is there any particular reason you're closing these? they seem generally reasonable, if unfortunately not backward compatible.

jhalmen:
it was actually a mistake, was trying to clean up my repository / branches, and move stuff around. don't have much experience with git yet, so this happened.

PR #766, opened 2017-04-01
stm32f7: pwr: added basic support for pwr (VOS and overdrive)

matthewlai:
This hasn't been tested yet. I'll test it over the weekend.

### DIFF
--- a/include/libopencm3/stm32/f7/pwr.h
+++ b/include/libopencm3/stm32/f7/pwr.h
@@ -1,0 +1,282 @@
+/** @defgroup pwr_defines PWR Defines
+
+@brief <b>Defined Constants and Types for the STM32F7xx Power Control</b>
+
+@ingroup STM32F7xx_defines
+
+@version 1.0.0
+
+@author @htmlonly &copy; @endhtmlonly 2017 Matthew Lai <m@matthewlai.ca>
+
+@date 12 March 2017
+
+LGPL License Terms @ref lgpl_license
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2017 Matthew Lai <m@matthewlai.ca>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef LIBOPENCM3_PWR_H
+#define LIBOPENCM3_PWR_H
+
+/**@{*/
+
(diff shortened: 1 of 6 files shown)


matthewlai:
This has been tested (chip boots up, and LED blinks at the right speed for 216 MHz).

include/libopencm3/stm32/f7/pwr.h
```diff
@@ -0,0 +1,284 @@
...
+@author @htmlonly &copy; @endhtmlonly 2017 Matthew Lai <m@matthewlai.ca>
+
+@date 12 March 2017
+
+LGPL License Terms @ref lgpl_license
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2011 Stephen Caudle <scaudle@doceme.com>
```
karlp: is there anything actually relevant here of their code remaining in any form?

include/libopencm3/stm32/f7/pwr.h
```diff
@@ -0,0 +1,284 @@
...
+
+/* WUPF1: Wakeup Pin flag for PA0 */
+#define PWR_CSR2_WUPF1			(1 << 0)
+
+/* --- Function prototypes ------------------------------------------------- */
+
+enum pwr_vos_scale {
+	PWR_SCALE1, // <= 180MHz w/o overdrive, <= 216MHz w/ overdrive
+	PWR_SCALE2, // <= 168MHz w/o overdrive, <= 180MHz w/ overdrive
+	PWR_SCALE3, // <= 144MHz
```
karlp: you can put inline /** comments */ here and doxygen will pick them as commentary for the enum values.

include/libopencm3/stm32/f7/pwr.h
```diff
@@ -0,0 +1,284 @@
...
+};
+
+BEGIN_DECLS
+
+void pwr_set_vos_scale(enum pwr_vos_scale scale);
```
karlp: Is this the same function name and enum name used in the other families?  (L0/L1/L4 to memory all have range scales)

include/libopencm3/stm32/f7/pwr.h
matthewlai: I don't think so. Removed.

include/libopencm3/stm32/f7/pwr.h
matthewlai: Done.

include/libopencm3/stm32/f7/pwr.h
matthewlai: Yeap! Copied from there.

matthewlai:
I just noticed there's a problem with the rcc_clock_setup_hse_3v3() function. It works fine if I copy the function into my code, but not when it's compiled into the library (frequency is wrong).

I am still debugging, but looking at the disassembly, it looks like rcc_osc_on() is inlined when this is part of the library (since it happens to be in the same file). That may or may not be the reason. Maybe there's some kind of delay required somewhere.

matthewlai:
Mystery solved. This PR is fine.

The problem is on F7 flash is mapped into two different regions - 0x08000000 like before (access through AXIM->AHB), and 0x00200000 (access through a dedicated ITCM bus, with ART cache). I was running through AXIM->AHB, without CPU caches enabled, which apparently made instruction alignment very important.

System clock rate was verified using a scope with MCO2 (clock-out from the chip).

I'll fix devices.data in another PR.

karlp:
this landed in 17553da sorry it didn't get closed at the time

PR #767, opened 2017-04-03
rename SEVEONPEND to SEVONPEND

tsaitgaist:
SEVEONPEND in probably a typo. According to the The ARM v7-M
Architecture Reference Manual SEVONPEND is the correct name.


karlp:

PR #768, opened 2017-04-03
stm32: fix the compilation in case of c99

amitesh-singh:
This fixes following warnings

../common/i2c_common_v2.c:431:3: error: 'for' loop initial declarations are only allowed in C99 or C11 mode
   for (size_t i = 0; i < rn; i++) {
   ^
../common/i2c_common_v2.c:431:3: note: use option -std=c99, -std=gnu99, -std=c11 or -std=gnu11 to compile your code


karlp:
if you're using a toolchain that's still defaulting to c89, I'll go ahead and add -std=c99 long before I go and live in the stone ages IMO :)  All the recent g-a-e toolchains are 5.x series, which switched to c99 by default.

PR #770, opened 2017-04-10
stm32f4: adc: Add VBat sensor enable/disable

0xdec:
Fairly straightforward addition of enable/disable functions for VBat sensor.


karlp:
merged via: 21b23f1ff1e0

PR #771, opened 2017-04-10
stm32f4: adc: Correct ADC documentation

0xdec:
Fix the demo code, add info about different connections for STM32F42x/STM32F43x, and remove errant param annotations.

### DIFF
--- a/lib/stm32/f4/adc.c
+++ b/lib/stm32/f4/adc.c
@@ -19,7 +19,7 @@ of 8. The ADC resolution can be set to 12, 10, 8 or 6 bits.
 
 Each A/D converter has up to 19 channels:
 @li On ADC1 the analog channels 16 is internally connected to the temperature
-sensor, channel 17 to V<sub>REFINT</sub>, and channel 18 to V<sub>BATT</sub>.
+sensor, channel 17 to V<sub>REFINT</sub>, and channel 18 to V<sub>BAT</sub>.
 @li On ADC2 and ADC3 the analog channels 16 - 18 are not used.
 
 The conversions can occur as a one-off conversion whereby the process stops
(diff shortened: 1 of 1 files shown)


lib/stm32/f4/adc.c
```diff
@@ -47,17 +47,17 @@ independent.
...
 @code
 gpio_mode_setup(GPIOA, GPIO_MODE_ANALOG, GPIO_PUPD_NONE, GPIO1);
-rcc_peripheral_enable_clock(&RCC_APB2ENR, RCC_APB2ENR_ADC1EN);
-adc_set_clk_prescale(RCC_CFGR_ADCPRE_BY2);
+rcc_periph_clock_enable(RCC_ADC1);
+adc_set_clk_prescale(ADC_CCR_ADCPRE_BY2);
 adc_disable_scan_mode(ADC1);
 adc_set_single_conversion_mode(ADC1);
-adc_set_sample_time(ADC1, ADC_CHANNEL0, ADC_SMPR1_SMP_1DOT5CYC);
+adc_set_sample_time(ADC1, ADC_CHANNEL0, ADC_SMPR_SMP_3CYC);
```
karlp: probably better here would be explaining how to choose the value, or simply putting in "xxxxx" :)

lib/stm32/f4/adc.c
```diff
@@ -406,13 +406,13 @@ bool adc_awd(uint32_t adc)
 /*---------------------------------------------------------------------------*/
 /** @brief ADC Enable The Temperature Sensor
 
-This enables both the sensor and the reference voltage measurements on channels
-16 and 17. These are only available on ADC1 channel 16 and 17 respectively.
-
-@param[in] adc Unsigned int32. ADC block register address base @ref adc_reg_base
```
karlp: Thanks for this,  I saw it in the context of your earlier diff and was going to strip it out when I got to it.

karlp:
looks reasonable enough, though tempting to remove that intro section altogether and simply refer to an example project.

lib/stm32/f4/adc.c
0xdec: Okay, I'll just put xxxx

0xdec:
Yeah I figured I would just leave it in because I think it's still useful to be able to see example code without an internet connection.

lib/stm32/f4/adc.c
0xdec: Added a short explanation

karlp:
pulled into my "next" branch, pure cleanup, nothing really to hold this up

karlp:
landed via c7d46c4f thanks

PR #774, opened 2017-04-17
Updated makefile to be able to build for specific target families only.

fjulian79:
Hi .. 

I'm tough it would be nice to be able to build libopencm3 only for a specific target family if desired. I like that as it saves time and avoids unneeded output in the build console. 

All what has to be done to allow this is to set TARGETS only if absent to the default as it has been done so far. When a user sets TARGETS (e.g. make TARGETS=stm32/f1) only this part of the library will be build. Furthermore I have changed the way the irq.json files are collected to get only the needed ones.

Br. J.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,14 +27,14 @@ space:=
 space+=
 SRCLIBDIR:= $(subst $(space),\$(space),$(realpath lib))
 
-TARGETS := stm32/f0 stm32/f1 stm32/f2 stm32/f3 stm32/f4 stm32/f7
-TARGETS += stm32/l0 stm32/l1 stm32/l4
-TARGETS += lpc13xx lpc17xx lpc43xx/m4 lpc43xx/m0
-TARGETS += lm3s lm4f
-TARGETS += efm32/tg efm32/g efm32/lg efm32/gg
-TARGETS += sam/3a sam/3n sam/3s sam/3u sam/3x
-TARGETS += sam/d
-TARGETS += vf6xx
+TARGETS ?=	stm32/f0 stm32/f1 stm32/f2 stm32/f3 stm32/f4 stm32/f7 \
+		stm32/l0 stm32/l1 stm32/l4 \
+		lpc13xx lpc17xx lpc43xx/m4 lpc43xx/m0 \
+		lm3s lm4f \
+		efm32/tg efm32/g efm32/lg efm32/gg \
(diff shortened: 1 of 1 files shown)


karlp:
The reason this hasn't been done in the past is to make sure people don't try submitting their fixes that weren't compiled against the rest of the codebase even.  It's only a one off build time, and work has been done to parallelize it to some extent, so for your first build, just do make -j8 and move on.

The irq fix is probably "correct" regardless, but I'm not super keen on encouraging this sort of usage.  (I'm open to arguments on this)

fjulian79:
Well thats a good point .. I haven't had that in mind when creating my request.

My use case has been a makefile based build system for mcu's (stm32, atmega, etc) I'm working on. At first my plan was to treat third party libraries (as libopencm3 is one for me) like my own ones. This would mean to have dependencies to them and therefore they are in the build chain. I found it useless to process all unused cpu families when calling make for my project .. which leads to "make -C ./libopencm3". 

Right now I I have switched to special treatment for third party libs by excluding them from my default build chain. The drawback is that I have to remove dependencies from my libs/aps to libopencm3. Well its not perfect for me in this way, but on the other side .. in a perfect world, developers would not commit untested code ;-)

My changes regarding the irq files have been made as I want to be consequent .. why processing all irq files when building just one target .. make use of it if you like it. 

Br J.

zyp:
In my opinion, building unused targets, especially when included into a project as a submodule, is a waste of time.

Verifying that changes doesn't break anything is the job of a CI system, not every user of something built on libopencm3.

fjulian79:
Well I can follow all those arguments ;-)

Hence that my change does not affect the default behavior as all libs will be build when calling just make without arguments .. so verifying change requests or patches can be done as before.

Ps. Time is not my major point .. the increased build output is more annoying for me.

Peque:
@karlp Any updates on this matter? :innocent:

karlp:
I've pulled this into my "next" tree, now that we have travis autobuilding, I can't really refuse this anymore :)

feliwir:
How is it looking now?

karlp:
like I said, it's pulled into my next tree. you can find that on my personal github, but it's not permanent, and it will be rebased and commits edited, so don't rely on anything.

karlp:
This landed via 74dd24a0

PR #775, opened 2017-04-17
stm32f7: flash: added proper support

matthewlai:
F7's flash has a few differences from F4's, and is not a superset. Some bits have been reused for different purposes.

### DIFF
--- a/include/libopencm3/stm32/f7/flash.h
+++ b/include/libopencm3/stm32/f7/flash.h
@@ -1,18 +1,22 @@
-/** @defgroup flash_defines FLASH Defines
- *
- * @ingroup STM32F7xx_defines
- *
- * @brief Defined Constants and Types for the STM32F7xx FLASH Memory
- *
- * @version 1.0.0
- *
- * @date 14 January 2014
- *
- * LGPL License Terms @ref lgpl_license
- */
+#ifndef LIBOPENCM3_FLASH_H
+#define LIBOPENCM3_FLASH_H
 
+/** @addtogroup flash_defines
+ *
+ * @author @htmlonly &copy; @endhtmlonly 2017
+ * Matthew Lai <m@matthewlai.ca>
+ * @author @htmlonly &copy; @endhtmlonly 2010
+ * Thomas Otto <tommi@viadmin.org>
+ * @author @htmlonly &copy; @endhtmlonly 2010
+ * Mark Butler <mbutler@physics.otago.ac.nz>
+ *
+ */
 /*
  * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2017 Matthew Lai <m@matthewlai.ca>
+ * Copyright (C) 2010 Thomas Otto <tommi@viadmin.org>
+ * Copyright (C) 2010 Mark Butler <mbutler@physics.otago.ac.nz>
  *
  * This library is free software: you can redistribute it and/or modify
  * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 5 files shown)


include/libopencm3/stm32/f7/flash.h
```diff
@@ -28,10 +32,174 @@
...
+#define FLASH_CR			MMIO32(FLASH_MEM_INTERFACE_BASE + 0x10)
+#define FLASH_OPTCR			MMIO32(FLASH_MEM_INTERFACE_BASE + 0x14)
+#define FLASH_OPTCR1			MMIO32(FLASH_MEM_INTERFACE_BASE + 0x18)
+
+/* --- FLASH Keys -----------------------------------------------------------*/
+
+#define FLASH_KEYR_KEY1			((uint32_t)0x45670123)
+#define FLASH_KEYR_KEY2			((uint32_t)0xcdef89ab)
+#define FLASH_OPTKEYR_KEY1		((uint32_t)0x08192a3b)
+#define FLASH_OPTKEYR_KEY2		((uint32_t)0x4c5d6e7f)
```
karlp: Is the cast here necessary?  shouldn't they just get a UL suffix to be the correct literal instead?

include/libopencm3/stm32/f7/flash.h
```diff
@@ -28,10 +32,174 @@
...
+
+/* --- FLASH registers ----------------------------------------------------- */
```
karlp: I think this line is a copy/paste remnant

include/libopencm3/stm32/f7/flash.h
```diff
@@ -28,10 +32,174 @@
...
+#define FLASH_ACR_LATENCY_6WS		0x06
+#define FLASH_ACR_LATENCY_7WS		0x07
+#define FLASH_ACR_LATENCY_8WS		0x08
+#define FLASH_ACR_LATENCY_9WS		0x09
+#define FLASH_ACR_LATENCY_10WS		0x0a
+#define FLASH_ACR_LATENCY_11WS		0x0b
+#define FLASH_ACR_LATENCY_12WS		0x0c
+#define FLASH_ACR_LATENCY_13WS		0x0d
+#define FLASH_ACR_LATENCY_14WS		0x0e
+#define FLASH_ACR_LATENCY_15WS		0x0f
```
karlp: Personally I see no use in names like these.  The mask sure, but...
```
flash_set_ws(1);
flash_set_ws(FLASH_ACR_LATENCY_1WS);
```

I know which one I prefer.

lib/stm32/f7/Makefile
```diff
@@ -40,9 +40,9 @@ TGT_CFLAGS	= -Os -g \
 
 ARFLAGS		= rcs
 
-OBJS		= rcc.o gpio.o gpio_common_all.o gpio_common_f0234.o
+OBJS		= flash.o rcc.o gpio.o gpio_common_all.o gpio_common_f0234.o
 
-OBJS		+= rcc_common_all.o flash_common_f234.o flash_common_f24.o
```
karlp: There's _nothing_ in common with flash f234/f24 anymore?  I find that a little hard to believe.

lib/stm32/f7/flash.c
```diff
@@ -0,0 +1,478 @@
...
+	FLASH_CR |= psize << FLASH_CR_PROGRAM_SHIFT;
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief Issue Pipeline Stall
+
+Issue a pipeline stall to make sure all write operations completed.
+
+RM0385: After performing a data write operation and before polling the BSY bit
+to be cleared, the software can issue a DSB instruction to guarantee the
```
karlp: can? must? that's pretty vague wording!

lib/stm32/f7/flash.c
```diff
@@ -0,0 +1,478 @@
...
+
+RM0385: After performing a data write operation and before polling the BSY bit
+to be cleared, the software can issue a DSB instruction to guarantee the 
+completion of a previous data write operation.
+
+*/
+
+static inline void flash_pipeline_stall(void)
+{
+	asm volatile("dsb":::"memory");
```
karlp: We've fixed it by specifying c99 in the makefiles, but see https://github.com/libopencm3/libopencm3/pull/769/commits/aab09823475953e855ec497a951fe8269443dbb9 for more info.  (just use __asm__ all the time for portability)

lib/stm32/f7/flash.c
```diff
@@ -0,0 +1,478 @@
...
+
+	reg32 = FLASH_ACR;
+	reg32 &= ~(FLASH_ACR_LATENCY_MASK);
+	reg32 |= ws;
+	FLASH_ACR = reg32;
+
+	/* Wait until the new wait states take effect.
+	 * RM0385: Check that the new number of wait states is taken into
+	 * account to access the Flash memory by reading the FLASH_ACR register.
+	 */
```
karlp: F7 sounds super dumb!

lib/stm32/f7/flash.c
```diff
@@ -0,0 +1,478 @@
...
+/** @brief Disable the FLASH Prefetch Buffer
+
+Note carefully the clock restrictions under which the prefetch buffer may be
+set to disabled. See the reference manual for details.
+*/
+
+void flash_prefetch_disable(void)
+{
+	FLASH_ACR &= ~FLASH_ACR_PRFTEN;
+}
```
karlp: I know for a fact that these two are shared widely.  They could probably even go into a "flash_common_all" :smile:

lib/stm32/f7/flash.c
```diff
@@ -0,0 +1,478 @@
...
+The page should be checked to ensure that it was properly erased. A sector must
+first be fully erased before attempting to program it.
+
+See the reference manual or the FLASH programming manual for details.
+
+@param[in] sector (0 - 11 for some parts, 0-23 on others)
+@param program_size: 0 (8-bit), 1 (16-bit), 2 (32-bit), 3 (64-bit)
+*/
+
+void flash_erase_sector(uint8_t sector, uint32_t program_size)
```
karlp: why is program size a useful argument for the user when they're erasing a sector?

lib/stm32/f7/flash.c
```diff
@@ -0,0 +1,478 @@
...
+	FLASH_CR &= ~FLASH_CR_PG;		/* Disable the PG bit. */
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief Program a Data Block to FLASH
+
+This programs an arbitrary length data block to FLASH memory.
+The program error flag should be checked separately for the event that memory
+was not properly erased.
+
```
karlp: Should probably mention that you need to erase yourself first?

karlp:
The rcc wait state changes should be a different commit.  Does that exist elsewhere in another PR maybe?

include/libopencm3/stm32/f7/flash.h
matthewlai: Done.

include/libopencm3/stm32/f7/flash.h
matthewlai: Removed.

include/libopencm3/stm32/f7/flash.h
matthewlai: I agree. That's from copy and paste. Removed.

lib/stm32/f7/Makefile
matthewlai: There are a lot in common, but there are also many small differences scattered throughout the registers, and I didn't really have time to look at reference manuals for every series to see what's common. More importantly, I don't think people adding definitions to the common files in the future will actually look at every single reference manual to decide where to add their definitions, and we'll probably end up with extraneous and/or wrong things being defined for some families.

As an example, at the moment we are including flash_common_f24.h, which has a lot of wrong definitions for f7.

In this case I would advocate having a totally separate file because -
1. If it's correct (accurate reflection of the reference manual, which I believe it is now), it's unlikely that we will update it much in the future. This is all pretty low level stuff.
2. It's more readable. People only have to look at one file to see everything that's defined for the family.
3. It's more likely to stay correct, and not be broken by people making changes to flash_common_f12347.h without actually checking the reference manual of every single family.
4. It's easier to verify that it's correct, by just comparing one file to the reference manual for one family.

Downsides:
1. Code duplication means changes to shared parts need to be applied multiple times. But to update a shared file correctly would require looking at every reference manual, which is a lot of work. If people are actually doing that, is it much more work to update each header, too?

In any case, even if we still want to refactor, we can do that later and get this merged first? Since this is at least correct (AFAIK), and the current version isn't.

lib/stm32/f7/flash.c
matthewlai: Yeah that's pretty shitty!

lib/stm32/f7/flash.c
matthewlai: Done.

lib/stm32/f7/flash.c
matthewlai: This is actually in the F4 reference manual, too, at least... we just never bothered.

lib/stm32/f7/flash.c
matthewlai: Yes, probably, if we want to do that (please see comment above).

lib/stm32/f7/flash.c
matthewlai: Because maximum program size (how many bytes erased at once) depends on supply voltage.

lib/stm32/f7/flash.c
matthewlai: Done.

matthewlai:
The RCC change needs to be in this commit because this commit changes flash interface, and would break rcc otherwise.

lib/stm32/f7/flash.c
karlp: Ok.  fair enough then, but perhaps that could be in the doc comment?

karlp:
This all landed in ba0e042 and was then heavily refactored later to get rid of all the actual duplication.  see https://github.com/libopencm3/libopencm3/pull/949

PR #776, opened 2017-04-20
F7 memory map fixes

matthewlai:
There are actually 2 groups of F7s - one with 64KB CCM and SP FPU, and one with 128KB CCM and DP FPU. They have different RAM offsets.

I split them into groups, but kept both FPU settings at SP for now, since we only build one library for the whole family. I believe it's theoretically possible to link SP code with DP code if they are both hard float, but I'll let someone who actually needs DP make that change and test it.

I also changed flash offset from 0x08000000 to 0x00200000. On the F7, accessing flash through the former goes through the AXIM-to-AHB bridge, and accessing it through the latter goes through a dedicated ITCM bus with ART cache + prefetch, which is much faster.
http://www.st.com/content/ccc/resource/technical/document/datasheet/96/ed/61/9b/e0/6c/45/0b/DM00166116.pdf/files/DM00166116.pdf/jcr:content/translations/en.DM00166116.pdf (Page 19).

However, the ITCM bus is read-only, and flashing still has to be done to the original 0x08000000 address. I do that by flashing the .bin file and specifying the address manually. Not sure if this will affect people using the elf file? Ideally we want the elf file to still place everything at 0x08000000, but all the branches in the code need to branch to ITCM addresses.


karlp:
Unfortunately the concerns about flashing via elfs make me rather hesitant to try merging this as is.  Was there also just a simpler problem of the CCM sizes?  Could that be merged as is?  I know many people use "load" via gdb, and that needs the right addresses.

matthewlai:
Yeah breaking ELF is pretty bad.

In this case I think we need to define a new region in linker script and use something like ">itcm AT >rom" for the .text section, similar to what we do for .data. I know nothing about linker scripts, but will take a look when I have a chance.

Yeah the CCM problem is separate. That should be safe to merge. It's not a serious issue, though. We are just using some CCM as regular SRAM now, and wasting some SRAM.

matthewlai:
Added separate VMA address for ROM. ELF is all good now. Verified by flashing with GDB.

Only tested on F7, but it shouldn't change anything on pre-F7.

karlp:
Ok, hve been playing with this locally, and you can still flash now, but it's totally busted for debug for me at least :)

openocd will load, but no breakpoints work.  I've had to back it out of my tree for now.

karlp:
not a clue sorry, but not being able to use a debugger is a dealbreaker for me sorry.  If you can split out the device details portions in some way, we can look at that, but i'm not going to merge things that break debugging.

matthewlai:
Rest of the changes is just taking out the "ROM_CCM_OFF=0x00200000" part. Maybe add a commented out option there so people can choose?

karlp:
Possibly related http://openocd.zylin.com/#/c/5097/

Issue #447, opened 2015-05-23
USB-F103 Lock up due to control race


kuldeepdhaka:
if DIR is set
		if CTR_TX and CTR_RX (both) is set:
			switch(endpoint type)
			case control:
				if transfer is control-out:
					process CTR_RX
					process CTR_TX
				else transfer is control-in
					process CTR_TX
					process CTR_RX
			default: (other than control)
				process {CTR_TX, CTR_RX} how ever we want, they are two independent endpoints
		else [CTR_RX is only set]:
			process CTR_RX
	else [CTR_TX is only set]
		process CTR_TX

note:
endpoint0 is always control
[ ] -> denotes that it is known, no need to test

kuldeepdhaka:
@gmccollister statement:
CTR_RX and CTR_TX both become set but the OUT handler is called, resulting
in an endless stall loop. This situation doesn't happen frequently, I haven't
been able to figure exactly why it happens though I'm able to reproduce it
within a few minutes by blasting the device with non-stop control and bulk
endpoint traffic that also generates USART traffic. One theory I have is that
another interrupt is sometimes occuring after LAST_DATA_IN, giving the
STATUS_OUT complete enough time to come in.

test:
since i do not have f1, i tested the code on f0 (everything same, only the PMA is bit different).
https://github.com/kuldeepdhaka/libopencm3-examples/tree/usb/f103-lock-up-due-to-control-race
this is a usb device firmware and python code,
try to send control traffic and bulk traffic in parallel.

results:
when the uC is running at 8Mhz, the uC is not able to keep up with the incoming traffic and in both case (same with patch and without patch) there is equal error rate.

when the uC is running at 16Mhz, 24Mhz, 48Mhz, there is no problem (in bulk) though some control transfer do fail. (same with patch and without-patch)

conclusion:
i find that theoritically the problem exists but was unable to reproduce it.
so, until we cannot reproduce the difference, we cannot proceed further.

devanlai:
As an additional data point, I've been running into this issue on a custom STM32F042C6 board shortly after enumeration (I can see the first few USB string descriptors are retrieved successfully, but some of the interfaces whose strings are retrieved later fail because of the infinite stall). The workaround in #446 and in the commits that reference this issue seems to resolve the symptoms.

PR #777, opened 2017-04-20
stm32: usb: Fix handling of CTR_RX

zyp:
The way the st_usbfs code handles the CTR_RX flag is bad and can result in both interrupt storms and lost data.

The current code clears CTR_RX either when ep_read() is called or when poll() is called and there is no callback configured to handle the event.

In the first case, the responsibility for the event being cleared is delegated away from the poll() function to the callback function, expecting it to call ep_read(). In a corner case where the callback doesn't read (e.g. in the control request handler code when it's in an unexpected state) the event won't be cleared before poll() returns, potentially resulting in a stuck endpoint or an interrupt storm if poll() is called from the USB ISR.

In the second case, CTR_RX flag is being cleared if there is no callback function configured. This however happens regardless of event type, so a TX event without a configured callback will clear any pending RX event on the same endpoint, resulting in lost data. The documentation for the DIR bit in USB_ISTR states that if both TX and RX events are pending, RX will be reported first, so the window for this race condition is pretty slim.

This patch fixes both issues by always clearing CTR_RX once we've determined that we're processing an RX event. CTR_RX is not used for anything other than for poll() to determine that there's an RX event to process, so this should not break any functionality for existing code.

I don't have a test case to demonstrate the problem and the fix, but the patch is simple and should be quick to review manually. The logic error in the original code is evident by just reading it, and existing test cases should be able to verify that this fix doesn't break usb functionality.

This problem is possibly the root cause of #447 and related issues.

### DIFF
--- a/lib/stm32/common/st_usbfs_core.c
+++ b/lib/stm32/common/st_usbfs_core.c
@@ -205,7 +205,6 @@ uint16_t st_usbfs_ep_read_packet(usbd_device *dev, uint8_t addr,
 
 	len = MIN(USB_GET_EP_RX_COUNT(addr) & 0x3ff, len);
 	st_usbfs_copy_from_pm(buf, USB_GET_EP_RX_BUFF(addr), len);
-	USB_CLR_EP_RX_CTR(addr);
 
 	if (!st_usbfs_force_nak[addr]) {
 		USB_SET_EP_RX_STAT(addr, USB_EP_RX_STAT_VALID);
@@ -236,6 +235,7 @@ void st_usbfs_poll(usbd_device *dev)
 			} else {
 				type = USB_TRANSACTION_OUT;
 			}
+			USB_CLR_EP_RX_CTR(ep);
 		} else {
 			type = USB_TRANSACTION_IN;
 			USB_CLR_EP_TX_CTR(ep);
@@ -243,8 +243,6 @@ void st_usbfs_poll(usbd_device *dev)
 
 		if (dev->user_callback_ctr[ep][type]) {
 			dev->user_callback_ctr[ep][type] (dev, ep);
-		} else {
-			USB_CLR_EP_RX_CTR(ep);
 		}
 	}
 
--- a/lib/stm32/common/st_usbfs_core.c
+++ b/lib/stm32/common/st_usbfs_core.c
@@ -235,7 +235,6 @@ void st_usbfs_poll(usbd_device *dev)
 			} else {
 				type = USB_TRANSACTION_OUT;
 			}
-			USB_CLR_EP_RX_CTR(ep);
 		} else {
 			type = USB_TRANSACTION_IN;
 			USB_CLR_EP_TX_CTR(ep);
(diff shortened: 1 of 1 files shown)


jhol:
I've tested this patch, and it fixes a problem I've been having with vendor control requests being sporadically lost if there is more than 1-2 microseconds delay between calls to `usbd_poll` in the main loop.

In the error I encountered, if a EP0 DATA OUT transaction (e.g. transfer of the device strings) was immediately followed by an EP0 DATA IN transaction, `_usbd_control_out` would be called, but `control_state.state` would be set to `DATA_IN`.

I tested it with a firmware that exposes a few interfaces: 1x CDC-ACM, 1xDFU, has a few device strings. The main loop calls `usbd_poll` every loop, with a `nop` delay-loop inserted to control the delay between USB polls.

I then have a script which:

1. Resets the device
2. Waits for the device to configure, and for the `/dev/ACM0` device node to appear.
3. Attempts to make the vendor control request using pyusb.

I found that if my delay loop was set to loop to more than 128 times @ 48MHz i.e. a few microseconds delay, the firmware would enter the state described above if a vendor control request 

This patch completely fixes the issue. Since applying the patch I haven't seen the problem behavior.

jhol:
I've published a test firmware to demonstrate the issue: https://github.com/jhol/opencm3-usb-error-demo

zyp:
@jhol's test case demonstrates a race condition in the control state machine, ref. #779.

The race condition causes the callback for the status OUT stage to be called in an unexpected state, leading to ep_read() not being called, leading to CTR_RX not being cleared, leading to this all repeating on every following invocation of poll().

This patch does not solve the original race condition, but clearing CTR_RX properly avoids getting stuck and allows the stack to fully recover.

zyp:
I just read up on how the st_usbfs core handles SETUP packets. The reference manual states that a SETUP token will always be ACKed even when STAT_RX is set to NAK or STALL, but the data will be dropped if CTR_RX is still set from the previous transfer. The implication of this is that once CTR_RX is cleared, a SETUP packet may overwrite the previously received packet even though STAT_RX has not been set to VALID yet.

This means that we should read the data before clearing CTR_RX, otherwise we have a new potential race condition in the case of aborted control transfers. I'll add another commit to this PR, moving clearing of CTR_RX back to after the callback is called.

zyp:
Further testing reveals that there seems to be an undocumented potential race condition in the hardware itself.

We see that for some control writes, the SETUP bit in USB_EP0R doesn't get cleared after receiving a SETUP packet, leading to the following OUT packet also being interpreted as a SETUP packet by the stack. @jhol's bus captures indicates that when this is happening, CTR_RX is being cleared around the border between the OUT token and the DATA token.

I suspect that this is caused by some sort of race condition in the hardware state machine itself, related to the fact that OUT flow control depends on STAT_RX while SETUP only relies on CTR_RX. Setting STAT_RX to VALID only after clearing CTR_RX seems to avoid this problem.

I've made a quick fix by moving handling of STAT_RX also out of ep_read(), but that is turning the OUT flow control code into an even bigger mess than it already is, so I'm starting to think it would be better to hold off on this entire PR until we can redo the entire OUT flow control with a more sane API.

zyp:
I'm closing this PR since it's not fit for merging and I'm not going to work more on it any time soon. The original problem should have been fixed by #785. Might revisit this later if more related issues show up.

PR #780, opened 2017-04-22
F7 clocks

matthewlai:
This is on top of #766 (pwr support) and #775 (flash support).

I added configurations for a few more clock speeds, and took PLLM out of the struct to be set at application time. This way we don't need a copy of the array for each HSE clock frequency we want (the ST boards use several different ones already) and one more for HSI.

All settings tested to at least boot up and blink an LED at the right frequency.

### DIFF
--- a/include/libopencm3/stm32/f7/flash.h
+++ b/include/libopencm3/stm32/f7/flash.h
@@ -1,18 +1,22 @@
-/** @defgroup flash_defines FLASH Defines
- *
- * @ingroup STM32F7xx_defines
- *
- * @brief Defined Constants and Types for the STM32F7xx FLASH Memory
- *
- * @version 1.0.0
- *
- * @date 14 January 2014
- *
- * LGPL License Terms @ref lgpl_license
- */
+#ifndef LIBOPENCM3_FLASH_H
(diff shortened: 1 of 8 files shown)


karlp:
I'm looking at the rcc clock struct one, what are the motivations for all the other intermediate speeds? are any of them significant in any way?

matthewlai:
Just thought it would be nice to have some intermediate speeds for saving power.

karlp:
your trick of setting pllm explicitly to the external crystal is... nice, but it ignores the explicit recommendations in the ref man that you should target 2Mhz here, not 1Mhz.

"It is recommended to select a frequency of 2 MHz to limit PLL jitter."

I'm not sure there's really all that many clock configs we need, and if we drop all the arbitrary intermediary speeds, it might be quite manageable?

matthewlai:
Well, the obvious solution is to use hse / 2 for PLLM instead, and divide all the PLLN's by 2.

It just seems to me that it's a bit verbose to have a set of structs for each HSE (+HSI), when only PLLM needs to be different, no matter how many clock settings we end up including.

karlp:
st tools ignore the 2mhz recommendation anyway, I think we'll just stick with it .)

PR #781, opened 2017-04-22
Remove empty libopencmsis directories when cleaning

joostrijneveld:
When calling `make clean`, some files and directories remain. This takes a step towards cleaning up some of the produced empty directories.


karlp:
I never was really excited about this, but I thought, hey, what the hell so I puilled it to try it out and merge.  However, this still leaves directories like include/libopencmsis/efm32 and /stm32 so it feels like... better, but why add python code that we have to maintain forever when it's just removing a partial heirarchy in a .gitignored directory already.  I'm afraid I really don't see any meaningful gain from this, to justify adding special python support for it.  Feel free to justify it further, but as it stands, I don't see a good reason to merge this.

PR #785, opened 2017-04-27
usb: Ensure control events are handled in order

### LINKED ISSUE #779: Race condition in USB control state machine
The USB control state machine has a race condition in the handling of control reads. If both the data IN packet completes and the status OUT packet arrives between two invocations of poll(), the status OUT packet might be passed to the control state machine first, before it's expected.

This was first reported in #668, ref. [my comment](https://github.com/libopencm3/libopencm3/issues/668#issuecomment-231044266) that describes the problem and my proposed solution. The implemented solution to #668 instead works around the problem by modifying the dwc_otg driver to always process TX events before RX events if both are pending.

During testing of #777 we see the same problem happening on st_usbfs based devices. The DIR bit in USB_ISTR will indicate RX if both events are pending, causing the RX event to be processed first.

To solve this, we could either ignore the DIR bit and manually check CTR_RX and CTR_TX to get the same behavior as the dwc_otg driver, or we could implement the solution that I outlined in the referenced comment that would make the control state machine more robust, not relying on the specific timing of each individual device driver.

zyp:
Use EP0 OUT flow control to NAK OUT packets when we're not yet expecting any. This prevents the status OUT event from arriving while the control state machine is still expecting the data IN completion event.

This fixes #779, and testing with the test case from #777 shows correct behavior regardless of whether #777 is also applied or not.

I haven't tested on any other devices than F0 yet. Since this change is in the common USB stack, it ought to be tested on other devices too, to ensure there's no regressions with other drivers.

### DIFF
--- a/lib/usb/usb_control.c
+++ b/lib/usb/usb_control.c
@@ -211,6 +211,8 @@ static void usb_control_setup_write(usbd_device *usbd_dev,
 	} else {
 		usbd_dev->control_state.state = LAST_DATA_OUT;
 	}
+
+	usbd_ep_nak_set(usbd_dev, 0, 0);
 }
 
 /* Do not appear to belong to the API, so are omitted from docs */
@@ -222,6 +224,8 @@ void _usbd_control_setup(usbd_device *usbd_dev, uint8_t ea)
 	(void)ea;
 
 	usbd_dev->control_state.complete = NULL;
+
+	usbd_ep_nak_set(usbd_dev, 0, 1);
 
 	if (usbd_ep_read_packet(usbd_dev, 0, req, 8) != 8) {
 		stall_transaction(usbd_dev);
@@ -294,6 +298,7 @@ void _usbd_control_in(usbd_device *usbd_dev, uint8_t ea)
 		break;
 	case LAST_DATA_IN:
 		usbd_dev->control_state.state = STATUS_OUT;
+		usbd_ep_nak_set(usbd_dev, 0, 0);
 		break;
 	case STATUS_IN:
 		if (usbd_dev->control_state.complete) {


karlp:
This is good.  My modifications to gadget-zero to delay 200usecs in each loop kept working fine on f4, but was failing on l0 (and presumably f1 too, I'll do more comprehensive tests before I collapse all of this)  after this patch is applied, l0 and f4 both still pass all tests.  Can I assume that #777 should simply not be applied?

zyp:
Yeah, this patch is independent of #777, and #777 as it stands is not ready for merging.

rkreis:
This also fixes my real-world use case, where an STM32L052 based USB device fails to connect to the host.

corecode:
this makes my system work better, not still not completely right.

karlp:

I know itt's nott the final answer, but it's better

H2OBrain:
This commit introduced the word loss again for me. Also other control transfers failed and caused libusb (at least on windows) to not "work" through. They just halted like when you debug a device, the transmission never finishes.

corecode:
@H2OBrain can you take a logic analyzer/scope trace of the handshake?  You might be seeing the same thing I'm seeing on a GD32 chip.

H2OBrain:
unfortunately i do not own a logic analyser which can decode usb.. i could look a the the traffic using wireshark though.. but in general the usb stack works fine for me without this patch, so i would have to generate the situation which causes the need for this patch first on the stm32f429i-disc1 board..

rkreis:
Keep in mind that a minimum polling rate is required for USB control
transfers to work correctly.  That is a requirement of the hardware and
we can not work around that.  So, as a first test, could you increase
the polling rate and see if that fixes the problem?  You can achieve
that by calling usb_poll more often, removing slow operations from your
main loop or increasing the clock rate.

H2OBrain:
Our devices were working for 1 year (24/7) without a crash. After i updated with this patch 3 of 15 devices crashed in one month and 1 crashed always after a couple control transfers. Also our custom control transfer lost 4 bytes in ~80% of the cases.
I can switch to using a timer interrupt for the usb polling, to not get timing interferences with the display (currently I am polling every cycle of the main-loop) to test if it runs more stable though.

zyp:
I've opened #873 to track the regression. Please take further discussion there.

mvirkkunen:
@H2OBrain If you at least have access to a scope you can wiggle a GPIO line on every USB poll, probe that and make sure the timing is OK. (In a pinch this should be slow enough to see even using a computer sound card as an "oscilloscope").

PR #786, opened 2017-04-30
rename some f4 bit definitions

jhalmen:
this was #764, accidentaly moved stuff around, so i'm opening this again.


karlp:
So.  This is valid, and generally "good" however it will break a lot of existing code.  Unfortunately, the choice of doing all of these bits with #defines means we can't simply add __attribute__("deprecated") to them and add duplicates with the new/better names.  For this reason, I'm not really itching on merging this right now.  I'm considering finding some time to make a whole slew of breaking changes like this, and have a little doc that covers the migration, but in the meantime it's .... just not fun to try and think forward on it.

karlp:
The ADC1 is a bit different, I actually simply don't to merge that one, because it would look like just another adc/adc1 alias, but for all the _other_ f4s, there is only a single reset bit, even when there's multiple enable bits.  I think the confusion of having enable_1,2,3 and reset/reset1 would be... odd.  The IOP/GPIO thing comments are as before.

karlp:
closed via the merge of https://github.com/libopencm3/libopencm3/pull/871 for the GPIO ones at least.

PR #787, opened 2017-04-30
devices.data: stm32f0: add stm32f042x[46] devices

gkelly:
Add both of the devices described in [this datasheet](http://www.st.com/content/ccc/resource/technical/document/datasheet/52/ad/d0/80/e6/be/40/ad/DM00105814.pdf/files/DM00105814.pdf/jcr:content/translations/en.DM00105814.pdf).


karlp:
merged thanks!

PR #788, opened 2017-05-01
stm32: rcc: extract osc_bypass functions

karlp:
rcc_osc_bypass_enable and rcc_osc_bypass_disable have been copy/pasted
around for the last time!  There's a compile bit to check for L0/L1, but
otherwise this is just code duplication for no gain.


karlp:

PR #789, opened 2017-05-01
stm32l0/l1 common flash code

karlp:


karlp:

PR #795, opened 2017-06-08
Support for reading CAN timestamp

FabianInostroza:
This enable us to get the hardware timestamp of a CAN message.


lib/stm32/can.c
```diff
@@ -522,6 +523,9 @@ void can_receive(uint32_t canport, uint8_t fifo, bool release, uint32_t *id,
 	 * memory to the fast internal RAM
 	 */
 
+	if( timestamp )
+		*timestamp = (CAN_RDTxR(canport, fifo_id) & CAN_RDTxR_TIME_MASK) >> CAN_RDTxR_TIME_SHIFT;
+
```
karlp: Please follow the existing style here.  try "make stylecheck" on your new branch with this code.  The code is good, but just needs this style cleanup.  Please expand the documentation the timestamp parameter too, that it's optional and to provide NULL.  (the unsigned int16 portion can be dropped from the doxygen, it's inferred from the types automatically)

(You need {} even for single line ifs, and the spacing is wrong)

karlp:
please add "stm32: can: " as a prefix on these commit subjects to.  YOu can just push -f to this branch, and github will upate the PR automatically.

FabianInostroza:
Done.

karlp:
Thanks, included in my staging now.

karlp:

PR #800, opened 2017-07-01
CAN: change Filter Match Index size

jhalmen:
Caught this by mistake, datasheet shows the FMI can only ever be 8 bit wide, so i set my variable to 8 bits, and gave that as an argument to the can_receive() function, not worrying about it.
The next variable in memory just happened to be my systick counter, which got reset to 0 every time i received a CAN message. Took me a while to figure out what was going on, but this fixes it, and brings it closer to what's in the datasheet.


jhalmen:
didn't mean all of those changes. i'll try again.

PR #803, opened 2017-07-05
STM32L1: Add functions to erase and write flash

br101:
Functions to erase and write program memory (flash):

void flash_erase_page(uint32_t address);
void flash_program_half_page(uint32_t address, uint32_t *data);
void flash_program_word(uint32_t address, uint32_t data);

Signed-off-by: Bruno Randolf <br1@einfach.org>

### DIFF
--- a/include/libopencm3/stm32/f7/usart.h
+++ b/include/libopencm3/stm32/f7/usart.h
@@ -1,0 +1,55 @@
+/** @defgroup usart_defines USART Defines
+ *
+ * @brief <b>Defined Constants and Types for the STM32F7xx USART</b>
+ *
+ * @ingroup STM32F7xx_defines
+ *
+ * @version 1.0.0
+ *
+ * @date 5 December 2012
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef LIBOPENCM3_USART_H
+#define LIBOPENCM3_USART_H
+
+#include <libopencm3/stm32/common/usart_common_all.h>
+#include <libopencm3/stm32/common/usart_common_v2.h>
+
+
+/** @defgroup usart_reg_base USART register base addresses
+ * @ingroup STM32F_usart_defines
+ * Holds all the U(S)ART peripherals supported.
+ * @{
+ */
+#define USART1				USART1_BASE
+#define USART2				USART2_BASE
+#define USART3				USART3_BASE
+#define UART4				UART4_BASE
+#define UART5				UART5_BASE
+/**@}*/
+
+BEGIN_DECLS
+
+END_DECLS
+
+#endif
+
--- a/include/libopencm3/stm32/usart.h
+++ b/include/libopencm3/stm32/usart.h
@@ -30,6 +30,8 @@
 #       include <libopencm3/stm32/f3/usart.h>
 #elif defined(STM32F4)
 #       include <libopencm3/stm32/f4/usart.h>
+#elif defined(STM32F7)
+#       include <libopencm3/stm32/f7/usart.h>
 #elif defined(STM32L0)
 #       include <libopencm3/stm32/l0/usart.h>
 #elif defined(STM32L1)
--- a/lib/stm32/f7/Makefile
+++ b/lib/stm32/f7/Makefile
@@ -48,6 +48,8 @@ OBJS		+= rcc_common_all.o flash_common_f234.o flash_common_f24.o
 
 OBJS		+= rng_common_v1.o
 
+OBJS		+= usart_common_all.o usart_common_v2.o
+
 VPATH += ../../usb:../:../../cm3:../common
 VPATH += ../../ethernet
 


lib/stm32/l1/flash.c
```diff
@@ -230,5 +230,50 @@ void eeprom_program_words(uint32_t address, uint32_t *data, int length_in_words)
...
+ *
+ * A half page is 32 words (128 byte).
+ *
+ * Half Page Write is possible only from SRAM, from external memory or from
+ * bank 1 to bank 2 and vice versa. Force into SRAM with the section attribute.
+ *
+ * @param[in] address must point to flash space, no checking!
+ * @param[in] data pointer to 128 bytes of data to write
+ */
+__attribute__((section(".data#")))
```
karlp: Unfortunately, this is the sort of thing that the user would have to control.  It's why I've not added this function before.  It needs correspondingly ld script changes that are pretty hard to work out a sane way of doing in the library by default.

lib/stm32/l1/flash.c
```diff
@@ -230,5 +230,50 @@ void eeprom_program_words(uint32_t address, uint32_t *data, int length_in_words)
...
 
+/** @brief Erase a page (256 Byte) of program memory (flash)
+ *
+ * @param[in] page_address must point to flash space, no checking!
+ */
+void flash_erase_page(uint32_t address)
+{
+	while (FLASH_SR & FLASH_SR_BSY);
+	FLASH_PECR |= (FLASH_PECR_ERASE | FLASH_PECR_PROG);
+	*(uint32_t*)address = 0;
```
karlp: iirc, the address has to be the beginning of a page?

lib/stm32/l1/flash.c
```diff
@@ -230,5 +230,50 @@ void eeprom_program_words(uint32_t address, uint32_t *data, int length_in_words)
...
+ * @param[in] address must point to flash space, no checking!
+ * @param[in] data pointer to 128 bytes of data to write
+ */
+__attribute__((section(".data#")))
+void flash_program_half_page(uint32_t address, uint32_t *data)
+{
+	uint32_t* p = (uint32_t*)address;
+	FLASH_PECR |= (FLASH_PECR_FPRG | FLASH_PECR_PROG);
+	while (FLASH_SR & FLASH_SR_BSY);
+	for (int i = 0; i < 32; i++)
```
karlp: style is that you _must_ have {} here.

lib/stm32/l1/flash.c
```diff
@@ -230,5 +230,50 @@ void eeprom_program_words(uint32_t address, uint32_t *data, int length_in_words)
...
+ * Half Page Write is possible only from SRAM, from external memory or from
+ * bank 1 to bank 2 and vice versa. Force into SRAM with the section attribute.
+ *
+ * @param[in] address must point to flash space, no checking!
+ * @param[in] data pointer to 128 bytes of data to write
+ */
+__attribute__((section(".data#")))
+void flash_program_half_page(uint32_t address, uint32_t *data)
+{
+	uint32_t* p = (uint32_t*)address;
```
karlp: same again, why not just make it a pointer?

lib/stm32/l1/flash.c
```diff
@@ -230,5 +230,50 @@ void eeprom_program_words(uint32_t address, uint32_t *data, int length_in_words)
...
+
+/** @brief Program memory: write one word
+ *
+ * @param[in] address must point to flash space, no checking!
+ * @param[in] data word to write
+ */
+void flash_program_word(uint32_t address, uint32_t data)
+{
+	while (FLASH_SR & FLASH_SR_BSY);
+	*(uint32_t*)address = data;
```
karlp: this is kinda useless.  It's blocking, eliminating that option from the user, but doesn't handle unlocking/locking?

lib/stm32/l1/flash.c
br101: Right. Still it would be a good reference to have this function in the library - what about just adding to the documentation that this needs to be placed into SRAM  or according to the rules (externam memory, other bank...) in project specific linker script?

lib/stm32/l1/flash.c
br101: Right it must be the beginning of the page. About the pointer, I was just following existing code in flash.c of other prcessors (f1/f0...) and in the same file `void eeprom_program_word(uint32_t address, uint32_t data)`

karlp:
For the half page write, I think perhaps the only realistic way of doing it is putting it in the examples repo.

lib/stm32/l1/flash.c
br101: Again, following the example in other flash_common*.c

lib/stm32/l1/flash.c
br101: Also this is similar to other flash operations e.g. in flash_common_f24.c `flash_program_half_word()` with the difference that there it is in a function `flash_wait_for_last_operation()`.

About locking/unlocking I reckon this should be done separately, since usually you'd want to write more than one word. Also it is similar to `flash_program_half_word()` in flash_common_f24.c

lib/stm32/l1/flash.c
br101: OK, I will remove this function, then

br101:
Updated the pull request to remove the half page programming and clarified the documentation of flash_erase_page(). Leaving the addresses as uint32_t as in most of the other flash functions throughout this library.

lib/stm32/l1/flash.c
karlp: right, fair enough

karlp:
Do you still have the half page flashing in a branch somewhere? I was going to try out a few methods of integrating it.

br101:
Sure, here it is:

/* Program memory: write half page which is 32 words (128 byte) */
// __attribute__((long_call, section(".data"))) may be needed in some cases?
__attribute__((section(".data")))
void flash_program_half_page(uint32_t address, uint32_t *data)
{
	uint32_t* p = (uint32_t*)address;
	FLASH_PECR |= (FLASH_PECR_FPRG | FLASH_PECR_PROG);
	while (FLASH_SR & FLASH_SR_BSY);
	for (int i = 0; i < 32; i++) {
		*p++ = *data++;
	}
	while (FLASH_SR & FLASH_SR_BSY);
	FLASH_PECR &= ~(FLASH_PECR_FPRG | FLASH_PECR_PROG);
}

You are welcome to take it like that, but if you prefer, I can make another pull request.

PR #807, opened 2017-07-15
Fix stm32 f1/f4 ADC_SMPRx_SMPy_MSK defines

vollst-induktion:
The `ADC_SMPRx_SMPy_MSK` macros in _stm32/f[14]/adc.h_ refer to the `ADC_SMPy_LSB` macros which are not defined. Instead the `ADC_SMPRx_SMPy_LSB` macros should be used.

Just one commit this time ;)


karlp:
merged in a42a058 with a reworded commit message, tthanks.

I think those definitions are absolutetly terrible, but at least let's fix them to be correct at least.

PR #809, opened 2017-07-24
Added SPI_SR_FRE constant (bit 8 of SPI_SR)

vampirefrog:


karlp:
What target are you working with?  Can we find a better place for this?  F2/4 already has SPI_SR_TIFRFE for instance, for this bit.

vampirefrog:
true, found it in stm32/f4/spi.h. in the reference manual it's simply called FRE, so I had trouble grepping for it. perhaps there should be an alias for it?

karlp:
What target are you working on?

vampirefrog:
F4

karlp:
I think the SPI_SR_TIFRFE should just be replaced with SR_FRE, not this addition.  Don't have time to look further and just fix that myself right this moment though, writing this for next time.

karlp:
this was resolved via bf125e9

PR #812, opened 2017-08-01
Fix typos in CAN register definitions

revzin:


karlp:
merged in b31b398 and b31b398^

Issue #557, opened 2015-10-30
adc_v2: adc_start_conversion_regular() causes a freeze if ADC is timer-triggered


MightyPork:
I'm working with stm32f3.

Timer is sending TRGO event to ADC, but it does not start converting until `ADSTART` is set.

I tried using `adc_start_conversion_regular()`, which seemed like the obvious right choice.

However, `adc_start_conversion_regular()` waits for `ADSTART` to be cleared after it sets it, which does not happen in all cases (see datasheet):

> It is cleared by hardware:
>   – in single conversion mode when software trigger is selected (EXTSEL=0x0): at the assertion of the End of Regular Conversion Sequence (EOS) flag.
>   – in all cases: after the execution of the ADSTP command, at the same time that ADSTP is cleared by hardware.
>     0: No ADC regular conversion is ongoing.
>     1: Write 1 to start regular conversions. Read 1 means that the ADC is operating and eventually converting a regular channel.

**Function with the bug**
https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/f3/adc.c#L487

``` c
void adc_start_conversion_regular(uint32_t adc)
{
    /* Start conversion on regular channels. */
    ADC_CR(adc) |= ADC_CR_ADSTART;

    /* Wait until the ADC starts the conversion. */
    while (ADC_CR(adc) & ADC_CR_ADSTART);
}
```

The while loop should be removed.

If I use `ADC_CR(adc) |= ADC_CR_ADSTART;` manually, it works fine.

karlp:
Totally correct. something that should be handled by further tests in https://github.com/karlp/libopencm3-tests/tree/master/tests/adc-power but I ran out of time there (for now)  needs some timer triggered stuff added to catch more differences like this.  This was some badly ported code in the original f3 commits (like a lot of that code unfortuntely :(

PR #815, opened 2017-08-21
adc: fix adc_start_conversion_regular()

xaxaxa:
the CR_ADSTART bit only gets cleared by hardware when the conversion of
all channels have completed (see RM0360 page 210), not when the
conversion has started as the code comment suggest, and this while loop
causes a hang if adc wait mode is used, as well as disallows any
possibility of reading all but the last channel when sequenced
conversion is used 
#557

### DIFF
--- a/lib/stm32/common/adc_common_v2.c
+++ b/lib/stm32/common/adc_common_v2.c
@@ -384,9 +384,6 @@ void adc_start_conversion_regular(uint32_t adc)
 {
 	/* Start conversion on regular channels. */
 	ADC_CR(adc) |= ADC_CR_ADSTART;
-
-	/* Wait until the ADC starts the conversion. */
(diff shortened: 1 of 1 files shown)


OpenUAS:
What keeps main devs from honouring this PR?

karlp:
Nothing other than time.  I try and look at all the adc code in one go, and I want to understand better how this even managed to get to where it is, if it's as broken as claimed.

karlp:
Also, this PR didn't fix the injected start, which had the same bug.  nor the comments, which were also wrong for the same reasons.

They've both been fixed in https://github.com/libopencm3/libopencm3/commit/01320881e7da131045050105086a6f40f453ae87  thanks for the pressure to get it fixed! never hurts

PR #819, opened 2017-08-30
DRAFT: vector: make common linker definitions available to users

karlp:
_data_loadaddr, _data, _edata, _ebss, _stack and vector_table all now
available in vector.h.

Suggested on IRC.  Feedback on the types and the general idea very welcome.


karlp:
this has landed.

PR #824, opened 2017-09-24
usb_standard: Care for terminating 0 in string handling.

UweBonnes:
When space for the UTF strings is allocated, the terminating 0 must be considered. Otherwise a string with length 64 will lead to errors when the hosts requests the string.

### DIFF
--- a/lib/usb/usb_standard.c
+++ b/lib/usb/usb_standard.c
@@ -196,7 +196,7 @@ static int usb_standard_get_descriptor(usbd_device *usbd_dev,
 			/* This string is returned as UTF16, hence the
 			 * multiplication
 			 */
-			sd->bLength = strlen(usbd_dev->strings[array_idx]) * 2 +
+			sd->bLength = (1 + strlen(usbd_dev->strings[array_idx])) * 2 +
 				      sizeof(sd->bLength) +
 				      sizeof(sd->bDescriptorType);
 


zyp:
I'm not sure what you're trying to do here, considering that string descriptors are not zero terminated. The length is specified explicitly so there's no terminator required. Besides, being UTF-16, every other byte would be zero for ascii characters, meaning zero would be useless as a terminator.

Can you elaborate on the original problem that you're trying to fix?

UweBonnes:
Zyp, this happend with a Black magic debug probe. When a string is exact 64 bytes long, strang things happenen. lusus -vv gives "error" for that string. (blackmagic)/scripts/stm32_mem.py crashes when accessing the string. And after plug in, enumaration happens immediate , but the ACM modems only appear after 5 seconds. By adding one word more to the allocation, things worked.

karlp:
Can you make a patch on the gadget-zero tests with a problematic length? should be fairly easy to see it there then.

UweBonnes:
Gareth just commented:
areth McMullin
@gsmcmullin
22:45
Ha. The problem is not that the string needs a trailing null, the problem is that the USB transfer needs a short packet to complete. If the size of the string descriptor is an exact multiple of bMaxPacketSize, a zero length packet should be sent at the end to complete the transfer. It looks like this is fixed in libopencm3 master, but I will need to do some testing before updating the pointer.

gsmcmullin:
This was fixed long ago in 3ed12b6fd.  The Black Magic probe firmware is using a very old version of libopencm3 that needs to be updated.

UweBonnes:
My workaround is smaller ;-)

karlp:
your workaround was _wrong_ :)

PR #830, opened 2017-09-30
STM32F0: Enable prefetch buffer in rcc_clock_setup_*()

adamheinrich:
This modification enables the prefetch buffer which makes code execution from the Flash memory faster.

Reasons for the modification:
- The `rcc_clock_setup_*()` functions should provide the maximum possible speed as described in 0259102560ad7d0be17d0b8c7a8ffecb04ab0650
- The code is now more consistent with other families as STM32L0/L1 functions `rcc_clock_setup_*()` also enable the prefetch buffer

I wanted to enable the prefetch buffer for STM32F1 parts as well, but it is already enabled after reset according to the reference manual (STM32F0 parts have the prefetch buffer disabled by default).

### DIFF
--- a/lib/stm32/f0/rcc.c
+++ b/lib/stm32/f0/rcc.c
@@ -543,6 +543,7 @@ void rcc_clock_setup_in_hse_8mhz_out_48mhz(void)
 	rcc_set_hpre(RCC_CFGR_HPRE_NODIV);
 	rcc_set_ppre(RCC_CFGR_PPRE_NODIV);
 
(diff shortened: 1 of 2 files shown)


karlp:
This seems totally reasonable to me.  Anyone else have thoughts on this?

karlp:
merged via rebase thanks.  It just turns up errors in f4 prefetch handling and the soup o f whether it should be enabled on l0, but hey, it's self consistent, and appears suitable for f0 :)

PR #834, opened 2017-10-04
doxygen overhaul for automatic file tracking

karlp:
Massive overhaul of doxygen.  Instead of listing dirs, and attempting to manage exclusion patterns, or manually updating file lists, we now parse .d files from the build, and augment with the few static doxygen files needed additionally.  See the second commit for full details.


karlp:
I've landed this now. auto generated docs should now be up to date.

PR #835, opened 2017-10-05
stm32f1: Fix bit definitions for CAN reset.

FabianInostroza:
Typo in the definitions. can_reset() call was not working


karlp:
Thanks! merged via rebase in 35ed692 after rewording

PR #838, opened 2017-10-12
STM32: add MASK defines in sdio.h

tsaitgaist:
also fixes a typo and renames two MSK to MASK, removes shifts from MSK


karlp:
merged in 3922cc7d thanks

PR #840, opened 2017-10-12
stm32: timer: Remove TIMER_IS_ADVANCED() checks

### LINKED ISSUE #839: STM32F303 timer "break" support
The function [`timer_enable_break_main_output`](https://github.com/libopencm3/libopencm3/blob/af7b2190a97ddadd9e227cc6b0a4cf18d49cba12/lib/stm32/common/timer_common_all.c#L1598) is no-op for any timers that aren't "advanced".

TIM15 on the STM32F303x(D/E) isn't "advanced" however it does have the "break" feature, which needs to be enabled for the output to work, even if the break feature is not used.

The workaround for anyone else who runs into this is to just enable it directly in the register:
`TIM_BDTR(TIM15) |= TIM_BDTR_MOE;`

zyp:
The TIMER_IS_ADVANCED() check is not correct for all timers on all targets, ref. #839. Instead of maintaining this check, I propose removing it entirely.

Rationale:
1. This code does only do anything if the user is trying to activate features on a timer where they don't exist. Invalid requests is arguably a user error that we don't have to handle, especially since silently ignoring it doesn't aid debugging.
2. The register bits involved reads zero and ignores writes on non-advanced timers, so letting the register accesses go through still results in silently ignoring invalid requests.

This fixes #839.

### DIFF
--- a/lib/stm32/common/timer_common_all.c
+++ b/lib/stm32/common/timer_common_all.c
@@ -114,18 +114,6 @@ knob.
 #include <libopencm3/stm32/timer.h>
 #include <libopencm3/stm32/rcc.h>
 
-#if (defined(TIM1_BASE) || defined(TIM8_BASE))
-#define ADVANCED_TIMERS 1
-#else
-#define ADVANCED_TIMERS 0
-#endif
-
-#if defined(TIM8)
-#define TIMER_IS_ADVANCED(periph) (((periph) == TIM1) || ((periph) == TIM8))
-#else
-#define TIMER_IS_ADVANCED(periph) ((periph) == TIM1)
-#endif
-
 /*---------------------------------------------------------------------------*/
 /** @brief Reset a Timer.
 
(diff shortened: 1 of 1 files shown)


karlp:
20:27 < karlp> did that compile properly?
20:27 < karlp> advanced timers was also added to be able to build the code on targets that simply didn't have 
               those registers
20:27 < karlp> if that compiles, it means timer register definitions are being pulled in for targets that don't 
               hv ethem, which is just something else.

*note to self to try and track down whether we really still need them*

zyp:
We already discussed this on IRC, but I'm summarizing it here for the future:

The ADVANCED_TIMERS define/check was added in 523943a3d205ec1d0121f9404548fd9cd4efd9f1 as part of adding L1 support. The runtime checks against TIM1/TIM8 already existed. Since L1 doesn't have TIM1/TIM8, those names are undefined, resulting in a compilation error until ifdeffed out.

Since I throw out all TIM1/TIM8 checks, there's no references to those names left, thus no need to keep the ifdef either.

As for the registers themselves, l1/timer.h pulls in common/timer_common_all.h which defines macros for the superset of all timers, so e.g. TIM_BDTR() is still available regardless of whether or not the particular chip we're building for has any timers with a BDTR register.

karlp:
merged, thanks.  Added more of the explantory text into the commit message too

PR #844, opened 2017-10-20
stm32: f4: Add DCMI peripheral register definitions

iqyx:


karlp:
:+1: Looks great thanks

karlp:
I was going to try and tidy up the doxygen grouping a bit, but ran into a bug that this exposes in the new doxygen generater.  I've pushed a fix for that, so you can see the doxygen generated for this now yourself.  I'll update it myself at some point when I merge it, if you haven't yourself by then.

karlp:
merged thanks.  I tweaked a tiny bit some doxygen stuff and then added some more.

PR #845, opened 2017-10-25
stm32: can: Fixed incorrect CAN_FMR_CAN2SB_SHIFT value

matthiasbock:
According to several reference manuals for MCUs of the STM32F4 series, the CAN_FMR_CAN2SB_SHIFT value should be 8, not 15.


karlp:
merged in 5fc4f48 thanks!  (for future reference, you normally want to use a separate branch in your repository for the PR, as otherwise any new commits in your master will come here too)

Issue #846, opened 2017-10-25
STM32 ADC_CDR register has the wrong address offset


LucasPickering:
The ADC_CDR register in `stm32/common/adc_common_c2_multi.h` has the wrong offset. It has 
`(adc) + 0x300 + 0xa`
but it _should_ be 
`(adc) + 0x300 + 0xc` 
This is verified by the [device datasheet](http://www.st.com/content/ccc/resource/technical/document/reference_manual/4a/19/6e/18/9d/92/43/32/DM00043574.pdf/files/DM00043574.pdf/jcr:content/translations/en.DM00043574.pdf) (p. 410).

I've tested this in my project, and the fix works.

karlp:
fixed via merge of the PR.  (just the PR by itself is normally enough, don't really need the issue too if you find this sort of thing again)  thanks again!

PR #847, opened 2017-10-25
stm32: Fix address offset for ADC_CDR register

LucasPickering:
See #846 
Tested and working in my project that uses this register.


karlp:
thanks for this! merged with a slight rewording

PR #849, opened 2017-10-29
Allows registration of non-continous description string

rgwan:
It can help register a special string index (0xee) for Windows's WCID for USBd stack which enables automatic WinUSB driver installation, just like HID or MSC class device.

### DIFF
--- a/lib/usb/usb.c
+++ b/lib/usb/usb.c
@@ -56,6 +56,8 @@ usbd_device *usbd_init(const usbd_driver *driver,
 	usbd_dev->num_strings = num_strings;
 	usbd_dev->ctrl_buf = control_buffer;
 	usbd_dev->ctrl_buf_len = control_buffer_size;
+	usbd_dev->extra_string_idx = -1;
+	usbd_dev->extra_string = NULL;
 
 	usbd_dev->user_callback_ctr[0][USB_TRANSACTION_SETUP] =
 	    _usbd_control_setup;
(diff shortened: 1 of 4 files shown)


lib/usb/usb_standard.c
```diff
@@ -174,6 +174,20 @@ static int usb_standard_get_descriptor(usbd_device *usbd_dev,
 				      sizeof(sd->wData[0]);
 
 			*len = MIN(*len, sd->bLength);
+		} else if (descr_idx == usbd_dev->extra_string_idx) {
+			/* This string is returned as UTF16, hence the
+			 * multiplication
+			 */
+			sd->bLength = strlen(usbd_dev->extra_string) * 2 +
```
karlp: If you're going to let people register custom strings, and those are proper usb strings, you probably don't want to rely on strlen either.

karlp:
Seems otherwise reasonable. WCID seems like a worthy goal

lib/usb/usb.c
```diff
@@ -94,6 +96,19 @@ void usbd_register_sof_callback(usbd_device *usbd_dev, void (*callback)(void))
 	usbd_dev->user_callback_sof = callback;
 }
 
+void usbd_register_extra_string(usbd_device *usbd_dev, int index, const char* string)
+{
+	if (string != NULL && index > 0)
```
karlp: given this, what's wrong with just using zero for the default of extra_string_idx?

karlp:
Reference for others https://github.com/pbatard/libwdi/wiki/WCID-Devices

rgwan:
@karlp Okay, I will cancel check on idx. And dont rely on strlen.

karlp:
Someone should add a usage of this into one of the gadget0 stacks or something, so we can make sure we don't break this sort of thing

karlp:
this has now landed in master via c13c2b3b3c4 woooooo only three years boooooo

PR #850, opened 2017-10-31
use arm-none-eabi-gcc-ar to support building LTO static libraries

walmis:
Now we can build libopencm3 using command 
`CFLAGS="-flto" LDFLAGS="-flto" make`


davidgfnet:
Can't we merge this? There's no harm on using this at all really.

karlp:
What version of your toolchain do you have?  binutils upstreams says you needed this in 4.9 era, but later than that it should support autoloading the plugin:  https://gcc.gnu.org/wiki/LinkTimeOptimizationFAQ

davidgfnet:
It all boils down (I think) whether you built binutils with plugin support and you have a recent compiler. But I'm using gcc 7.1 and you still need to use gcc-ar (and folks -nm,-ranlib). Adding a switch somewhere where one can enable LTO and it automatically uses gcc-ar instead of ar would be very useful.
Thanks!

karlp:
I experimented a little with building this all with clang, but it didn't go particularly well, and that was still using gcc for linker.  So I'm inclined to just go with this for now, until we get some other options.  Will pull it locally and do some more testing.

karlp:
this has landed

davidgfnet:
Is there any way we can help merging this? This should be harmless no matter whether one uses LTO or not, provided they use a "recent" compiler.

karlp:
So, I've got this locally, and I can rebuild the library ok, using your example command line.  And can build my _original_ demo projects with those, (compiled only, haven't even gotten to testing them yet) but I can't build LTO with my projects.    Any tips/advice?

```
karlp@teros:~/src/libopencm3-tests/tests/adc-power (hw1-hacking *)$ make CFLAGS="-flto" LDFLAGS="-flto" V=1
make -f Makefile.stm32l4-disco all
make[1]: Entering directory '/home/karlp/src/libopencm3-tests/tests/adc-power'
  CC	main-stm32l4-disco.c
arm-none-eabi-gcc -Os -std=c99 -ggdb3 -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -fno-common -ffunction-sections -fdata-sections -Wextra -Wshadow -Wno-unused-variable -Wimplicit-function-declaration -Wredundant-decls -Wstrict-prototypes -Wmissing-prototypes -flto -MD -Wall -Wundef -I. -I../../shared -DSTM32L4 -DSTM32L476VGT6 -I../../libopencm3//include -o bin-stm32l4-disco/main-stm32l4-disco.o -c main-stm32l4-disco.c
  CC	adc-power.c
arm-none-eabi-gcc -Os -std=c99 -ggdb3 -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -fno-common -ffunction-sections -fdata-sections -Wextra -Wshadow -Wno-unused-variable -Wimplicit-function-declaration -Wredundant-decls -Wstrict-prototypes -Wmissing-prototypes -flto -MD -Wall -Wundef -I. -I../../shared -DSTM32L4 -DSTM32L476VGT6 -I../../libopencm3//include -o bin-stm32l4-disco/adc-power.o -c adc-power.c
  CC	../../shared/trace.c
arm-none-eabi-gcc -Os -std=c99 -ggdb3 -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -fno-common -ffunction-sections -fdata-sections -Wextra -Wshadow -Wno-unused-variable -Wimplicit-function-declaration -Wredundant-decls -Wstrict-prototypes -Wmissing-prototypes -flto -MD -Wall -Wundef -I. -I../../shared -DSTM32L4 -DSTM32L476VGT6 -I../../libopencm3//include -o bin-stm32l4-disco/trace.o -c ../../shared/trace.c
  CC	../../shared/trace_stdio.c
arm-none-eabi-gcc -Os -std=c99 -ggdb3 -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -fno-common -ffunction-sections -fdata-sections -Wextra -Wshadow -Wno-unused-variable -Wimplicit-function-declaration -Wredundant-decls -Wstrict-prototypes -Wmissing-prototypes -flto -MD -Wall -Wundef -I. -I../../shared -DSTM32L4 -DSTM32L476VGT6 -I../../libopencm3//include -o bin-stm32l4-disco/trace_stdio.o -c ../../shared/trace_stdio.c
  GENLNK  stm32l476vgt6
arm-none-eabi-gcc -E -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -DSTM32L4 -DSTM32L476VGT6 -D_ROM=1024K -D_RAM=96K -D_RAM2=32K -D_ROM_OFF=0x08000000 -D_RAM_OFF=0x20000000 -D_RAM2_OFF=0x10000000 -D_RAM3_OFF=0x20040000 -P -E ../../libopencm3//ld/linker.ld.S > generated.stm32l476vgt6.ld
  LD	adc-power-stm32l4-disco.elf
arm-none-eabi-gcc -Tgenerated.stm32l476vgt6.ld -L../../libopencm3//lib -nostartfiles -mcpu=cortex-m4 -mthumb -mfloat-abi=hard -mfpu=fpv4-sp-d16 -specs=nano.specs -u _printf_float -Wl,--gc-sections -flto bin-stm32l4-disco/main-stm32l4-disco.o bin-stm32l4-disco/adc-power.o bin-stm32l4-disco/trace.o bin-stm32l4-disco/trace_stdio.o -lopencm3_stm32l4 -specs=nosys.specs -o adc-power-stm32l4-disco.elf
/home/karlp/tools/gcc-arm-none-eabi-6-2017-q2-update/bin/../lib/gcc/arm-none-eabi/6.3.1/../../../../arm-none-eabi/lib/thumb/v7e-m/fpv4-sp/hard/libc_nano.a(lib_a-writer.o): In function `_write_r':
writer.c:(.text._write_r+0x10): undefined reference to `_write'
collect2: error: ld returned 1 exit status
../../rules.mk:137: recipe for target 'adc-power-stm32l4-disco.elf' failed
make[1]: *** [adc-power-stm32l4-disco.elf] Error 1
make[1]: Leaving directory '/home/karlp/src/libopencm3-tests/tests/adc-power'
Makefile:12: recipe for target 'Makefile.stm32l4-disco.all' failed
make: *** [Makefile.stm32l4-disco.all] Error 2
```

davidgfnet:
You are using some library (libc_nano.a) which implements write() function. Is that expected? In that case, I think you might have not build that library with LTO support, but not sure really.

davidgfnet:
BTW this patch is unnecessary if one does run make like this:
CFLAGS="-flto" LDFLAGS="-flto" make -j20 AR=arm-none-eabi-gcc-ar
Maybe we could just document it and not update any Makefiles, what do you think?

karlp:
if I can't use newlib-nano or newlib at all, it's a bit difficult to get excited about lto, yes :)  I'll close this, you can make a new patch to the readme if you like.  I'm certainly not rebuilding newlib, I want to be able to use the ones in the regular toolchains.

davidgfnet:
Gotcha. It's not that you cannot use them, is that ideally they should be built using LTO. Still don't get why this whole thing fails: lto is supposed to fallback to regular linking in case some symbols are not lto compatible.

karlp:
Note to try this again....

```
14:37 <via> okay, i think i know whats happening
14:37 <via> you have the _write implementation, right?
14:37 <karlp> yes.
14:37 <via> libc depends on that symbol
14:37 <via> but when compiling your project with lto it sees _write isn't used i'm guessing
14:37 <via> try adding __attribute__((externally_visible)) to it
```

via:
Also worth noting that at least on the toolchain I am using I have to build libopencm3 with:
<pre>
make CFLAGS="-flto" LDFLAGS="-flto" AR="arm-none-eabi-gcc-ar"
</pre>

And this might have changed but at least in my ancient toolchain you need -O3 to do whole program optimization and inlining, so I modified your toplevel Makefile to use -O3 instead of -Os.  The resulting binary looks extremely inlined of libopencm3 functions.

deece:
Autoloading the plugins only works if the gcc plugins have been installed in the location that binutils has compiled in, this is typically not the case if the user is using a non-distro, precomputed toolchain (eg. The ARM GCC toolchain).

Looking at the Makefiles, the value of AR is forced (it has an =, rather than ?=), so we would still need a patch to allow AR to be overridden (and probably centralised).

deece:
The problem with wrapped functions (such as _write) is that the linker does not see any references to the wrapped functions, since it isn't explicitly called, and eliminates it before the compiler runs. I had a similar problem in Mbed with a wrapped main: https://github.com/InfernoEmbedded/mbed-os/commit/cb406be586bca58f1dbe4f05d847ecc14fd3cd6c

The only solutions I am aware of are to either remove the wrapping (if it doesn't value-added), or declare the wrapped function with the used attribute (which is doable, since it's code local to libopencm3), since that makes it clear to the linker that there are references to the function that the linker connect see. I think you would have the same problem calling a C function from assembler when using -ffunction-sections and --gc-sections.

The problem is not visible without LTO, since at link time, the compiler has already expanded the wrapped functions, so a real reference exists.

deece:
Hmm, just a thought, in an LTO build, we could probably get away without -ffunction-sections & --gc-sections, which would keep the linker from prematurely eliminating the symbols. Unused symbols would then be eliminated during the final compile phase.

howard0su:
> Hmm, just a thought, in an LTO build, we could probably get away without -ffunction-sections & --gc-sections, which would keep the linker from prematurely eliminating the symbols. Unused symbols would then be eliminated during the final compile phase.

remove them causes negative impact to the final size. please keep those.

howard0su:
we use LTO for libopencm3 successfully when optimizing a bootloader for STM32. This is pretty useful patch. Can we merge this and give the user option to try LTO when needed?

deece:
> > Hmm, just a thought, in an LTO build, we could probably get away without -ffunction-sections & --gc-sections, which would keep the linker from prematurely eliminating the symbols. Unused symbols would then be eliminated during the final compile phase.
> 
> remove them causes negative impact to the final size. please keep those.

@howard0su Is this confirmed? LTO should be able to remove everything the linker can, if it doesn't, its worth reporting upstream as a missed optimisation.

howard0su:
> > > Hmm, just a thought, in an LTO build, we could probably get away without -ffunction-sections & --gc-sections, which would keep the linker from prematurely eliminating the symbols. Unused symbols would then be eliminated during the final compile phase.
> > 
> > 
> > remove them causes negative impact to the final size. please keep those.
> 
> @howard0su Is this confirmed? LTO should be able to remove everything the linker can, if it doesn't, its worth reporting upstream as a missed optimisation.

This is confirmed at least for the project I am working on. The following result is what I tested:
-fdata-sections -ffunction-sections &  -Wl,-gc-sections:
ROM: 0x08004000 - 0x080399b4 = 214.43kB
RAM: 0x20000000 - 0x20003f90 =  15.89kB

-lto + -fdata-sections -ffunction-sections &  -Wl,-gc-sections
ROM: 0x08004000 - 0x08037f5c = 207.84kB
RAM: 0x20000000 - 0x20003f58 =  15.84kB

-lto (no sections)
ROM: 0x08004000 - 0x0803a25c = 216.59kB
RAM: 0x20000000 - 0x20003fa8 =  15.91kB

I didn't run the test with LTO or non-LTO against libopencm3. But in general, leave those flags are actual good thing.

deece:
Could you please try with -fdata-sections & -gc-sections (but not -ffunction-sections)? The problem with linker elimination of functions is that it runs before wrapped functions are evaluated, so it incorrectly removes functions since there are no references to them, resulting in link failures. I think this is the problem that @karlp is experiencing above.

howard0su:
@deece you are right. remove function sections gives 0.4KB gain.
No-LTO with data-sections and function-sections
ROM: 0x08004000 - 0x0803935c = 212.84kB
RAM: 0x20000000 - 0x20003ff0 =  15.98kB

LTO with data-sections and function-sections
ROM: 0x08004000 - 0x08037b90 = 206.89kB
RAM: 0x20000000 - 0x20003fc8 =  15.95kB

LTO - only data-sections without function-sections
ROM: 0x08004000 - 0x08037a50 = 206.58kB
RAM: 0x20000000 - 0x20003fc8 =  15.95kB

LTO - neither data-sections nor function-sections
ROM: 0x08004000 - 0x08038358 = 208.84kB
RAM: 0x20000000 - 0x20003ff4 =  15.99kB

#arm-none-eabi-gcc --version
arm-none-eabi-gcc (GNU Tools for Arm Embedded Processors 8-2018-q4-major) 8.2.1 20181213 (release) [gcc-8-branch revision 267074]


And I notice there are some makefile is missed in this change and they are still using normal ar.

deece:
Great, thanks for confirming. This should give us the best outcome, while also fixing the link issues @karlp saw.

howard0su:
Looking forward see LTO is officially supported.

karlp:
so, has anything changed as far as building my projects?  I don't see anything new here other than some off topic optimization dicsussions?  But I may be missing soething in the discussion.

deece:
@karlp drop -ffunction-sections from your CFLAGS, it is responsible for the premature elimination of your write implementation.

ChuckM:
Yes, the take is that if you remove `-ffunction-sections` from CFLAGS it does not change the optimization of the code that doesn't use LTO (so it has no net negative effect) and without it you can safely use LTO to get those optimizations without causing the missing symbol failures. So net, net, it is the correct fix here.

PR #857, opened 2017-11-23
Added hsi clock setup for STM32F4.

swilde:
I needed this for a NUCLEO F401RE board, but it should be working with all F4 MCUs (as long as you select frequencies within the spec).


karlp:
Would it not be simpler to just add a rcc_clock_setup_pll() function like l0/l1/f3 has?  I can't say I'm a fan of trying to use the hse definitions as hsi definitions, I don't think anyone would ever find them like that.

ChuckM:
Apropos of nothing the F401RE nucleo board routes the MCLOCK out from the STLINK side over to the '401 so you can use the HSE option with an 8Mhz external clock.

karlp:
Rendered obsolete by https://github.com/libopencm3/libopencm3/pull/1032 which implemented the unified API

PR #860, opened 2017-12-03
STM32F3: add function to setup 72mhz clock from 8mhz bypass source

markus-k:
This patch adds an utility function for the STM32F3 line to set up a 72MHz system clock from a 8MHz HSE bypass source. This is useful for quickly getting the STM32F3-Discovery board set to it's maximum clock speed and having the USB peripheral run at 48MHz.


karlp:
yeah no.  We're not doing any more fixed clock function stuff, there's a reason you haven't seen more of them.  You can make a function that takes a clock config structure like the rest, but we're not going back down that path we went with F1 in the beginning. Sorry.

PR #861, opened 2017-12-04
stm32f7: usart: enable usart peripheral

Lampoo:
Add usart-v2 to stm32f7 to provide usart support in f7 series.


karlp:
merged in 580a2a4 thanks!

PR #868, opened 2018-01-04
doc: clarify that usbd_register_control_callback() must be called from config callback

WGH-:


WGH-:
This problem was seen before e.g. here https://github.com/libopencm3/libopencm3-examples/commit/393fe8e449d5334c715a3ca5f538328dbf387182

karlp:
No, it's not must, it's _after_

The easiest way of doing that is inside the callback, but it doesn't have to be done there.

WGH-:
Sound reasonable. Is changing "must" to "should" with clarification "or after" sounds good enough?

karlp:
just change the "inside" to "inside or after" and we should be good to go.  I'll just do that now and merge it.

karlp:
merged with some more rewording in f59d47c

PR #871, opened 2018-01-11
stm32: rcc: change gpio bit defines to be consistent with reference manual

Lampoo:
Found these "incompatible" gpio bit defines in RCC when switch to stm32f7 library from stm32f4. Checked the reference manual for f2/f4/f7 and the new h7, they all use GPIO prefix instead of IOP.

Not sure if this is the right way to go as this breaks compatibility and leave it to the app to maintain the difference, alternative would be either change the f7 defines back to use the "IOP" or create alias in it.


karlp:
Duplicate but extends #786 
same concerns as there.  Will break existing code, but is "right" and better for the future

Lampoo:
@karlp Thanks for the information, shall I close this duplicated PR or keep it here for reference?

karlp:
you can keep it really, this one addresses f2 as well, which the other one doesn't.

As I mentioned on irc, if we can get a few other breaking things done together, I'll bring this in with them all.

Lampoo:
Cool, thanks!

karlp:
I've jsut merged this, but followed it up with a commit that adds them back in with the old names marked as deprecated.  Seems like a polite way forward. thanks for this!

see https://github.com/libopencm3/libopencm3/commit/eeef996cb0b524b329c47019adc9f451f69a8d4d and https://github.com/libopencm3/libopencm3/commit/0965e691a948640846d2721f8c7faa9e88d09732

PR #872, opened 2018-01-15
stm32: make reset_handler reusable in app

Lampoo:
If app wishes to have its own reset_handler, e.g. do clock initialzation
or text relocation, it'll have to override the default one (declared as weak)
at link time, rewrite everything in reset_handler and can't reuse what's
already in libopencm3. This patch breaks reset_handler into 2: reset_handler
remains as weak and can be overriden by app, and cm3_reset_handler which
provides a generic reset handling and make it available to the app.

### DIFF
--- a/include/libopencm3/cm3/nvic.h
+++ b/include/libopencm3/cm3/nvic.h
@@ -157,6 +157,8 @@ void sv_call_handler(void);
 void pend_sv_handler(void);
 void sys_tick_handler(void);
 
+void cm3_reset_handler(void);
+
 /* Those defined only on ARMv7 and above */
 #if defined(__ARM_ARCH_7M__) || defined(__ARM_ARCH_7EM__)
 void mem_manage_handler(void);
--- a/lib/cm3/vector.c
+++ b/lib/cm3/vector.c
@@ -61,6 +61,11 @@ vector_table_t vector_table = {
 
 void __attribute__ ((weak, naked)) reset_handler(void)
 {
+	cm3_reset_handler();
+}
+
+void cm3_reset_handler(void)
+{
 	volatile unsigned *src, *dest;
 	funcp_t *fp;
 


karlp:
Looks reasonable to me.

I'm curious what you're doing that you both need this, and yet still want to use the existing code is though.

Lampoo:
What I intends to do is to have a minimal bootloader that can relocate itself from flash and runs at itcm, with this minimal bootloader and set things up, we can update and program full flash contents, and/or  load critical / common libraries of bigger size into itcm at the maximum speed for RTOS uses. I thought a simple way was to override the "reset_handler" and put the relocate logic in there, but now I realize the vector table contains the "address" of the reset handler, which means if link the bootloader at itcm address would break things. The vector table MUST have a valid address that resides in itcm-flash / axim-flash address space that instructions can be fetched correctly. To make things work, I will have to use macro in the vector table to do the address translation. Any suggestion on this?

Lampoo:
Looks like I can define the reset_handler  in the linker script like this:

```
@@ -37,7 +37,8 @@
 MEMORY
 {
-       rom (rx)  : ORIGIN = 0x00200000, LENGTH = 16K
+       rom (rx)  : ORIGIN = 0x00000000, LENGTH = 16K
+       flash (rx): ORIGIN = 0x00200000, LENGTH = 16K
        ram (rwx) : ORIGIN = 0x20000000, LENGTH = 128K
 }

@@ -49,8 +50,12 @@ SECTIONS
 {
         . = ORIGIN(rom);

+        _srom = .;
+
         .text : {
-                *(.vectors)     /* Vector table */
+                KEEP(*(.vectors))     /* Vector table */
+               reset_handler = . + ORIGIN(flash);
+                KEEP(*(.text.boot))     /* Reset handler */
                 *(.text*)       /* Program code */
                 . = ALIGN(4);
                 *(.rodata*)     /* Read-only data */
@@ -81,6 +86,8 @@ SECTIONS
                __fini_array_end = .;
        } >rom

+        _erom = .;
+
```

This will give a valid reset vector address in the vector table.

karlp:
So, do you still actually need library code changed/added?

eqvinox:
I think doing this through linker magic is way better than having a function call in the reset handler. Additionally, if you're tinkering with things on this low level, copying the entire `reset_handler()` to your application and thus overriding the `weak` libopencm3 version is probably a good idea.

=> I'd close this.

karlp:
no updates in a month, closing

PR #875, opened 2018-02-06
stm32f0: Removed duplicate uart_{enable,disable}_{rx,tx}_dma functions

jhol:
These are now defined in `lib/stm32/common/usart_common_all.c`


karlp:
this has now merged thanks

PR #879, opened 2018-02-15
Makefile: allow specific targets to be built

fenugrec:
If "TARGETS" is defined, use that; otherwise defaults to all targets.
Example : "make TARGETS=stm32/f0"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,14 +27,17 @@ space:=
 space+=
 SRCLIBDIR:= $(subst $(space),\$(space),$(realpath lib))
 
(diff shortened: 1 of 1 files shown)


karlp:
see also https://github.com/libopencm3/libopencm3/pull/559 and the comments from flixr

fenugrec:
ah, I had skipped a section in the make docs :

> If a variable has been set with a command argument (*note Overriding
> Variables: Overriding.), then ordinary assignments in the makefile are
> ignored.

So, environment variables are clobbered by ":=" and "+=" , but command-line variables are not.

PR #880, opened 2018-02-16
stm32 USB : gadget0 test + fix for packet memory corruption

fenugrec:
finally... Original issue : #476 

This PR contains
- a new gadget0 test that fails (expected since current code is broken)
- the correct implementation for st_usbfs_set_ep_rx_bufsize(), which allows the new gadget0 test to pass. 

This commit is identical to the one in my original PR #477 . Merging this current PR means 477 can be closed; I wanted to leave the old one open because its comments are still relevant.

**********
Here's how the gadget0 test works :

1. Set the correct configuration. I had to add a new once since I wanted different a different OUT callback
1. Host sends a bulk transfer to the OUT EP. The buffer for this EP is immediately after the EP0 area in the packet memory buffer.
1. The device does nothing immediately with the data, so the OUT EP stays in NAK
1. Host sends a control transfer with bMaxPacketSize0 of data.
1. One would think that the initial OUT EP buffer would still be intact at this point...
1. Device, in response to the control transfer, reads the initial OUT packet, and calls write_packet() to echo the initial bulk transfer
1. Host does a bulk IN transfer, and compares the received data with what was sent.

**********
I hope the explanation isn't too horrible. Needless to say, on F072 (the only target I can test) this fails until I apply the fix.


karlp:
If you've got a moment, can you rebase and push -f this?  We got travis turned on and you created this before the travis.yml file got created, so it doesn't know how to build your branch

fenugrec:
Bonus : merging this will also close #537

karlp:
Just so you don't think I've forgotten this, I've made a (much) simpler test case.  I've just got a second endpoint, which collides easily.  Below trace prints '0' for one endpoint and 'o' for the other, first section is with 64byte ep assignment (known to be incorrect) and second is with 32byte assignment (known to be correct)

I've still got to tidy this up a bit, but it's much easier to see now, with a lot less device code and a lot less reliance on buffers staying in place.

```
oOOodata corruption on 0x3
Oodata corruption on 0x3
OoOoOoOoOoOoOoOoOOodata corruption on 0x3
OOodata corruption on 0x3
OoOoOoOOodata corruption on 0x3
OOodata corruption on 0x3
OoOoOoOOodata corruption on 0x3
OoOoOoOoOoOOodata corruption on 0x3
OOodata corruption on 0x3
OoOoOoOOodata corruption on 0x3
OOodata corruption on 0x3
OoOoOoOoOoOodata corruption on 0x3
Ooooooooooobootup complete
set cfg 2
I

^^ lots of corruption of second endpoint by first endpoint trampling it's buffer.

OOOoOoOOoOOOOOoOoOoOoOOoOoOoOoOoOoOoOoOoOOoOOoOOOOoOoOoOOOOOoOOOOoOOoOoOoOoOoOOOOOOoOoOOoOOOoOoOoOoOoOOOOoOooOoOoOOOOOoOoOOoOoOOoOoOoOoOoOoOoOoOoOOOOOoOoOOOoooooooooooooooooooooooooooooooooooooooooooo
```

PR #856, opened 2017-11-21
Rewrite ld script generator to Python

noonien:

PR #882, opened 2018-02-20
Make it possible to specify custom python interpreter

plaes:
Newer distributions that are coming out are python3 out of the box.

Signed-off-by: Priit Laes <plaes@plaes.org>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,8 @@
 PREFIX		?= arm-none-eabi
 #PREFIX		?= arm-elf
 
+PYTHON		?= python
+
 STYLECHECK      := scripts/checkpatch.pl
 STYLECHECKFLAGS := --no-tree -f --terse --mailback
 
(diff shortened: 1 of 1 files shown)


karlp:
No. I'd rather make the script python2/python3 compatible than add yet more options for people.

karlp:
in the interim, could you explain a little further?  is it that "python" is now python3? would just making it an explicit "python2" invocation help?

plaes:
It happened with Ubuntu 17.x (helped troubleshoot issues by one of my students) who struggled on getting it properly working. As far as I know, he had following issues:
* only python3 installed
* missing awk (Ubuntu seems to hijack those missing commands with certain "please-run-foo-to-install" feature, which also confused make)

I was planning to try it out with clean install vm, but haven't found time yet. Also, there seems to be an issue when using python3.

karlp:
the awk is only for the optional linker script generation, and that's explicitly in the readme.  You're welcome to test the alternate proposed linker script generator in https://github.com/libopencm3/libopencm3/pull/856 and I'll try and make sure at least the basic python usage works.  I'd rather it "just works" rather than trying to allow/require people to specify it.

karlp:
added a request for info a month ago, no follow up. closing.

plaes:
Sorry for keeping you hanging, unfortunately the student with this "funny" setup disappeared... :(

PR #883, opened 2018-02-22
Add RAM3 support for L4-Plus to linker generation

ChristianTacke:
This only adds basic support.
Specific settings still need to be added.


karlp:
sure, looks perfectly reasonable.  merged in 914521d thanks

PR #884, opened 2018-02-22
Let the created ld script depend on devices.data

ChristianTacke:
If devices.data changes, rebuild the linker scripts, so they have all the
new information in it.


karlp:
thanks! seems perfectly reasonable

ChristianTacke:
Thanks for the fast merge!

PR #885, opened 2018-02-24
For merge 2018 mk3

karlp:
I've got more lined up too, but let's get them coming out in chunks.

Yes, the usb_dwc changes will "break" some existing PRs, but only in very easy to fix ways, and that's my problem.


lib/stm32/l0/rcc.c
```diff
@@ -336,6 +336,19 @@ void rcc_set_pll_divider(uint32_t factor)
...
+ * Set the pll source.
+ * @param pllsrc RCC_CFGR_PLLSRC_HSI16_CLK or RCC_CFGR_PLLSRC_HSE_CLK
+ */
+void rcc_set_pll_source(uint32_t pllsrc)
+{
+	uint32_t reg32;
+
+	reg32 = RCC_CFGR;
+	reg32 &= ~(RCC_CFGR_PLLSRC_HSE_CLK << 16);
+	RCC_CFGR = (reg32 | (pllsrc<<16));
```
ChristianTacke: Could you consider replacing the `16` by something like `RCC_CFGR_PLLSRC_SHIFT`?

karlp: I could, and I largely left it there to see if anyone was even reading :)  It's a copy of the existing L1 code, which would then also need to be done, and we don't tend to define single bits.  It's only a single bit on L0/L1, but a range on others, which is why the bit is defined unshifted here, vs shifted, as most single bits are.  joy and confusion

ChuckM:
Merge it already! 👍 We have been waiting a long time and I've got so many things that I could switch to HEAD once you do this.

lib/usb/usb_standard.c
```diff
@@ -129,8 +129,9 @@ static uint16_t build_config_descriptor(usbd_device *usbd_dev,
 		}
 	}
 
-	/* Fill in wTotalLength. */
-	*(uint16_t *)(tmpbuf + 2) = totallen;
+	/* Fill in wTotalLength.
+	 * Note that tmpbuf is sometimes not halfword-aligned */
+	memcpy((tmpbuf + 2), &totallen, sizeof(uint16_t));
```
fenugrec: here again, calling memcpy for 2 bytes seems like total overkill. It will work, but....

ChristianTacke: Often, gcc is able to optimize memcpy if the size is known. For example compiling on STM32L4:
```
void copy_uint16(void *a, const void *b) {
    memcpy(a, b, sizeof(uint16_t));
}
   0:   880b            ldrh    r3, [r1, #0]
   2:   8003            strh    r3, [r0, #0]
   4:   4770            bx      lr
```

fenugrec: Interesting, I seem to recall testing something similar a while ago and gcc didn't do that optimization.

karlp:
landed!

PR #886, opened 2018-02-26
L4: Fix USART3 definition/typo

ChristianTacke:
USART*3* should point to *3* not *2*.


karlp:
fixed in a1264f50 thanks

PR #887, opened 2018-02-26
genlink: Add LIBDEPS for libopencm3 itself

ChristianTacke:
If $(OPENCM3_DIR)/lib/libopencm3_*.a exists, it will be linked in.
If we do that, we should also add it to the deps.
That way a newer *.a will result in a relink.


karlp:
merged via rebase in the linked commits above, forgot to close the ticket

PR #888, opened 2018-03-04
STM32L4: USART: Fix baudrate calculation for LPUART1

ChristianTacke:
LPUART1 needs BRR to be 256 times the normal value.

To calculate this correctly with rounding, the best answer
is to use 64 bit values: A clock of say 20 MHz times 256 is
already 5e9 MHz, which doesn't fit in 32 bits.

There are probably some tricks using GCD, fixed point math
or something, but really, a little 64 bit math wont hurt.
This is setup code, which usually is not performance
critical.

### DIFF
--- a/lib/stm32/common/usart_common_all.c
+++ b/lib/stm32/common/usart_common_all.c
@@ -41,6 +41,9 @@ USART1/6) or the APB low-speed prescaler clock (for other USARTs). These values
 must be correctly set before calling this function (refer to the
 rcc_clock_setup-* functions in RCC).
 
+Note: For LPUART, baudrates over 2**24 (~16.7 Mbaud) may overflow
+the calculation and are therefore not supported by this function.
+
 @param[in] usart unsigned 32 bit. USART block register address base @ref
 usart_reg_base
 @param[in] baud unsigned 32 bit. Baud rate specified in Hz.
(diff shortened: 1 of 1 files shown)


karlp:
Applies to L0 as well, not _realllly_ sure about the implementation though :)

ChristianTacke:
I can rephrase the commit to do 's/STM32L4/STM32L/'?

Should I explain something about the implementation?

zyp:
I saw some discussion on IRC regarding avoiding to have to pull in library functions to handle the 64-bit division. I puzzled around with it a bit and concluded that we can avoid the 64-bit division by doing two 32-bit divisions and a modulo operation.

The concept is simple; first we do a plain division between clock and baud, before multiplying with 256. We then take the modulo between clock and baud, multiply it by 256 and divide it by baud again. That gives us full precision, and since the modulo output is bounded by the baudrate, it can't overflow as long as the baudrate is smaller than 2**24, i.e. a bit over 16.7 Mbaud, which is probably more than any LPUART is capable of.

The simple implementation (flooring) looks like this:

    USART_BRR(usart) = clock / baud * 256 + clock % baud * 256 / baud

To get rounding like the old implementation, we add 1/2 to the second term:

    USART_BRR(usart) = clock / baud * 256 + (clock % baud * 256 + baud / 2) / baud

ChristianTacke:
Updated to include the 32 bit variant and comments/notes by zyp.
I left the original 64 bit version in the commit history for "historic" reasons, so that people can look at it.

karlp:
somehow this PR didn't get closed, even though it's code has landed now.

PR #889, opened 2018-03-06
Changes for 469

ChuckM:
As I ported my STM32F469 examples to the latest head of LOC3 I found a few things that were still missing, they have been collected into this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,3 +9,9 @@ addons:
     - sourceline: 'ppa:team-gcc-arm-embedded/ppa'
     packages:
     - gcc-arm-embedded
+
+notifications:
+  irc:
+    channels:
+      - "chat.freenode.net#libopencm3"
+    use_notice: true
--- a/include/libopencm3/stm32/f4/ltdc.h
+++ b/include/libopencm3/stm32/f4/ltdc.h
@@ -153,6 +153,14 @@
 #define LTDC_GCR_HSPOL_ACTIVE_LOW  (0<<31)
 #define LTDC_GCR_HSPOL_ACTIVE_HIGH (1<<31)
 
+/* GCR - register bit defines (no semantics) */
+#define	LTDC_GCR_HSPOL			(1 << 31)
+#define LTDC_GCR_VSPOL			(1 << 30)
+#define LTDC_GCR_DEPOL			(1 << 29)
+#define LTDC_GCR_PCPOL			(1 << 28)
+#define LTDC_GCR_DITHER			(1 << 16)
+#define LTDC_GCR_LTDCEN			(1 << 0)
+
 /* --- LTDC_SRCR values ---------------------------------------------------- */
 
 /* Vertical Blanking Reload */


karlp:
for the ltdc one, why would you care whether the bit was high or low?  the whole point of the defines was so that you _didn't_ have to worry about what the values were?

include/libopencm3/stm32/common/flash_common_f234.h
```diff
@@ -60,6 +60,12 @@
 #define FLASH_ACR_LATENCY_6WS		0x06
 #define FLASH_ACR_LATENCY_7WS		0x07
 
+#define FLASH_ACR_DCRST			(1 << 12)	/* Data cache reset */
+#define FLASH_ACR_ICRST			(1 << 11)	/* Instruction cache reset */
+#define FLASH_ACR_DCEN			(1 << 10)	/* Data Cache enable */
+#define FLASH_ACR_ICEN			(1 << 9)	/* Instruction Cache enable */
+#define FLASH_ACR_PRFTEN		(1 << 8)	/* Prefetch enable */
```
karlp: a) these should be added to stm32/f4/flash.h if they are f4 specifc, and more importantly...
b) these all exist in https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/stm32/common/flash_common_f24.h#L52-L56 already :)  Are you sure you have an updated library locally? you should be getting errors about redefining things.

ChuckM: Hmm, if you look at the master branch of the repo where I generate this PR from it shows it as up to date with the main repo. I did not pull it against a staging repo if that was the question.

ChuckM:
On the LTDC one I initially thought the register bits weren't defined because I used the standard form in my code and they were undefined. Then looking at the include file I saw there were two defines for each bit one _HIGH and one _LOW. I added the bit defines without the suffixes.

include/libopencm3/stm32/common/flash_common_f234.h
ChuckM: So just rebased and remerged this with the latest and I can see the duplicates.

karlp:
I don't really want to have multiple versions of this ltdc stuff, that seems really unhelpful.  should we just drop the others? (they seem kinda better to me, but I've not used any of it)  you've rebased, but you're still adding the duplicates, did you mean to do taht?

ChuckM:
I was looking at the flash stuff too. And having rebased and then pulled all of the commits (even the bogus extra ones) and leaving the LTDC commit it still builds without error, so there was some brain fart going on about needing the flash defines.

On the LTDC defines it is your call for one form (and which of the two)  or both forms. The STM libraries typically defined things with semantic names which seems like what Oliver copied when he created them. But all of the other defines in the library are 'bit names'. 

So as I see it the argument for keeping only the existing ones: that is the way they were and you can read the include file to figure out how to use them as bit specifiers.

The argument for keeping both: New users to libopencm3 will see the defines which are like the other bit defines, and they may use them without even consulting the source file. Existing users won't have to change their code.

The argument for keeping the new defines only: The library becomes more consistent overall and it continues to emphasize a consist methodology for defining bits in registers, that leading to less surprise when a bitfield is defined in an unusual way.

Your call. We don't need the flash commit feel free to cherry pick the LTDC commit if you decide to go with it.

ChuckM:
Ok then, took out the FLASH changes (they were superfluous) and left just the LTDC change.

karlp:
pulled locally, will close when it lands

karlp:
landed in 1e9a2e641cbf1f8d50870f3b30fa7c97a5eb97cc

PR #890, opened 2018-03-07
Fix some clang warnings

keepkeyjon:


karlp:
heh, I'd actually been doing a bit of this locally as well, had both the first two already.  I'd not managed to get the third one in a way that actually worked.

Could you elaborate on how you're testing this, and what your build environment is please?

keepkeyjon:
Our build environment at keepkey is currently gcc7+newlib. These patches work in that environment, and pass our device tests.

Independently of that, I've been working on adding baremetal-arm support to clang. There is still quite a bit of toolchain work before that can really be declared "done" and production-ready in an upstream release of clang, but in the meantime, sorting out small issues like this helps to be able to test that toolchain on something beefier than a "hello world" scale project.

> I'd not managed to get the third one in a way that actually worked.

What symptoms did you observe?

karlp:
merged in f2589bc7, f53e12d2 and 8071c6cf thanks

PR #892, opened 2018-03-14
ld scripts: drop duplication of standard sections

karlp:
Instead of every "simple" target having their own duplicate file with
all the section mappings, just provide a single, simple,
"cortex-m-generic.ld" that works with our startup code and any simple
rom/ram system.  This also drops the pointless copying of files all over
the place.  Using -L flags properly is sufficient, and the standard file
is now in the root of the library already.


karlp:
This has landed via 9a05dcb6

PR #893, opened 2018-03-17
STM32F0: Add RCC API for I2C1 clock source

mattande:
* Providing API to set/clear RCC_CFGR3_I2C1SW on STM32F0, duplicated
	from STM32F3 applies only to I2C1.


karlp:
We're goign to want to think about an API for this in future across all families, similar functionality exists on l0 and l4, but this is at least self consistent.

merged in 0d5e51a8 thanks

PR #894, opened 2018-03-17
STM32: I2Cv2 handle NAK in i2c_transfer7()

mattande:
Attached commit handles the off-nominal conditions in i2c_common_v2.c i2c_transfer7(), where if a I2C device responds with NAK, or does not respond to the address phase, the condition is caught and returned to the user in a status result.
Tested on STM32F030C6.

### DIFF
--- a/include/libopencm3/stm32/common/i2c_common_v2.h
+++ b/include/libopencm3/stm32/common/i2c_common_v2.h
@@ -389,6 +389,7 @@ void i2c_peripheral_disable(uint32_t i2c);
 void i2c_send_start(uint32_t i2c);
 void i2c_send_stop(uint32_t i2c);
 void i2c_clear_stop(uint32_t i2c);
+void i2c_clear_nak(uint32_t i2c);
 void i2c_set_own_7bit_slave_address(uint32_t i2c, uint8_t slave);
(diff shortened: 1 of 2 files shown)


karlp:
Do you  have a test case for this we can use?  Or some logic analyser before/after traces?

mattande:
I don't have any test cases to submit. 
The previous implementation was rather incomplete and has two clear issues:
* Without clearing NACKF, if a transfer to an address which does not exist on the I2C bus, or if the slave responds with NACK for any reason (end of block transfer for example) all future transfers will fail util NACKF is cleared.
* NACKF is was handled with /*FIXME*/ while (1); tight loop on the write case, and unhaneled on the read case if the I2C device does not respond to the slave address query with indefinite wait for RXNE.

Most users will want to write their own interrupt based I2C handler, though for a simple use case it is helpful to have this utility function. It could be frustrating for some out of the box without handling NACK.

gkeeth:
Any update on this? I haven't tried this code, but it looks correct.

The existing i2c_transfer7 function seems pretty broken in the case of a NACK. The RM is clear that if a NACK is received after an attempted transfer, the transfer should end immediately. The existing function instead will instead loop forever, which seems not great.

PR #895, opened 2018-03-18
add STM32L0xx4 w/ 16K flash

jcw:
I've added a load script for the smaller STM32L011 series with 16 KB flash.
This is also a test to find out how you want to get new pull requests, and whether it can be accepted.

My reason for doing this is to then try and improve the µC support a bit further in PlatformIO.
It's rather limited right now, see https://platformio.org/boards?filter%5Bframeworks%5D=libopencm3

### DIFF
--- a/lib/stm32/l0/stm32l0xx4.ld
+++ b/lib/stm32/l0/stm32l0xx4.ld
@@ -1,0 +1,32 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2012 Karl Palsson <karlp@tweak.net.au>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


karlp:
This is covered by the linker script generator: https://github.com/libopencm3/libopencm3/blob/master/ld/devices.data#L151

Given the vast number of stm32 targets available, trying to maintain linker scripts for every one of them is simply untenable.  The existing ones are allowed to stay for now, but we're not adding more of them

jcw:
Ok. Are you interested in getting more support for libopencm3 in PlatformIO?

karlp:
sure! but let's discuss that either in it's own issue, or come by irc.freenode.org#libopencm3

scttnlsn:
@jcw Do you have any examples of libopencm3 code for the STM32L0XX4?  I've been having a hard time getting blinky running on an ST32L011F4 w/ libopencm3.

jcw:
As of recently, I can point you to Karl's [miniblink](https://github.com/karlp/libopencm3-miniblink) - if you do a make, you'll see a binary for the L011 nucleo, *but...* it needs to have the ram size set to 2K max - see https://github.com/karlp/libopencm3-miniblink/issues/5 for details.

scttnlsn:
Thanks!  I wasn't able to get it working but I also had to tweak the GPIO port/pin since I'm using the TSSOP-20 package with no PB3 exposed.  I'll follow along with that other issue and perhaps I can puzzle through it.

PR #897, opened 2018-03-20
crypto: fix iv key not set correctly

Lampoo:
Found a mismatched declaration: CRYP_IVR() is defined as MMIO32 however crypto_set_iv use a uint64_t as parameter. This cause the iv key not set into the register correctly.

With the patches, I have verified to run AES CBC 128bit decryption on a STM32F756IGT chip.


karlp:
For the first one, what's better from the user point of view? two entry 64bit array, or a 4 entry 32bit array?  (ie, should the fix be what you've done here, or to move the param to be an uint32 array?)

Lampoo:
To be consistent with the API of crypto_set_key(enum crypto_keysize keysize, uin64_t key[]), I chose crypto_set_iv(uint64_t iv[]) in my code. However with a uint64_t type, I will need to do extra work before calling the API, take a AES CBC 128-bit decryption for example:

```
typedef union {
        uint8_t  u8[32];
        uint8_t  u16[16];
        uint32_t u32[8];
        uint64_t u64[4];
} CRYPTO_U32;

typedef union {
        uint8_t  u8[16];
        uint8_t  u16[8];
        uint32_t u32[4];
        uint64_t u64[2];
} CRYPTO_U16;

AES128_cbc_decrypt(const unsigned char *in, unsigned char *out, int len, const unsigned char key[16], unsigned char ivec[16])
{
       CRYPTO_U32 ukey;
       CRYPTO_U16 uivec;

        addr = (const uint32_t*) key;
        ukey.u32[0] = 0;                        /* K0LR, not used */
        ukey.u32[1] = 0;                        /* K0LR, not used */
        ukey.u32[2] = 0;                        /* K1LR, not used */
        ukey.u32[3] = 0;                        /* K1LR, not used */
        ukey.u32[4] = __bswap32(*addr++);       /* K2LR */
        ukey.u32[5] = __bswap32(*addr++);       /* K2RR */
        ukey.u32[6] = __bswap32(*addr++);       /* K3LR */
        ukey.u32[7] = __bswap32(*addr++);       /* K4RR */

        addr = (const uint32_t*) ivec;
        uivec.u32[0] = __bswap32(*addr++);      /* IV0LR */
        uivec.u32[1] = __bswap32(*addr++);      /* IV0RR */
        uivec.u32[2] = __bswap32(*addr++);      /* IV1LR */
        uivec.u32[3] = __bswap32(*addr++);      /* IV1RR */

        .....
        crypto_set_key(CRYPTO_KEY_128BIT, ukey.u64);
        crypto_set_iv(ivec.u64);
        ......
}
```

And it took me quite some time to read through the document to understand what's the right value that should be passed to the API. In this way, I would tend to think the APIs are not easy to use or user-friendly enough.

What my expectation is to have API like crypto_set_key(enum crypto_keysize key_size, unsigned char keys[]), the key_size tells the length of array keys and will handle what's necessary to set the register, e.g. reverse byte order, registers that to be written.

What do you think?

karlp:
I'm totally ok with changing this api, it had just a single submission in the past.  Make up a clear proposal on what the apis should really be for user friendly operation and we can look at getting them in instead

Lampoo:
@karlp There it goes. Kindly review and give your comments.

lib/stm32/common/crypto_common_f24.c
```diff
@@ -36,6 +36,10 @@
 
 #define CRYP_CR_ALGOMODE_MASK	((1 << 19) | CRYP_CR_ALGOMODE)
 
+#ifndef __REV
+#define __REV(x) __builtin_bswap32(x)
+#endif
```
karlp: out of scope of this commit! please make this separate, if it's even necessary. (it shuldn't be)

karlp:
Your example is missing the example data, how it's actually stored, why it's easier this way, what tool format can be natively pasted into code for instance

karlp:
sorry, pushed close instead of comment, very accidental

lib/stm32/common/crypto_common_f24.c
Lampoo: This is needed for the the crypto to work correctly, I had tried to grep the libopencm3 and found the exiting defines at 
```
./include/libopencmsis/core_cm3.h:62:#define __REV(x) __builtin_bswap32(x)
```
But see nowhere use this headers. what is the best place to put this define?

Lampoo:
@karlp Updated the description and added some comments, can you take a look again?

Lampoo:
@karlp Verified with AES tests from NIST document SP800-38A, at https://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38a.pdf, for AES 128/192/256 CBC/CTR/ECB.

Lampoo:
@karlp Is this good? Anything else I need to change / modify? And about the test codes, I can submit a PR later to libopencm3-example if you think this is good.

karlp:
I havent looked at your latest stuff yet, I've been doing any "real" libopencm3 work recently, just braindead stuff snuck in here and there.

karlp:
duplicate, with very different impl of https://github.com/libopencm3/libopencm3/pull/794

PR #866, opened 2017-12-22
STM32L4: EXTI, CRC, SPI, DMA, IWDG, RTC, CRS, HSI48, USB

br101:
Adding more support for L4, some parts taken from https://github.com/libopencm3/libopencm3/pull/854
All tested on STM32L433CC.

PR #898, opened 2018-03-23
STM32L4: RCC, PWR, Flash, IRQ updates

br101:
These are the remaining commits from https://github.com/libopencm3/libopencm3/pull/866 plus a few more to make L4 support complete for what I needed.


lib/stm32/l4/rcc.c
```diff
@@ -369,4 +454,136 @@ void rcc_set_msi_range_standby(uint32_t msi_range)
...
+ *
+ * @param clksel One of the definitions above
+ */
+void rcc_set_clock48_source(uint32_t clksel)
+{
+	RCC_CCIPR &= ~(RCC_CCIPR_CLK48SEL_MASK << RCC_CCIPR_CLK48SEL_SHIFT);
+	RCC_CCIPR |= (clksel << RCC_CCIPR_CLK48SEL_SHIFT);
+}
+
+void rcc_clock_setup_hsi16(void)
```
karlp: why isn't this just taking a clock structure too? this seems odd and out of place.  is this leftover from something else?

lib/stm32/l4/rcc.c
```diff
@@ -369,4 +454,136 @@ void rcc_set_msi_range_standby(uint32_t msi_range)
...
+		RCC_BDCR |= (RCC_BDCR_RTCSEL_HSEDIV32 << RCC_BDCR_RTCSEL_SHIFT);
+		break;
+	case RCC_LSE:
+		RCC_BDCR |= (RCC_BDCR_RTCSEL_LSE << RCC_BDCR_RTCSEL_SHIFT);
+		break;
+	case RCC_LSI:
+		RCC_BDCR |= (RCC_BDCR_RTCSEL_LSI << RCC_BDCR_RTCSEL_SHIFT);
+		break;
+	default:
+		/* none selected */
```
karlp: this will actually result in clearing the source if you provide an invalid one? I guess that's ok?

karlp:
Most of this looks good.  Some of the preset clocks look pretty useless though, make a case for them, or I'll probably kill some of the odd middle speed weirdos

lib/stm32/l4/rcc.c
br101: Code seemed clearer and simpler that way. Other families have similar different functions for different clock sources. I am actually using it in bootloader code...

lib/stm32/l4/rcc.c
br101: i suppose so, there are many functions in libopencm3 like this

br101:
Feel free to kill the middle speed weirdos ;-)

lib/stm32/l4/rcc.c
karlp: which ones? you have right below one that uses hsi16 as the clock source and takes a structure?

br101: ok, but it uses PLL as sysclock source, now hsi raw.

L1: rcc_clock_setup_msi(), rcc_clock_setup_hsi(), rcc_clock_setup_pll()
F1: rcc_clock_setup_in_hsi_out_64mhz(), rcc_clock_setup_in_hsi_out_48mhz(), rcc_clock_setup_in_hsi_out_24mhz(), rcc_clock_setup_in_hse_8mhz_out_24mhz(), ... etc... just to name a few

br101:
Hi! Rebased and updated to fix clock presets:
- before the clock speed was only half of the intended speed (HPRE was DIV2)
- removed the "middle speed weirdos"

lib/stm32/l4/rcc.c
karlp: no, not just to name a few :)  F1 is _not_ the example to follow.  Your clock strcut should just include the source.

include/libopencm3/stm32/l4/rcc.h
```diff
@@ -460,6 +462,8 @@ Twelve frequency ranges are available: 100 kHz, 200 kHz, 400 kHz, 800 kHz,
 #define RCC_APB1ENR1_USART2EN			(1 << 17)
 #define RCC_APB1ENR1_SPI3EN			(1 << 15)
 #define RCC_APB1ENR1_SPI2EN			(1 << 14)
+#define RCC_APB1ENR1_WWDGEN			(1 << 11)
+#define RCC_APB1ENR1_RTCAPBEN			(1 << 10)
```
karlp: you need to add these to the rcc_periph_clken enum too please

lib/stm32/l4/rcc.c
karlp: some, most of them do nothing if you pass in the wrong thing.  Just making the value to |= a variable would fix that. (not really important though)

lib/stm32/l4/rcc.c
```diff
@@ -37,12 +37,65 @@
...
+		.hpre = RCC_CFGR_HPRE_NODIV,
+		.ppre1 = RCC_CFGR_PPRE1_DIV2,
+		.ppre2 = RCC_CFGR_PPRE2_DIV2,
+		.voltage_scale = PWR_SCALE1,
+		.flash_waitstates = 4,
+		.ahb_frequency = 80000000,
+		.apb1_frequency = 40000000,
+		.apb2_frequency = 40000000
+	},
+	{ /* 72MHz PLL and 48MHz USB from 24MHz HSE */
```
karlp: This... works, but better is to just still be 80Mhz, and use pllsaiq for clock48 instead of having to cut down to 72Mhz.

include/libopencm3/stm32/l4/rcc.h
```diff
@@ -450,7 +450,9 @@ Twelve frequency ranges are available: 100 kHz, 200 kHz, 400 kHz, 800 kHz,
 #define RCC_APB1ENR1_OPAMPEN			(1 << 30)
 #define RCC_APB1ENR1_DAC1EN			(1 << 29)
 #define RCC_APB1ENR1_PWREN			(1 << 28)
+#define RCC_APB1ENR1_CAN2_USBFSEN		(1 << 26)
```
karlp: Given that these are not available on the same parts, I'd rather see two defines,
RCC_APB1ENR1_CAN2EN for bit 26 and _also_ RCC_APB1ENR1_USBFSEN for bit 26.

include/libopencm3/stm32/l4/rcc.h
karlp: And also to the sleep mode defines for all of these extra ones too.

lib/stm32/common/rtc_common_l1f024.c
```diff
@@ -106,6 +106,7 @@ void rtc_set_wakeup_time(uint16_t wkup_time, uint8_t rtc_cr_wucksel)
 	 *    down-counting.
 	 */
 	RTC_WUTR = wkup_time;
+	RTC_CR &= ~(RTC_CR_WUCLKSEL_MASK << RTC_CR_WUCLKSEL_SHIFT);
 	RTC_CR |= (rtc_cr_wucksel << RTC_CR_WUCLKSEL_SHIFT);
```
karlp: This is probably a place to use an intermediate reg to avoid having to write back multiple times, but it's a minor optimzation, and not something we need to worry about right now. The fix is more important!

karlp:
I've merged most of this series, except, "stm32:l4: rcc: Add some clock presets"  (I still want to see the hsi() function take a clock struct with options like the others do, and "stm32:l4: rcc: Add RCC_APB1ENR1 definitions" (they still need to be added to the RCC_xx helper macros and sleep definitions)

I've closed _this_ PR so it doesn't get polluted with followups, but the remainder are still good stuff.

Thanks for your patience with these!

PR #900, opened 2018-03-27
stm32: iwdg: fix typo in (unused) register name

grevaillot:
well, that's it. no code uses iwdg_sr_wvu yet (window watchdoh), but at least keep naming correct


karlp:
thanks! well spotted

Issue #533, opened 2015-09-27
Nominal STM32 F0 watchdog clock is 40kHz, not 32kHz according to datasheet


onnokort:
Hi,

According to http://www.st.com/st-web-ui/static/active/en/resource/technical/document/reference_manual/DM00031936.pdf, page 553, the IWDG watchdog in STM32F0 parts has a nominal watchdog LSI clock of 40kHz.

Also, my STM32F0DISCOVERY test board's IWDG nominal frequency is close to 40kHz, but not 32kHz.

IMO, there should be either one of two things:

a) a prominent note in the libopencm3 code that iwdg_set_period_ms(..) assumes a 32kHz IWDG LSI clk
or 
b) a separate implementation of iwdg_set_period_ms(..), that calculates the watchdog settings from the nominal 40kHz clock.

Thoughts?

karlp:
Blame ST for making LSI 32k on some parts and 40k on other parts.  You're right though, this should be addressed in some way.  Suggestions welcome.

onnokort:
I suggest splitting it up and having a function that divides by 40 for any STM32F0 part as a start and copy & paste the old implementation for all others & put in a warning that the LSI might be at 40kHz. I volunteer to make a patch for that if that has a chance of being accepted.

Am I right that the linked PDF describes all STM32F0's? I am a bit confused with all the various documents from ST...

karlp:
it would be perfectly valid, just depends how you do it :)  Please bear in mind, the patch queue is fairly long right now, and the current priority is merging f0 and l0 usb support.

If you want to be super awesome, please make sure the commits have good clear summary lines, and ideally, for things like this, some example device code that should run on both old/new and can show the failure.  It makes things much easier for people to verify the changes.  Finally, please make sure you check all the STM32 reference manuals to get the right values for the right targes

karlp:
I propose easiest for this is to simply move the #define LSI_FREQUENCY 32000
out of the iwdg.c file and into the iwdg.h headers for each target.  Should be pretty easy

PR #901, opened 2018-03-27
stm32l0: enable iwdg, same peripheral as f0, f3, l4..

grevaillot:
Basically same peripheral than f0, f3, l4. tested and working, so same job as 
https://github.com/libopencm3/libopencm3/commit/c90c9fe801a6168518b411b3a829cc515e6c94ee

### DIFF
--- a/include/libopencm3/stm32/iwdg.h
+++ b/include/libopencm3/stm32/iwdg.h
@@ -30,6 +30,8 @@
 #       include <libopencm3/stm32/f3/iwdg.h>
 #elif defined(STM32F4)
 #       include <libopencm3/stm32/f4/iwdg.h>
+#elif defined(STM32L0)
+#       include <libopencm3/stm32/l0/iwdg.h>
 #elif defined(STM32L1)
 #       include <libopencm3/stm32/l1/iwdg.h>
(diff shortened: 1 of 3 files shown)


grevaillot:
Note that currently, libopencm3's iwdg_common_all.c iwdg_set_period_ms does not take in account device LSI clock frequency - so, prescaler selection will be off on devices with a faster/slower LSI than the expected 32khz - but i think that that issue is not in the scope of that PR.

I guess that the two addionnal (and unused) definition could be centralized - or removed as no use is made yet of the window function.

karlp:
yes, the LSI clock is a disaster. I think there's even an issue for it already. https://github.com/libopencm3/libopencm3/issues/533 covers most of it I think. but yes, out of scope of this PR

karlp:
I actually think it's time to make a iwdg_common_v2.h and use that for f3, l0, f0, l4, rather than keep copying this little stub around actually.

grevaillot:
ok, i'll have a look and prepare a PR.

grevaillot:
new pr is https://github.com/libopencm3/libopencm3/pull/903, closing this one

PR #902, opened 2018-03-28
stm32: add some F0 devices

grevaillot:
i'm going to bring-up some f030cc based boards soon, add device, and a couple of other f0 missing devices.


karlp:
tahnks for this!

PR #903, opened 2018-03-28
stm32: refactor iwdg WINR defines into iwdg_common_v2, enable stm32l0 iwdg

grevaillot:
two part PR, following https://github.com/libopencm3/libopencm3/pull/901 :
- move IWDG_SR's WVU bit and IWDG_WINR defines from copypasted device header (f0, f3, l4) to a iwdg_common_v2.h header
- enable stm32l0 iwdg


karlp:
Looks good.  I don't think you should actually need the stub .c file, but I guess you're copying the others, and it's out of scope here, it's just more doxygen work for me later I think :)

grevaillot:
it seems the stub .c file was needed to have the doxygen properly generated, I'm not exactly sure why, so indeed, i aligned :).

karlp:
merged thanks, along with some doxygen I cleaned up with it.  The stub .c files are for another day indeed.

PR #742, opened 2017-01-23
STM32 SPI Support for L0, restructuring for L1, F0, F1, F2, F3, F4

SevenW:
I have prepared SPI support for L0. I carefully studied the reference manual of several STM32 versions, as explained in #740.

I did not address the value line F0x0 and F100 here. For all other versions, the proposed header file structure reflects the reference manuals. I moved two functions the were in spi_common_all.h to another header file, as the corresponding registers/bits are not available on the F0 and F3.

I am open for renaming one of the levels to spi_common_v2.

The thing that I am not happy with are the C-files. For spi_common_f1.c and  spi_common_l01f24.c I now have a duplication of three functions.
I see solutions. None of them perfect.

I like to see this pull request considered in combination with #663 to make a big step in supporting L0.

Tested on STM32L052 (JeeNode Zero rev3)

Issue #740, opened 2017-01-20
SPI driver structure for L0 L1 F0 F1 F2 F4


SevenW:
I am in the process of adding a STM32 L0 SPI driver. Once done, I like to make a pull request. I see a lot of similarity among the L's and F's, and now I am seeking advice on structuring the files, I compared "SPI register map" tables from various STM32 reference manuals. Compared to starting with spi_common_all.h the following types of STM32 add capabilities:

```
common_all < F1 < {L0, L1, F2, F4}
and:
common_all < F1 < F0 (this adds other functionality then the above)
```
Note: some low budget mcu's F100, F0x0 have fewer registers then common_all. Missing two full registers (not only bits).

Observation on current implementation:
`common_all < L1F124 < F24`

What I want to do is separate out L1 and make a common L01F24

Please advice on the naming of the spi_common_*.h

fenugrec:
As I understand, I think the current direction is to drop the "L1F124" naming when it gets to a certain point of complexity. ADC and USB do this by grouping into arbitrary "versions", such as "st_usbfs_v1" and "st_usbfs_v2".

What are the registers "missing" from the low-end F1 and F0x0 chips you mention ?

(note, I'm mostly a spectator in this project, I don't make decisions ! )

SevenW:
I was in doubt on the _v2 naming convention. In tow cases, the _v2 suffix is an addition to the original _common_all (adc, usbfs) and in one case it is a full replacement (usart). 
Looking at this it seems appropriate to give the F1 version the name spi_common_v2.h

I will come back to the "reduced function" F1 and F0.

SevenW:
Missing from value line F0x0 and F100, with respect to other F0/F1:
```
register 0x1c: SPIx_I2SCFGR
register 0x1e: SPIx_I2SPR
flag b2: CHSIDE from register 0x08:SPIx_SR
flag b3: UDR from register 0x08:SPIx_SR
```

karlp:
it was my understanding that all L0 chips have the "spi v2+" peripheral, with the fifos, that comes in with f3 and f0.  There's probably not a lot you want to try and get too common.  Are you sure there's not a PR for spi on l0 already?

SevenW:
Is a PR a github "issue"? Could not find it.
I also could not find a pull request.

I have prepared and tested a L0 implementation. It is not a lot of change. I try to prepare a pull request tonight. (The only open question for me: should I rename spi_common_l1.h to spi_common_v2.h?)

[...]

PR #906, opened 2018-04-04
stm32: rework spi, based on PR #740 and #742.

grevaillot:
ok, so i did a quick pr based on #742 and #740 comments and previous work from @SevenW - this is more a draft than anything else yet, but seems a bit nicer than current master tree state. l0 spi is now enabled too.

This split spi stuff in three part:
- v1 : basic spi peripheral
- v1_frf : v1 spi with frf mode additional bit in spi_cr2 / spi_sr
- v2 : spi with variable datasize, fifo and other fancy stuff.

Then,
- v1 maps to f1 chips
- v1_frf to f2, f4 and l0,l1
- v2 to f0, f3 and l4

Note: This **breaks** spi_master_init API for v2 devices : function prototype from common spi header used to be abused, with DFF bit reused for CRCL bit.
Now, v2 spi_master_init does not handle anymore CRCL bits, as it does not usually mess with other crc configuration. I'm not exactly happy of that, as I keeping a consistent api between devices is nice, but DFF just does not exist on f0,f3 and l4,


lib/stm32/common/spi_common_v2.c
```diff
@@ -59,40 +59,35 @@ LSB first.
...
 
 	reg32 |= br;		/* Set baud rate bits. */
 	reg32 |= cpol;		/* Set CPOL value. */
 	reg32 |= cpha;		/* Set CPHA value. */
-	reg32 |= crcl;		/* Set crc length (8 or 16 bits). */
 	reg32 |= lsbfirst;	/* Set frame format (LSB- or MSB-first). */
 
-	/* TODO: NSS pin handling. */
-
+	SPI_CR2(spi) |= SPI_CR2_SSOE;
```
karlp: this is out of scope of this commit, and I'll probably split it locally.  It's consistent with what the v1 code does, ie, it's _right_ but it's not part of the refactor here :)

karlp:
what do you think about dropping the dff altogether from spi init master v1? then they can both be same api

grevaillot:
I'm all for it. I guess it could be possible to do something to but back spi_init_master into spi_common, i'll have a look.

grevaillot:
So, this could work but, i'm wondering if (why?) we really need to keep the CRC bits untouched in spi_init_master while reseting the whole register. Also, do we clear DFF bit ? Current logic clears everything except CRC related stuff and SPIE.

Currently we would need to add a not-so-nice mask, I would prefer to avoid that, ie : https://github.com/grevaillot/libopencm3/commit/4e805dc48716681e8124a65ea232bc9a747b6fb3

karlp:
I've merged this for now to get a groundwork going forward.

PR #907, opened 2018-04-12
Add STM32L0 to dispatch/irqhandlers.h

MarSik:
The interrupt dispatch header file was missing support for STM32L0.

### DIFF
--- a/include/libopencmsis/dispatch/irqhandlers.h
+++ b/include/libopencmsis/dispatch/irqhandlers.h
@@ -10,6 +10,8 @@
 #	include <libopencmsis/stm32/f4/irqhandlers.h>
 #elif defined(STM32L1)
 #	include <libopencmsis/stm32/l1/irqhandlers.h>
+#elif defined(STM32L0)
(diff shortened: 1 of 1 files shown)


karlp:
not really missing, more like deliberately ommitted. no-one really uses this opencmsis stuff, it's a compat layer that isn't really needed anymore.  Could you elabroate why you need it?

MarSik:
Hi,

it might have been a leftover in my code, but I do not see a wfi method
anywhere else for example. Not that it is hard to do it manually, but..

I just though it would be nice if the code actually worked when it is part
of the repo.

Martin




2018-04-12 11:56 GMT+02:00 Karl Palsson <notifications@github.com>:

karlp:
what do irq handlers have to do with wfi wrappers?

MarSik:
__WFI is defined in opencmsis/core_cm3.h which includes
libopencmsis/dispatch/irqhandlers.h and that together throws warnings
when used on STM32L0. This simple fix just make that combination
working properly.

2018-04-13 12:07 GMT+02:00 Karl Palsson <notifications@github.com>:

PR #916, opened 2018-05-05
stm32 F3 rcc add clock configs for HSE, drop bad misleading ones

karlp:


karlp:
merged in ab9e4252 and ee8b5bf4

PR #919, opened 2018-05-10
EFM32HG: Add USHFRCODIV2 clock

xobs:
The USHFRCODIV2 clock is derived from a 48 MHz PLL that is trimmed using clock recovery from USB.  It is the most accurate PLL on the EFM32HG, and should be used if any vaguely timing-accurate clocks are necessary.

This patchset adds a definition for it, enables switching the HFCLK to derive the core clock from it, and fixes some warnings for things like oscillator enabling, where the new USHFRCODIV2 clock can't really be enabled separately.

### DIFF
--- a/include/libopencm3/efm32/hg/cmu.h
+++ b/include/libopencm3/efm32/hg/cmu.h
@@ -352,6 +352,7 @@
 #define CMU_CMD_HFCLKSEL_HFXO		CMU_CMD_HFCLKSEL(2)
 #define CMU_CMD_HFCLKSEL_LFRCO		CMU_CMD_HFCLKSEL(3)
 #define CMU_CMD_HFCLKSEL_LFXO		CMU_CMD_HFCLKSEL(4)
+#define CMU_CMD_HFCLKSEL_USHFRCODIV2	CMU_CMD_HFCLKSEL(5)
 
 /* CMU_LFCLKSEL */
 /* Bits 31:21 - Reserved */
@@ -633,6 +634,7 @@ enum cmu_osc {
 	LFXO, /**< External, 32.768kHz */
 	AUXHFRCO, /**< Internal, 1-28Mhz */
 	USHFRCO, /**< Internal, 48MHz */
+	USHFRCODIV2, /**< Internal, 24MHz */
 };
 
 /* --- Function prototypes ------------------------------------------------- */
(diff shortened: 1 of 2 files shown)


include/libopencm3/efm32/hg/cmu.h
```diff
@@ -633,6 +634,7 @@ enum cmu_osc {
 	LFXO, /**< External, 32.768kHz */
 	AUXHFRCO, /**< Internal, 1-28Mhz */
 	USHFRCO, /**< Internal, 48MHz */
+	USHFRCODIV2, /**< Internal, 24MHz */
```
karlp: I don't really think this should be treated as a separate oscillator.  I think the comment above should be changed, but normally, and on other platforms, these optionally divided clocks are treated as dividers in the clock configs, not separate clocks.

karlp: But I also see that it's a command option in the HFCLKSEL.  *shrugs*  I'll look at the rest of the patch.  refman is a little weird here maybe.

xobs: The HFCLKSEL can take one of 5 values: `HFRCO`, `HFXCO`, `LFRCO`, `LFXCO`, and `USHFRCODIV2`.  `USHFRCO` is not one of the values.

How do you think this should be handled in `cmu_set_hfclk_source()` if `USHFRCODIV2` is not given an enum in `cmu_osc`?

karlp: looking at it again, should we just drop the USHFRCO from the list? would that not be most self consistent?

karlp: nah, that's wrong too.  Hoenstly, I think the problem is trying to use the same enumeration for oscillators on/off, and _also_ for "cpu clock sources"

karlp: nah, I think that's even more confusing :)  I don't really like it, but I don't see anything better at the moment, and the clock tree for the HG is actually fairly simply really, so I'll probably just pull this as is and see what turns out in the future

karlp:
Following my own conclusion and merging this as to to see what shakes out.  No idea how this didn't land when I said that.  sorry about that.

PR #922, opened 2018-05-19
stm32f0:rcc: add fonction to configure system clock from 16MHz HSE

cashew22:


karlp:
No, these will be replaced by the same stuct based methods that all the other families (except f1) use, we're not going to continue adding functions for everything like this.

PR #924, opened 2018-05-26
minor USB improvements

eqvinox:
(see individual commits)

### DIFF
--- a/include/libopencm3/usb/usbd.h
+++ b/include/libopencm3/usb/usbd.h
@@ -70,7 +70,14 @@ extern const usbd_driver efm32hg_usb_driver;
  *
  * It is required that the 48MHz USB clock is already available.
  *
- * @param driver TODO
+ * @param driver Pointer to an USB driver's information structure.  The
+ *               following drivers are available:
+ *               @li @link st_usbfs_v1_usb_driver @endlink
+ *               @li @link stm32f107_usb_driver @endlink
+ *               @li @link stm32f207_usb_driver @endlink
+ *               @li @link st_usbfs_v2_usb_driver @endlink
+ *               @li @link efm32lg_usb_driver @endlink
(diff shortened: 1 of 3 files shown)


include/libopencm3/usb/usbd.h
```diff
@@ -70,24 +70,40 @@ extern const usbd_driver efm32hg_usb_driver;
...
  *
- * @param driver TODO
+ * @param driver Pointer to an USB driver's information structure.  The
+ *               following drivers are available:
+ *               @li @link st_usbfs_v1_usb_driver @endlink
+ *               @li @link stm32f107_usb_driver @endlink
+ *               @li @link stm32f207_usb_driver @endlink
+ *               @li @link st_usbfs_v2_usb_driver @endlink
+ *               @li @link efm32lg_usb_driver @endlink
+ *               @li @link efm32hg_usb_driver @endlink
```
karlp: oooh nice doxygen.  is there anyway to j ust link to a type name? we can then make the drivers be members of that instead.  ie, the efm32 driver is already refactored in my local tree.

karlp:
Can you not just put your desired disconnect function into the driver struct?

include/libopencm3/usb/usbd.h
eqvinox: I have no idea, I wrote this with googling and the Doxygen manual open on the side ;)

eqvinox:
Modifying the driver struct is kindof impossible:
- the struct itself is `const` (rightfully so, as it should be in flash)
- the driver functions are static (for some drivers, anyway) and thus can't be externally referenced (and that'd be ugly anyway)
- creating a modified copy of the driver struct could therefore only happen at runtime and waste precious RAM

I've thought about it quite a bit and really think a weak `usbd_disconnect` is the best way to go.

karlp:
I've pulled these locally, they all look good.

PR #926, opened 2018-05-31
Support for circular mode ADC DMA

mmone:
This completes the support for circular mode ADC DMA transfers. Allowing for continuous transfer of ADC readings to memory without the need for external triggers. This mode needs to be configured on both the ADC and the DMA side.
Until yet only the DMA side was implemented.

The ADC part is documented in the [STM32F3 Reference Manual](http://www.st.com/resource/en/reference_manual/dm00043574.pdf) on page 349

### DIFF
--- a/include/libopencm3/stm32/common/adc_common_v2.h
+++ b/include/libopencm3/stm32/common/adc_common_v2.h
@@ -205,7 +205,8 @@ bool adc_get_overrun_flag(uint32_t adc);
 void adc_clear_overrun_flag(uint32_t adc);
 uint32_t adc_read_regular(uint32_t adc);
 void adc_start_conversion_regular(uint32_t adc);
-
+void adc_enable_dma_circular_mode(uint32_t adc);
+void adc_disable_dma_circular_mode(uint32_t adc);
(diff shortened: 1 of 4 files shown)


karlp:
This needs to be added to adc_common_v2.c, not just for the f3.  This is a common bit definition, for F0,L0, F3,L4.

mmone:
I'm aware of that but have only F3 hardware for testing.

karlp:
So mention that. But you have access to the documentation for the others, and I certainly won't merge it if you try and only add it to the f3 :)  All you need to do is move the functions to another file, test it still behaves as you expect on your device, and mention something like, "Tested on an f3, but matches documentation for the other parts"

mmone:
Fair enough. I will crunch some pdfs.

karlp:
I already did, that's why I said this is common and you should apply it to the common code :)

mmone:
I've updated the PR. Thanks for the assist.

mmone:
Seems Travis has a hickup.

karlp:
merged in f6517f78 thanks!

PR #928, opened 2018-06-12
stm32: Add GD32F1x0 series support stub.

rgwan:
Currently this commit has confirmed USB, interrupt, SPI is working.

USART seems have some problem, GPIO set/reset/toggle works, but gpio_set_mode doesn't work because I haven't implemented yet: The GPIO configuration registers is totally different from STM32F1 series.

We can sure that the GD32 use the same USB IP as the STM32, and RCC register is almost the same.

But some peripheral's clock switch is not correct. It needs time and people working on it.


include/libopencm3/stm32/common/timer_common_all.h
```diff
@@ -50,7 +50,9 @@ specific memorymap.h header before including this header file.*/
 #endif
 #define TIM5				TIM5_BASE
 #define TIM6				TIM6_BASE
+#if defined(TIM8_BASE)
```
karlp: this should be TIM7 surely, and you can add these as standalone commit

include/libopencm3/stm32/f1x0/adc.h
```diff
@@ -0,0 +1,423 @@
+/** @defgroup adc_defines ADC Defines
+
+@brief <b>Defined Constants and Types for the STM32F1xx Analog to Digital
+Converters</b>
+
+@ingroup STM32F1xx_defines
```
karlp: you need to clean up all your typos here from copy/paste

include/libopencm3/stm32/f1x0/adc.h
```diff
@@ -0,0 +1,423 @@
...
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**@{*/
+
+#ifndef LIBOPENCM3_ADC_H
+#define LIBOPENCM3_ADC_H
+
+#include <libopencm3/stm32/common/adc_common_v1.h>
+
```
karlp: There looks to be a lot of duplication in the register defines here? is there room for more extraction into adc_common_v1?

include/libopencm3/stm32/f1x0/can.h
```diff
@@ -0,0 +1,34 @@
+/** @defgroup can_defines CAN defines
+
+@ingroup STM32F1xx_defines
```
karlp: you have these sorts of errors everwhere.

karlp:
This looks like you just copied the entire f1 directory to f10x and changed.... somethings?

Please, that won't be maintainable like that, please actually work on extracting common code if you want to use it like this.

I'm really not sure that stm32/f10x is a very clear location.  Can it not just be in gd32/f10x?  It can still use vpath to see code in the stm32 folders without any problem, and that should help keep it clear what's common and what is not.

rgwan:
Okay, I will revoke this commit and fix that problem you have pointed out at my spare time, maybe next yr. Thanks.

rgwan:
Okay, I will revoke this commit and fix that problem you have pointed out at my spare time, maybe next yr because now I have to prepare my postgraduate entrance, sorry for that. Thanks you for understanding.

karlp:
No problem, you're always welcome.

Icenowy:
@karlp I'm trying to redo this PR (@rgwan says he's too busy to do this).

By the way I found that if we place it into a dedicated GD32 folder when dealing with headers, we will meet a mess when trying to reuse STM32 code, as STM32 code won't try to include GD32 headers.

In order to ease the code reuse, I suggest something like stm32/gdf1x0 , although the name seems not so direct.

karlp:
I _kinda_ feel that putting it libopencm3/gd32/blah and having the stm32 files include the gd32 headers _only for the peripherals that are the same_ is probably the best choice.  So the things that are different for gd32 stay entirely withing the gd32 tree, and gd32 refers into the stm32 tree, rather than dropping all of gd32 into stm32.  Gigadevices have some less compatible parts too, so I think that's better for future locations for gd parts, even if the original gd parts were stm32 "compatible"

PR #929, opened 2018-06-14
Added linker scripts for STM32F03xz4, STM32F04xz4, STM32F05xz4 devices.

Misaka0x2730:
Added linker scripts for family STM32F0 with 16kB Flash.


karlp:
no, we're not adding any more of these, it explodes into too many files too easily.  Use the linker script generator code or just use the generic one in your own projects.

PR #931, opened 2018-06-20
USB: Add streaming audio descriptors

schnommus:
This PR adds a bunch of audio descriptors; which are needed for enumerating as a streaming feature unit (i.e speakers or microphones).
The data is taken directly from the USB spec.

I have verified that they are correct by emulating a 2-channel 8KHz microphone on an EFM32HG, demo code is here: https://github.com/schnommus/libopencm3-examples/blob/audio/examples/efm32/efm32hg/tomu-efm32hg309/usb_audio/usb_audio.c

(Note that the USB stack currently requires a hack for isochronous endpoints to work correctly, but I'll file that separately)


include/libopencm3/usb/audio.h
```diff
@@ -6,10 +6,11 @@
 
 @version 1.0.0
 
-@author @htmlonly &copy; @endhtmlonly 2014
+@author @htmlonly &copy; @endhtmlonly 2018
 Daniel Thompson <daniel@redfelineninja.org.uk>
+Seb Holzapfel <schnommus@gmail.com>
 
-@date 19 April 2014
+@date 20 June 2018
```
karlp: don't change this, you should be (if you must) _adding_ years, not dropping them

schnommus: no worries - reverted. (accidentally blindly changing dates...)

include/libopencm3/usb/audio.h
```diff
@@ -6,8 +6,9 @@
 
 @version 1.0.0
 
-@author @htmlonly &copy; @endhtmlonly 2014
+@author @htmlonly &copy; @endhtmlonly 2018
```
karlp: and this one :)

include/libopencm3/usb/audio.h
schnommus: also I'm having trouble getting doxygen to actually render this file so it's hard for me to check :(

include/libopencm3/usb/audio.h
```diff
@@ -88,6 +90,143 @@ struct usb_audio_header_descriptor_body {
...
+	uint8_t bBitResolution;
+	uint8_t bSamFreqType;
+	/* ... */
+} __attribute__((packed));
+
+/* Table 2-2: Continuous Sampling Frequency */
+struct usb_audio_format_continuous_sampling_frequency {
+	/* ... */
+	uint32_t tLowerSamFreq : 24;
+	uint32_t tUpperSamFreq : 24;
```
karlp: is this really the best way?  bitfields have notorious portability problems, so I tend to just avoid them outright.  I can't relaly think of anythign better though, and it _works_ so... probably the best way?

schnommus: I mean, we could use an array of 3 uint8_t's with a special macro; but I would prefer the bitfield

karlp: yeah, it's good, let's keep it as is.

include/libopencm3/usb/audio.h
schnommus: Done.

include/libopencm3/usb/audio.h
```diff
@@ -88,6 +90,143 @@ struct usb_audio_header_descriptor_body {
...
+struct usb_audio_feature_unit_descriptor_head {
+	uint8_t bLength;
+	uint8_t bDescriptorType;
+	uint8_t bDescriptorSubtype;
+	uint8_t bUnitID;
+	uint8_t bSourceID;
+	uint8_t bControlSize;
+	uint16_t bmaControlMaster; /* device can assume 16-bit, given highest
+				    * defined bit in spec is bit #9.
+				    * (it is thus required bControlSize=2) */
```
karlp: seems reasonable from the spec, and people could construct private ones with 
```
struct extended {
   struct usb_audio_feature_unit_descriptor_head head;
   struct private_extended_body body;
   struct usb_audio_feature_unit_descriptor_tail tail;
}
```
so we're not blocking anyone out either

karlp:
I've pulled this locally to my staging area

PR #934, opened 2018-06-27
bugfix flash STM32f24 dual bank devices with memoery of 2MB. No prope…

MerijnvanEijk:
Proposed Bugfix flash STM32f24 dual bank devices with memory of 2MB. No proper sector assignment occurred.
The STM32F4 devices with dual bank 2MB memory use an additional bit in the FLASH_CR register.
This is bit 7, and the SNB bits is 0:4 and not only 0:3. But the definition is not given to this bit 7. So opted for a positive modification to the sector value to align it with the reference manual allowed value.

From reference manual
STM32 Doc ID DM00031020. Rev 17 Page 106:
SNB[3:0]: Sector number _<typo in STM32 reference manual should be [4:0]>_
These bits select the sector to erase.
0000: sector 0
0001: sector 1
...
01011: sector 11
01100: not allowed
01101: not allowed
01110: not allowed
01111: not allowed
10000: section 12
10001: section 13
...
11011 sector 23
11100: not allowed
11101: not allowed
11110: not allowed
11111: not allowed

### DIFF
--- a/include/libopencm3/stm32/common/flash_common_f24.h
+++ b/include/libopencm3/stm32/common/flash_common_f24.h
@@ -88,6 +88,8 @@
 #define FLASH_CR_PROGRAM_X64		3
 /**@}*/
 
+#define FLASH_CR_SNB_BANK_BIT_OFFSET	0x04
+
 /* --- FLASH_OPTCR values -------------------------------------------------- */
 
 /* FLASH_OPTCR[27:16]: nWRP */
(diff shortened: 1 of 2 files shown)


zyp:
Looks good, but the name `FLASH_CR_SNB_BANK_BIT_OFFSET` doesn't make much sense to me.

I propose `#define FLASH_CR_SNB_BANK2_START 16` and `sector += FLASH_CR_SNB_BANK2_START - FLASH_SECTOR_BANK2_START` to make everything more readable.

PR #939, opened 2018-07-05
Power fixes

ChuckM:
This PR updates the pwr.h file and the rcc.c file to deal with a multibit VOS field.

### DIFF
--- a/include/libopencm3/stm32/common/flash_common_f234.h
+++ b/include/libopencm3/stm32/common/flash_common_f234.h
@@ -50,7 +50,8 @@
 
 /* --- FLASH_ACR values ---------------------------------------------------- */
 
-#define FLASH_ACR_LATENCY_MASK		0x07
+#define FLASH_ACR_LATENCY_MASK		0x0f
+#define FLASH_ACR_LATENCY(w)		((w) & FLASH_ACR_LATENCY_MASK)
 #define FLASH_ACR_LATENCY_0WS		0x00
 #define FLASH_ACR_LATENCY_1WS		0x01
 #define FLASH_ACR_LATENCY_2WS		0x02
--- a/include/libopencm3/stm32/f4/pwr.h
+++ b/include/libopencm3/stm32/f4/pwr.h
@@ -43,10 +43,11 @@ LGPL License Terms @ref lgpl_license
 
 /* --- PWR_CR values ------------------------------------------------------- */
 
-/* Bits [31:15]: Reserved */
+/* Bits [31:16]: Reserved */
 
 /* VOS: Regulator voltage scaling output selection */
-#define PWR_CR_VOS			(1 << 14)
+#define PWR_CR_VOS_SHIFT		14
+#define PWR_CR_VOS_MASK			0x3
 
 /* Bits [13:10]: Reserved */
 
@@ -72,14 +73,13 @@ LGPL License Terms @ref lgpl_license
 
 /* --- Function prototypes ------------------------------------------------- */
 
-enum pwr_vos_scale {
-	PWR_SCALE1,
-	PWR_SCALE2,
-};
+#define	PWR_SCALE1	0x3
+#define PWR_SCALE2	0x2
+#define PWR_SCALE3	0x1
 
 BEGIN_DECLS
 
-void pwr_set_vos_scale(enum pwr_vos_scale scale);
+void pwr_set_vos_scale(int8_t scale);
 
 END_DECLS
 
--- a/lib/stm32/f4/pwr.c
+++ b/lib/stm32/f4/pwr.c
@@ -36,11 +36,9 @@
 
 #include <libopencm3/stm32/pwr.h>
 
-void pwr_set_vos_scale(enum pwr_vos_scale scale)
+void pwr_set_vos_scale(int8_t scale)
 {
-	if (scale == PWR_SCALE1) {
-		PWR_CR |= PWR_CR_VOS;
-	} else if (scale == PWR_SCALE2) {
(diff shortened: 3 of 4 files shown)


include/libopencm3/stm32/common/flash_common_f234.h
```diff
@@ -50,7 +50,8 @@
 
 /* --- FLASH_ACR values ---------------------------------------------------- */
 
-#define FLASH_ACR_LATENCY_MASK		0x07
+#define FLASH_ACR_LATENCY_MASK		0x0f
+#define FLASH_ACR_LATENCY(w)		((w) & FLASH_ACR_LATENCY_MASK)
 #define FLASH_ACR_LATENCY_0WS		0x00
 #define FLASH_ACR_LATENCY_1WS		0x01
 #define FLASH_ACR_LATENCY_2WS		0x02
```
karlp: I've been inclinded to just drop all of these.  latency is just a natural number.  as long as we have the mask we should be good.  doing it like this is probably good interim step, but in all reality, without versioned tags, there's no way of doing a timed deprecate,remove phase

include/libopencm3/stm32/f4/pwr.h
```diff
@@ -72,14 +73,13 @@ LGPL License Terms @ref lgpl_license
 
 /* --- Function prototypes ------------------------------------------------- */
 
-enum pwr_vos_scale {
-	PWR_SCALE1,
-	PWR_SCALE2,
-};
+#define	PWR_SCALE1	0x3
+#define PWR_SCALE2	0x2
+#define PWR_SCALE3	0x1
```
karlp: hrm? what's worng with the enum?  we use an enum on all the other platforms with multi bits, eg f7, l1, l0, (stopped looking)

lib/stm32/f4/pwr.c
```diff
@@ -36,11 +36,9 @@
...
+void pwr_set_vos_scale(int8_t scale)
 {
-	if (scale == PWR_SCALE1) {
-		PWR_CR |= PWR_CR_VOS;
-	} else if (scale == PWR_SCALE2) {
-		PWR_CR &= PWR_CR_VOS;
-	}
+	uint32_t reg = PWR_CR & ~(PWR_CR_VOS_MASK << PWR_CR_VOS_SHIFT);
+	reg |= (scale & PWR_CR_VOS_MASK) << PWR_CR_VOS_SHIFT;
+	PWR_CR = reg;
```
karlp: it does make less code doesn't it...  perhaps the enums with defined values?

lib/stm32/f4/rcc.c
```diff
@@ -59,9 +59,9 @@ const struct rcc_clock_scale rcc_hse_8mhz_3v3[RCC_CLOCK_3V3_END] = {
 		.hpre = RCC_CFGR_HPRE_DIV_NONE,
 		.ppre1 = RCC_CFGR_PPRE_DIV_4,
 		.ppre2 = RCC_CFGR_PPRE_DIV_2,
-		.power_save = 1,
+		.power_save = PWR_SCALE1,
 		.flash_config = FLASH_ACR_DCEN | FLASH_ACR_ICEN |
-				FLASH_ACR_LATENCY_3WS,
+				FLASH_ACR_LATENCY_1WS,
```
karlp: this looks like a standalone bugfix?

lib/stm32/f4/rcc.c
```diff
@@ -75,6 +75,7 @@ const struct rcc_clock_scale rcc_hse_8mhz_3v3[RCC_CLOCK_3V3_END] = {
 		.hpre = RCC_CFGR_HPRE_DIV_NONE,
 		.ppre1 = RCC_CFGR_PPRE_DIV_2,
 		.ppre2 = RCC_CFGR_PPRE_DIV_NONE,
+		.power_save = PWR_SCALE1,
```
karlp: I think we should stop calling this power_save while we're here, it's called ```voltage_scale``` on l1/l0, and ```vos_scale``` everywhere else.  (We should even move l1/l0 to vos_scale as well, but that's out of scope of this)

lib/stm32/f4/rcc.c
```diff
@@ -254,9 +260,9 @@ const struct rcc_clock_scale rcc_hse_25mhz_3v3[RCC_CLOCK_3V3_END] = {
 		.hpre = RCC_CFGR_HPRE_DIV_NONE,
 		.ppre1 = RCC_CFGR_PPRE_DIV_4,
 		.ppre2 = RCC_CFGR_PPRE_DIV_2,
-		.power_save = 1,
+		.power_save = PWR_SCALE1,
 		.flash_config = FLASH_ACR_DCEN | FLASH_ACR_ICEN |
-				FLASH_ACR_LATENCY_3WS,
+				FLASH_ACR_LATENCY_1WS,
```
karlp: another pre-existing bug?

lib/stm32/f4/rcc.c
ChuckM: copypasta code for when someone created the 48Mhz struct they didn't update wait states.

lib/stm32/f4/pwr.c
ChuckM: either works although when they change it to 3 bits the enum needs to be reworked. The defined values allows code transparency while the changes happen at the loss of compiler ability to detect mis-matched arguments. This function isn't called in a lot of code (certainly on the F4 side)

karlp:
fixed all via https://github.com/libopencm3/libopencm3/pull/949/commits/2649c5e493028b54b67baa6c202ac17a4c9b8951 aiui

PR #943, opened 2018-07-13
Device Data: Add Specs for STM32F4[67]9 variant

ChuckM:
This commit adds 3 lines to the devices.data file to include the
STM32F469 and STM32F479. This is the only F4 with 240K of contiguous
memory.


karlp:
I count 320k contiguous?
sram1,2,3 are all contiguous on the datasheet?

ChuckM:
Ok, it gets weirder.  I built with the 240K number and my code crashed (hard fault) it has run when it was still reporting 192K. I looked at the reference manual RM0386 Rev 3 (Jan 2017) and it shows RAM is only mapped to the first 192K of SRAM addresses even though there is more RAM than that. The datasheet from May of 2018 show a different mapping for RAM showing the 320K. Going to do a bit more validation on the hardware I'm running (the '469 disco board)

ChuckM:
Ok, so this is the final resolution on this. 

I compared the Data sheet and the reference manual. Originally to Rev 3 but then to the June 2018 Rev 5 version. They both show only 192K worth of mappings. The Data sheet however is correct in that it shows the correct 320K mapping. The devices data file has been updated to reflect the actual part. 

I ran into a bit of a red herring in that my failing code was using malloc, which in the gcc embedded library starts with a small heap (4K typically) although with the "old" LD script it was getting bumped to 4956 bytes (not sure why) and my application used malloc and allocated all of it. Once I added by my _sbrk_r function which gave the memory below the stack and above the .bss to the heap it went back to working as expected, with the 320K memory limit.

karlp:
merged via rebase and squish in 63e0b6df thanks

PR #946, opened 2018-07-22
Added basic support for MSP432E4xx:

Misaka0x2730:
1. Added full memory map.
2. Added some functions for system control.


karlp:
_lots_ of this is duplicated with lm4f files.  Far more than I was expecting.

Further, if you're going to the effort of putting in all the comments, if you put them in doxygen style, you automatically get them documented as well.

```
#define BLAH_XXX  1234 /* this comment will never be shown anywhere */
/** This comment will automatically document BLAH_GOOD_DOXY */
#define BLAH_GOOD_DOXY 1234
```

include/libopencm3/msp432/e4/systemcontrol.h
```diff
@@ -0,0 +1,827 @@
...
+ * @li POWER_DISABLE - Module is not powered and does not receive a clock
+ * @li POWER_ENABLE  - Module is powered but does not receive a clock
+ *
+ * @note If the module is in run, sleep or deep-sleep mode - the module
+ * is powered and receives a clock regardless of the value of power mode.
+ */
+enum msp432_power_mode {
+    POWER_DISABLE = 0,
+    POWER_ENABLE
+};
```
karlp: what's wrong with just a bool for this?

Misaka0x2730:
@karlp one more question - is there a need to put into doxygen all definitions? 
I understand correctly, what you say, for example, about it?
~~~~
/* SYSCTL Interrupt Mask Control (SYSCTL_IMC) ------------------------------- */
#define SYSCTL_IMC_MOSCPUPIM    (1 << 8) /* MOSC Power Up Raw Interrupt Status*/
#define SYSCTL_IMC_PLLLIM       (1 << 6) /* PLL Lock Raw Interrupt Status */
#define SYSCTL_IMC_MOFIM        (1 << 3) /* Main Oscillator Failure Raw
                                          * Interrupt Status */
#define SYSCTL_IMC_BORIM        (1 << 1) /* Brown-Out Reset Raw
                                          * Interrupt Status */
~~~~
First, it will greatly increase the length of the file.
Secondly, I examined the header files (especially stm32) and did not see such comments in the doxygen. Please, if I'm wrong - give an example or tell the file where I can see how to do it.

karlp:
You don't need to know.  You could have just left off all the comments, no problem.

But if you're going to include them, including them in doxygen style means that good IDEs can automatically provide them in tooltips, and the generated doxygen docs are more useful.  The length of the header is completely irrelevant.  

See https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/stm32/f4/rcc.h#L48 for an example.  Again, this isn't required, but if you're going to the effort of including comments, including them in a parseable form is vastly better :)

A lot of code has been contributed without it in the past, but slowly restyling so that generated docs are useful is an ongoing goal.

Misaka0x2730:
@karlp Okay, thank you very much, I understand!

Misaka0x2730:
@karlp 
> Further, if you're going to the effort of putting in all the comments, if you put them in doxygen style, you automatically get them documented as well.

Fixed.

lib/msp432/e4/systemcontrol.c
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * @date 22 July 2018
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2012 Alexandru Gagniuc <mr.nuke.me@gmail.com>
```
karlp: did you keep anything from the?  I can't really see it, but if you think it's worthwhile, sure, keep it.

lib/msp432/e4/Makefile
```diff
@@ -0,0 +1,47 @@
...
+PREFIX	        ?= arm-none-eabi
+
+CC		= $(PREFIX)-gcc
+AR		= $(PREFIX)-ar
+TGT_CFLAGS	= -Os \
+		  -Wall -Wextra -Wimplicit-function-declaration \
+		  -Wredundant-decls -Wmissing-prototypes -Wstrict-prototypes \
+		  -Wundef -Wshadow \
+		  -I../../../include -fno-common \
+		  -mcpu=cortex-m4 -mthumb $(FP_FLAGS) \
```
karlp: replace the cpu and thumb bits with $(ARCH_FLAGS)

lib/msp432/e4/Makefile
```diff
@@ -0,0 +1,47 @@
...
+		  -Wstrict-prototypes \
+		  -ffunction-sections -fdata-sections -MD -DSTM32F4
```
karlp: you've got the wrong define here, all your includes will fail.  (Is this code the version you're actually working with?)

lib/msp432/e4/systemcontrol.c
Misaka0x2730: During development, I relied heavily on systemcontrol from lm4f. So, I want keep it.

lib/msp432/e4/Makefile
Misaka0x2730: Oops, many thanks! I will fix it. 
Unfortunately, I'm not very good at makefile, but I'll try!

Misaka0x2730:
@karlp Yes, I'm still working on MSP432. Tonight I will fix everything.

Misaka0x2730:
@karlp  
> you've got the wrong define here, all your includes will fail.

Fixed.

karlp:
merged thanks!

PR #947, opened 2018-07-22
Add comparator register definitions for STM32F3

markus-k:
This pull request adds register definitions for comparators on STM32F3.

Some bits in the CSR register have a different meaning depending on the comparator used (e. g. blanking has different timers/OCs). I added a simple macro to directly select a value for those bits which shifts them to the correct position.

For example:
````C
    COMP_CSR(COMP7) |= COMP_CSR_BLANKING(2) | COMP_CSR_EN;
````

### DIFF
--- a/include/libopencm3/stm32/comparator.h
+++ b/include/libopencm3/stm32/comparator.h
@@ -22,7 +22,8 @@
 
 #if defined(STM32F0)
 #       include <libopencm3/stm32/f0/comparator.h>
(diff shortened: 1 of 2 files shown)


karlp:
I've pulled this locally and converted to the normal unshifted field style.

karlp:
sorry for not closing this earlier, it merged in 49a8c041  and 0d0f59d8 thanks!

PR #952, opened 2018-08-07
Typo fix in STM32F4 rcc.h

Suyooo:
Just a small typo I came across while trying to get MCO to work on my board. The define is not used in any other files as far as I can tell, but of course applications might break if they use the misspelt variant.


karlp:
merged in 8b1ac585 thanks!

PR #958, opened 2018-08-26
Added USB OTG support for F7

matthewlai:
Verified that the CDC ACM peripheral example from F4 still works.

### DIFF
--- a/include/libopencm3/usb/dwc/otg_fs.h
+++ b/include/libopencm3/usb/dwc/otg_fs.h
@@ -28,7 +28,7 @@
 #include <libopencm3/usb/dwc/otg_common.h>
 
 /* Memory map is required for USB_OTG_FS_BASE address */
-#if defined(STM32F1) || defined(STM32F2) || defined(STM32F4)
+#if defined(STM32F1) || defined(STM32F2) || defined(STM32F4) || defined(STM32F7)
 #	include <libopencm3/stm32/memorymap.h>
 #elif defined(EFM32HG)
 #	include <libopencm3/efm32/memorymap.h>
--- a/include/libopencm3/usb/dwc/otg_hs.h
+++ b/include/libopencm3/usb/dwc/otg_hs.h
@@ -28,7 +28,7 @@
 #include <libopencm3/usb/dwc/otg_common.h>
 
 /* Memory map is required for USB_OTG_HS_BASE address */
-#if defined(STM32F2) || defined(STM32F4)
+#if defined(STM32F2) || defined(STM32F4) || defined(STM32F7)
 #	include <libopencm3/stm32/memorymap.h>
 #else
 #	error "device family not supported by dwc/otg_hs."
--- a/lib/stm32/f7/Makefile
+++ b/lib/stm32/f7/Makefile
@@ -52,6 +52,9 @@ OBJS		+= rng_common_v1.o
 
 OBJS		+= usart_common_all.o usart_common_v2.o
 
+OBJS    += usb.o usb_standard.o usb_control.o usb_dwc_common.o \
+          usb_f107.o usb_f207.o usb_msc.o
+
 VPATH += ../../usb:../:../../cm3:../common
 VPATH += ../../ethernet
 


karlp:
soemthing's _reallllly_ weird then.  I've had this locally for ages, along with a bunch of other f7 work to enable all the clocks, but it never enumerates for me.  What else are you carrying locally?  can you share your "cdc acm f4 example" ?

matthewlai:
I am using the unmodified example here (besides using "rcc_hse_12mhz_3v3[RCC_CLOCK_3V3_120MHZ]" instead for clock settings): libopencm3-examples/examples/stm32/f4/stm32f4-discovery/usb_cdcacm, with my libopencm3, which does have quite a few changes (mostly all my unmerged PRs), but it's not clear what change caused it.

https://github.com/matthewlai/libopencm3

However, I noticed that after enumeration, it seems to miss a bunch of packets before it starts echoing. I am investigating that.

matthewlai:
Just verified the behaviour is the same on stm32f4 discovery, so this is not specific to F7. F7 seems to be working as well as F4. Still don't know why it ignores a bunch of input in the beginning.

karlp:
What board are you using?  My f746 disco will enumerate with cubemx software, but not with my own.  My code is https://github.com/karlp/libopencm3/blob/wip/dev-2018-9/tests/gadget-zero/main-stm32f746-disco.c

karlp:
Unrelated to this, but what other PRs do you have outstanding? I thought it was just the one with different load addresses left?

matthewlai:
I am using my own board with STM32F765ZGT6 (not the 745 specified in the schematics), schematics here: https://matthewlai.ca/d/8561_ostrich.pdf

I did have to hack in Vbus -> GPIO A9.

matthewlai:
Yeah it does look like all my changes have been merged except for this one: https://github.com/matthewlai/libopencm3/commit/c483cd6cdcace3ef31fc2dae9697db24372fbe8f

But I just tried removing it and it still works, so it's not that.

matthewlai:
Toolchain maybe? I am using gcc-arm-none-eabi-7-2018-q2-update on Linux.

matthewlai:
Do you know at what stage of enumeration does it fail? Before the vbus hack mine wasn't going out of USB reset, and after that I had some trouble with the chip not responding to GET CONFIGURATION because I wasn't zeroing the internal fields of some of the descriptor structs, but that shouldn't be a problem if you are initializing using designated initializers in C.

I used wireshark to capture USB traffic.

matthewlai:
I discovered that my packet drop problem was due to ModemManager probing and poking the CDC-ACM port. Disabling it makes my chip work perfectly.

Leaving it here for the benefit of future visitors from Google: https://stackoverflow.com/questions/31849664/cdc-acm-initialization-takes-a-long-time

karlp:
So now I'm just back to why on earth this doesn't work on my f7 disco board :[  

(Honestly, I'm stunned you made it this far without noticing modem manager's intrusions)

RabbeS:
Hi guys,

I just looked a little bit over the mamory map. 
However, I noticed at the Boundary addresses, that the USB OTG HS starts with 0x4004 0000.
But this isn't registered as a PERIPH_BASE. Doesn't cause that any problems here?

Best regards,
Stephan

matthewlai:
I updated my libopencm3 and my USB CDC ACM stopped working, too. I did some bisecting.

This commit is the culprit: https://github.com/libopencm3/libopencm3/commit/26ab78a710b4cb7abcace769768c4781332b17ba

Unfortunately I don't know enough about USB to know why.

Novakov:
One of the culpits is `else if` in line 400 in `usb_dwc_common.c`. If SETUP packet is received it is read but callbacked in never fired. After splitting it into two `if` statements my device started to be recognized(-ish).

But...

Device was not recognized correctly and Linux was logging:
```
new full-speed USB device number 85 using dwc_otg
usb 1-1.3: config index 0 descriptor too short (expected 32, got 9)
usb 1-1.3: config 1 has 0 interfaces, different from the descriptor's value: 1
```
I tried to trace why only 9 bytes were sent but was unlucky and stepping through the code didn't show anything suspicious. 

For anyone going through the problem: the master branch (commit 6de149c57860d9fcfd050468a70de8f44814ef30) in @matthewlai repository (https://github.com/matthewlai/libopencm3) works.

karlp:
fwiw that commit also is why f446 doesn't work with usb.  still, it was _fixing_ things for other targets then. reverting is not really a solution, we need a path forward that works for all targets (versions of dwc_otg)

matthewlai:
Can we do a conditional on dwc_otg version in the mean time?

karlp:
well, allegedly, we've had cases of the same version (as reported by the dwc registers) not working in the different implementation.  We _know_ that we are either breaking f4, or f7, so it's all suck, all the way down.....  (I'm not actively looking at this, fwiw)

karlp:

PR #961, opened 2018-08-29
stm32f7: added desig to Makefile

matthewlai:
desig_get_unique_id_as_dfu() returns same ID as DFU.


karlp:
Comment ere doesn't match contents?  Are you referring to some other dfu code?

karlp:
added in de538c4b

matthewlai:
Oh yeah I just meant that's how I tested it.

PR #963, opened 2018-09-09
Add timer to STM32F7

Hoeze:
Please tell us if something is missing :)


karlp:
If you've tested this, I've already got it locally and will just merge that instead. I've just not got aroudn to testing my own locally :)

karlp:
merged equivalent in 9ddafa62

PR #970, opened 2018-09-23
Add GPIO support for MSP432.

Misaka0x2730:
Difference from LM4F:
1. Added full-range pin drive strength (2, 4, 6, 8, 10, 12 mA).
2. Added AF defines for all pins. 
3. And many more things...


include/libopencm3/msp432/e4/gpio.h
```diff
@@ -0,0 +1,2524 @@
...
+/** GPIO Alternate Function 8 Identifier */
+#define GPIO_AF8            0x8
+/** GPIO Alternate Function 11 Identifier */
+#define GPIO_AF11           0xB
+/** GPIO Alternate Function 13 Identifier */
+#define GPIO_AF13           0xD
+/** GPIO Alternate Function 14 Identifier */
+#define GPIO_AF14           0xE
+/** GPIO Alternate Function 15 Identifier */
+#define GPIO_AF15           0xF
```
karlp: Can we just drop these? the af number is just ... the af number.

karlp:
This all looks really good, can it just be in one commmit? are there any portions you want to land separately?

include/libopencm3/msp432/e4/gpio.h
Misaka0x2730: Yes, but i think it makes it easier to read.

Misaka0x2730:
Okay, no problem.

PR #972, opened 2018-09-28
Debian

rufelo:


debian/control
```diff
@@ -0,0 +1,21 @@
...
+Package: libopencm3
+Section: fixme
+Architecture: all
+Depends: ${misc:Depends}
+Description: firmware library for ARM Cortex-M3 microcontrollers
+ The libopencm3 library provides the base code for running software without an
+ operating system on ARM Cortex-M3 microcontrollers, and contains both system
+ initialization routines and periphery implementation.
+ .
+ It supports the STM32, TX03, SAM3U and LPC1000 families of chips.
```
karlp: you probably want to update this, where did it come from?

karlp:
I'm not terrrribly against this sort of thing, but I reallllly wish in general that distros maintained distro things, rather than expecting packages to maintain things for all distros. Regardless, you're gonna want to squish this to a single commit.

More philosophically, do you see any particular blocking issues outstanding?  What would tagged "release" points change for you, if at all?

PR #973, opened 2018-09-29
Set PLL source in rcc_clock_setup_pll() for STM32L0

jarodgers:
I discovered this bug when trying to use HSE as PLL source on the STM32L0. In the function rcc_clock_setup_pll() in l0/rcc.c, when the PLL source is not set correctly, it will get stuck in an endless loop waiting for the PLLON flag.


karlp:
merged thanks!

PR #977, opened 2018-10-05
stm32: fix nvic channels name to match dma1/2 on stm32f09x

grevaillot:
Fix irq list to match stm32f09x additional dma controller/channels.


include/libopencm3/stm32/f0/irq.json
```diff
@@ -10,8 +10,8 @@
         "exti4_15", 
         "tsc", 
         "dma1_channel1", 
-        "dma1_channel2_3", 
-        "dma1_channel4_5", 
+        "dma1_channel2_3_dma2_channel1_2", 
+        "dma1_channel4_5_6_7_dma2_channel3_4_5",
```
karlp: those are horrific.  what does cmsis call these?

grevaillot: Indeed, and, yes, good point - I must admit i never looked at cmsis stuff..

Apparently, they have multiple aliases apparently.

> #define DMA1_Channel4_5_6_7_IRQHandler DMA1_Ch4_7_DMA2_Ch3_5_IRQHandler
> #define DMA1_Channel4_5_IRQHandler     DMA1_Ch4_7_DMA2_Ch3_5_IRQHandler

I guess aligning to the "largest" definition, following the st naming would be the way to go, Ie _dma1_channel4_7_dma2_channel_3_5_ ?

grevaillot:
ok, i fixed that commit to match cmsis "nicer" naming.

karlp:
merged in 708fe151 thanks

PR #979, opened 2018-10-06
Add GPIO support for MSP432

Misaka0x2730:
Difference from LM4F:

Added full-range pin drive strength (2, 4, 6, 8, 10, 12 mA).
Added AF defines for all pins.
And many more things...

This pull request is copy of #970, but in one commit.


Misaka0x2730:
Oops, I'm really sorry, it's my mistake. 
I will definitely try in the future.

karlp:
merged thanks!

PR #981, opened 2018-10-11
common: add assembler support

alexkazik:
When using gcc as an assembler:
- the C/C++ declarations are skipped
- the macros are adapted (when needed)

The definition `__ASSEMBLER__` is automatically added when gcc is invoked as a assembler, e.g. `gcc test.S`.

Once this is merged I'll update the Makefile and add a small example to the template.

### DIFF
--- a/include/libopencm3/cm3/common.h
+++ b/include/libopencm3/cm3/common.h
@@ -20,6 +20,8 @@
 #ifndef LIBOPENCM3_CM3_COMMON_H
 #define LIBOPENCM3_CM3_COMMON_H
 
+#ifndef __ASSEMBLER__
+
 #include <stdint.h>
 #include <stdbool.h>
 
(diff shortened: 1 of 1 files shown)


include/libopencm3/cm3/common.h
```diff
@@ -59,6 +61,29 @@
...
+#define MMIO16(addr) (addr)
+#define MMIO32(addr) (addr)
+#define MMIO64(addr) (addr)
+
+/* Generic bit-band I/O accessor functions */
+#define BBIO_SRAM(addr, bit) \
+	(((addr) & 0x0FFFFF) * 32 + 0x22000000 + (bit) * 4)
+
+#define BBIO_PERIPH(addr, bit) \
+	(((addr) & 0x0FFFFF) * 32 + 0x42000000 + (bit) * 4)
```
karlp: completely out of scope of this commit

include/libopencm3/cm3/common.h
```diff
@@ -59,6 +61,29 @@
...
+
+/* Skip all C/C++ declarations */
+#define BEGIN_DECLS .if 0
+#define END_DECLS .endif
+
+/* All addresses are just numbers */
+#define MMIO8(addr) (addr)
+#define MMIO16(addr) (addr)
+#define MMIO32(addr) (addr)
+#define MMIO64(addr) (addr)
```
karlp: completely out of scope of this commit

include/libopencm3/cm3/common.h
alexkazik: Sorry, but why? The BBIO_* C/C++ macros are just above and they do the same. But I can remove them from the commit if that helps.

karlp:
Can you prepare the other portions of this that you need too?  So we can see the whole thing you're doing, not just parts like this? Just refer to them in each one.

karlp:
merged somethign equivalent in 4edba311 thanks

PR #984, opened 2018-10-21
Repaired spi for the f7

Hoeze:
Fixes some SPI bugs:
- change from `spi_common_v1_fr.hf` to `spi_common_v2.h`
- add missing `SPI_CR2` definitions

### DIFF
--- a/include/libopencm3/stm32/f7/spi.h
+++ b/include/libopencm3/stm32/f7/spi.h
@@ -31,7 +31,14 @@ LGPL License Terms @ref lgpl_license
 #ifndef LIBOPENCM3_SPI_H
 #define LIBOPENCM3_SPI_H
 
-#include <libopencm3/stm32/common/spi_common_v1_frf.h>
+#include <libopencm3/stm32/common/spi_common_v2.h>
+
+/* --- SPI_CR2 values ------------------------------------------------------ */
+
+/* LDMA_TX: Last DMA transfer for transmission */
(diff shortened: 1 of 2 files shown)


Hoeze:
@karlp yep, was my fault :)

Hoeze:
Thanks for including my bug fix, but unfortunately you forgot some crucial lines in your commits (did you see the includes?).
Could you maybe instead merge my pull request?

In general, simply re-creating bug fixes is not very nice to your contributors.
Merging the pull request of someone else is a way to say "Thank you, I appreciate your contribution" :wink:

Also, if you intend to include the changes of a contributor, please also accept his commits.
If you are not happy with some aspects of a pull request, just tell him; your contributor will fix it.
Therefore, you save yourself and others time with coding and testing.
Otherwise, you can still commit your own changes in addition to a pull request.

At the very last, if you accept my pull request, you save me time to resolve double-merges :)

karlp:
You're absolutely right.  I do always try and use contributors commits to give them credit as much as possible.  This one was very small, but I wanted to move the frame format bit definitions to the v2 header, rather than in the f7 header, and given how small it was, (and that I was splitting in two even smaller commits!) I didn't feel it was something I should chase you to change. (For better or worse, it has not been a happy truth that contributors always fix things)  You're absolutely right that I missed the include line, that was poor form on my part, and was fixed in the next commit 3af05fb86

Hoeze:
Thank you very much for fixing it :)

PR #985, opened 2018-10-21
CAN improved RX fifo handling, ADC bugfix injected channel reading

jsphuebner:
Made adc_read_injected() return signed int because result can become negative
Added function can_fifo_pending() - number of pending RX messages
can_receive() returns number of pending messages prior to release

### DIFF
--- a/include/libopencm3/stm32/can.h
+++ b/include/libopencm3/stm32/can.h
@@ -668,12 +668,13 @@ void can_disable_irq(uint32_t canport, uint32_t irq);
 
 int can_transmit(uint32_t canport, uint32_t id, bool ext, bool rtr,
 		 uint8_t length, uint8_t *data);
-void can_receive(uint32_t canport, uint8_t fifo, bool release, uint32_t *id,
+uint32_t can_receive(uint32_t canport, uint8_t fifo, bool release, uint32_t *id,
 		 bool *ext, bool *rtr, uint8_t *fmi, uint8_t *length,
 		 uint8_t *data, uint16_t *timestamp);
 
 void can_fifo_release(uint32_t canport, uint8_t fifo);
 bool can_available_mailbox(uint32_t canport);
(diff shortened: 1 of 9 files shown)


karlp:
Please try and have one commit per change.    The ADC injected changes I've commented on before, you would at _least_ need to update the documentation to go with it.

jsphuebner:
Documentation updated
One commit per change next time, sorry about that

VictorLamoine:
:ping_pong: @karlp anything against merging this?

karlp:
thhe can and ADC changes need to be split.  teh adc change is still per https://github.com/libopencm3/libopencm3/pull/285 not perfect in my mind.  There's a claim that "it's tested" but not what those test conditions are or were.  (did they test all the bit sizes? the alignments?)   so no, I'm definitely not merging this as is.  if the can change is split out I'll at least look at it.

jsphuebner:
How do I split those commits in the aftermath?
ADC change is tested right aligned only and with an offset that results in negative results.

VictorLamoine:
- `git checkout can_and_adc_fix`
- `git reset HEAD~2` undo the last two commits but keeps the changes
- `git add include/libopencm3/stm32/can.h lib/stm32/can.c` add modified CAN files to the stash
- `git commit -m "Improve CAN Rx fifo handling"` commit the CAN files
- `git add include/libopencm3/stm32/common/adc_common_v1.h lib/stm32/common/adc_common_v1.c` add modified ADC files to the stash
- `git commit -m "Fix ADC injected channel reading"`` commit ADC files

Ideally when you work on two different features (ADC has nothing to do with CAN) use two branches to keep your modifications independent.

jsphuebner:
Thanks Victor. Lets see how that goes

esden (approved):
LGTM, thank you for the contribution.

PR #990, opened 2018-11-06
stm32f09: add register definitions for DMA2

aklomp:
Add missing register definitions for the DMA2 controller on the STM32F091 and similar. Code was tested with that core.

Source for the address values is ST Reference manual RM0091, which covers the STM32F0x1/STM32F0x2/STM32F0x8.

### DIFF
--- a/include/libopencm3/stm32/common/dma_common_l1f013.h
+++ b/include/libopencm3/stm32/common/dma_common_l1f013.h
@@ -138,11 +138,6 @@ specific memorymap.h header before including this header file.*/
 #define DMA2_CMAR4			DMA2_CMAR(DMA_CHANNEL4)
 #define DMA2_CMAR5			DMA2_CMAR(DMA_CHANNEL5)
 
-/* DMA channel selection register (DMAx_CSELR) */
-#define DMA_CSELR(dma_base)		MMIO32((dma_base) + 0xA8)
-#define DMA1_CSELR			DMA_CSELR(DMA1)
-#define DMA2_CSELR			DMA_CSELR(DMA2)
-
 /* --- DMA_ISR values ------------------------------------------------------ */
 
 /* --- DMA Interrupt Flag offset values ------------------------------------- */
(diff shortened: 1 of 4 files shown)


include/libopencm3/stm32/common/dma_common_l1f013.h
```diff
@@ -138,6 +138,11 @@ specific memorymap.h header before including this header file.*/
 #define DMA2_CMAR4			DMA2_CMAR(DMA_CHANNEL4)
 #define DMA2_CMAR5			DMA2_CMAR(DMA_CHANNEL5)
 
+/* DMA channel selection register (DMAx_CSELR) */
+#define DMA_CSELR(dma_base)		MMIO32((dma_base) + 0xA8)
+#define DMA1_CSELR			DMA_CSELR(DMA1)
+#define DMA2_CSELR			DMA_CSELR(DMA2)
+
```
karlp: These should be in the f0 file, not the l1f013 file, this register doesn't exist on l1, f1, nor f3

aklomp: Thanks for the correction. I pushed a fixup commit (to be merged with the previous commit), but I can also force-push the branch if you prefer.

karlp: I'm happy to squish when I merge, no problem

include/libopencm3/stm32/f0/rcc.h
```diff
@@ -244,6 +244,7 @@ Control</b>
 #define RCC_AHBENR_CRCEN			(1 << 6)
 #define RCC_AHBENR_FLTFEN			(1 << 4)
 #define RCC_AHBENR_SRAMEN			(1 << 2)
+#define RCC_AHBENR_DMA2EN			(1 << 1)
```
karlp: You also need to add it to enums  down at https://github.com/libopencm3/libopencm3/blob/a9c0008290ae7026fd4bc5c662966c6d88e80697/include/libopencm3/stm32/f0/rcc.h#L413 please

aklomp: Isn't it [already there](https://github.com/libopencm3/libopencm3/blob/a9c0008290ae7026fd4bc5c662966c6d88e80697/include/libopencm3/stm32/f0/rcc.h#L416)?

karlp: true, looks like someone added that, and just didn't add the rest of it in 54c6c833f ok

karlp:
merged thanks!

aklomp:
Thanks for merging, but the fixup was not squashed. Probably not a very big deal.

karlp:
eh, my bad, oh well :|

PR #991, opened 2018-11-07
gadget0 : fix deprecation warning by updating openocd cfg to use "stlink.cfg"

fenugrec:
Since openocd 31c58c (committed 2017), all stlink configuration files are merged into a single
"stlink.cfg" file that covers all versions of stlink hardware.


karlp:
iirc, that change is _after_ the 0.10 release, so there's no "release" version of openocd that contains that commit yet, even though it's quite a long time ago now.

karlp:
I'll land this as soon as openocd makes a new release.

fenugrec:
nice ! this PR doesn't even smell like compost yet ! hehe

PR #992, opened 2018-11-09
stm32: pll: disable before configuring

aklomp:
This bug was found and fixed by a coworker. The clock configuration function for the STM32F427 failed to reconfigure the PLL correctly, because it did not disable the PLL before changing its configuration.

This has only been tested on the STM32F427. I ported the fix to the F2 and the F7 as they seem to share the same logic, but I admit that I haven't tested on those.

### DIFF
--- a/lib/stm32/f2/rcc.c
+++ b/lib/stm32/f2/rcc.c
@@ -355,6 +355,10 @@ void rcc_clock_setup_hse_3v3(const struct rcc_clock_scale *clock)
 	rcc_set_ppre1(clock->ppre1);
 	rcc_set_ppre2(clock->ppre2);
 
+	/* Disable PLL oscillator before changing its configuration. */
+	rcc_osc_off(RCC_PLL);
+
+	/* Configure the PLL oscillator. */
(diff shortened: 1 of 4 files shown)


karlp:
do you have more? just curious whether I can review a few in one go, or just one new one as soon as I merge the last :)

karlp:
Coul you add it to L1 too please? all the rest look to have it covered, or not yet implemented.

aklomp:
Yeah, I have a few more pull requests coming, but not that many, maybe two or three. They're part of a company initiative to upstream the fixes we made in our private repo for our STM32F09 and STM32F42 targets.

The remaining pull requests should be fairly standalone, so  if you prefer I can push all of them at once instead of drip-feeding them. I'll also fixup this pull request to include the L1. It will have to wait till Monday though :)

karlp:
No stress, just so I know what's going on :)  glad to see you guys pushing changes back up.

aklomp:
I added the L1 and force-pushed the branch.

karlp:
thanks, merged!

PR #995, opened 2018-11-22
add nRF5 series to libopencm3: add nrf52 to doc, include, lib and modify related files for nrf52

vluongle:
Add the nRF52 series to the libopencm3:
- A new folder "nrf5" with sub-folder "2" will be added to doc, lib and include.
- Some related files will be modified for integrating new series: devices.data, makefiles, nvic and vector

Once this is merged, Some examples will be pulled.


include/libopencm3/nrf5/common/temp_common.h
```diff
@@ -0,0 +1,142 @@
...
+#define TEMP_TASKS_START                 MMIO32((TEMP_BASE) + 0x000)
+#define TEMP_TASKS_STOP                  MMIO32((TEMP_BASE) + 0x004)
+/* Events Registers */
+#define TEMP_EVENTS_DATARDY              MMIO32((TEMP_BASE) + 0x100)
+/* Registers */
+#define TEMP_INTENSET                    MMIO32((TEMP_BASE) + 0x304)
+#define TEMP_INTENCLR                    MMIO32((TEMP_BASE) + 0x308)
+#define TEMP_TEMP                        MMIO32((TEMP_BASE) + 0x508)
+#define TEMP_A                           PTRIO32((TEMP_BASE) + 0x520)
+#define TEMP_A0                          MMIO32((TEMP_BASE) + 0x520)
```
karlp: this have often just been defined as TEMPA *TEMPA0, is there any real need for the PTRIO thing? or TEMP_A itself?

include/libopencm3/nrf5/2/memorymap.h
```diff
@@ -0,0 +1,107 @@
...
+
+/* Memory map for all busses */
+#define PERIPH_BASE                0x40000000U
+#define PERIPH_BASE_APB            0x40000000U
+#define PERIPH_BASE_AHB            0x50000000U
+
+/* Register boundary addresses */
+
+/* APB */
+#define CLOCK_BASE                (PERIPH_BASE_APB + 0x0000)
```
karlp: Are these really all meant to have the ~same address?

include/libopencm3/nrf5/common/clock_common.h
```diff
@@ -0,0 +1,224 @@
...
+@ingroup clock_defines
+@{*/
+#define CLOCK_INTENSET_HFCLKSTARTED                (1 << 0)
+#define CLOCK_INTENSET_LFCLKSTARTED                (1 << 1)
+#define CLOCK_INTENSET_DONE                        (1 << 3)
+#define CLOCK_INTENSET_CTTO                        (1 << 4)
+/**@}*/
+
+/* --- CLOCK_INTENCLR values ----------------------------------------------- */
+/** @defgroup clock_intenset Disable interrupt
```
karlp: duplicate group here.

karlp: (iirc, you don't need the ingroup, that's automatic because of where you are in the file)

include/libopencm3/nrf5/common/clock_common.h
```diff
@@ -0,0 +1,224 @@
...
+@{*/
+#define CLOCK_LFCLKSTAT_SRC_RC                     (0 << 0)
+#define CLOCK_LFCLKSTAT_SRC_XTAL                   (1 << 0)
+#define CLOCK_LFCLKSTAT_SRC_Xtal                   (1 << 0)
+#define CLOCK_LFCLKSTAT_SRC_SYNTH                  (0x2 << 0)
+#define CLOCK_LFCLKSTAT_SRC_Synth                  (0x2 << 0)
+#define CLOCK_LFCLKSTAT_STATE_NOTRUNNING           (0 << 16)
+#define CLOCK_LFCLKSTAT_STATE_NotRunning           (0 << 16)
+#define CLOCK_LFCLKSTAT_STATE_RUNNING              (1 << 16)
+#define CLOCK_LFCLKSTAT_STATE_Running              (1 << 16)
```
karlp: Why do you need both of these?  I'd _really_ rather not have both, but if this has a good reason, please make the defines more like
```
#define WEIRD_NAME ORIGINAL_NAME
```
rather than duplicating the content

include/libopencm3/nrf5/common/clock_common.h
```diff
@@ -0,0 +1,224 @@
...
+#define RC                                         (0 << 0)
+#define CLOCK_LFCLKSRC_SRC_RC                      (0 << 0)
+#define XTAL                                       (1 << 0)
+#define CLOCK_LFCLKSRC_SRC_XTAL                    (1 << 0)
+#define SYNTH                                      (0x2 << 0)
+#define CLOCK_LFCLKSRC_SRC_SYNTH                   (0x2 << 0)
+#define CLOCK_LFCLKSRC_BYPASS_ENABLED              (1 << 16)
+#define CLOCK_LFCLKSRC_BYPASS_Enabled              (1 << 16)
+#define CLOCK_LFCLKSRC_EXTERNAL_ENABLED            (1 << 17)
+#define CLOCK_LFCLKSRC_EXTERNAL_Enabled            (1 << 17)
```
karlp: Same re names

include/libopencm3/nrf5/common/clock_common.h
```diff
@@ -0,0 +1,224 @@
...
+@{*/
+#define CLOCK_TRACECONFIG_TRACEPORTSPEED_32MHZ        (0 << 0) //TRACECLK = 16 MHz
+#define CLOCK_TRACECONFIG_TRACEPORTSPEED_16MHZ        (1 << 0) //TRACECLK = 8 MHz
+#define CLOCK_TRACECONFIG_TRACEPORTSPEED_8MHZ         (0x2 << 0) //TRACECLK = 4 MHz
+#define CLOCK_TRACECONFIG_TRACEPORTSPEED_4MHZ         (0x3 << 0) //TRACECLK = 2 MHz
+//gpio: GPIOs multiplexed onto all trace-pins
+#define CLOCK_TRACECONFIG_TRACEMUX_GPIO               (0 << 16) 
+//Serial: SWO multiplexed onto P0.18, GPIO multiplexed onto other trace-pins
+#define CLOCK_TRACECONFIG_TRACEMUX_SERIAL             (1 << 16) 
+//Parallel: TRACECLK and TRACEDATA multiplexed onto P0.20, P0.18, P0.16, P0.15, P0.14
```
karlp: use /** here, so this useful documentation makes it into the documentation.
(same fo the traceclk / traceportspeed docs above)

include/libopencm3/nrf5/common/gpio_common.h
```diff
@@ -0,0 +1,966 @@
...
+#define GPIO_OUT_PIN29_HIGH              (1 << 29)
+#define GPIO_OUT_PIN29_High              (1 << 29)
+#define GPIO_OUT_PIN30_LOW               (0 << 30)
+#define GPIO_OUT_PIN30_Low               (0 << 30)
+#define GPIO_OUT_PIN30_HIGH              (1 << 30)
+#define GPIO_OUT_PIN30_High              (1 << 30)
+#define GPIO_OUT_PIN31_LOW               (0 << 31)
+#define GPIO_OUT_PIN31_Low               (0 << 31)
+#define GPIO_OUT_PIN31_HIGH              (1 << 31)
+#define GPIO_OUT_PIN31_High              (1 << 31)
```
karlp: I don't see any value in these defines personally.  
They certainly don't need to be duplicated for GPIO_OUTSET below.

include/libopencm3/nrf5/common/gpio_common.h
```diff
@@ -0,0 +1,966 @@
...
+#define GPIO_LATCH_PIN29_Latched                  (1 << 29)
+#define GPIO_LATCH_PIN30_NOTLATCHED               (0 << 30)
+#define GPIO_LATCH_PIN30_NotLatched               (0 << 30)
+#define GPIO_LATCH_PIN30_LATCHED                  (1 << 30)
+#define GPIO_LATCH_PIN30_Latched                  (1 << 30)
+#define GPIO_LATCH_PIN31_NOTLATCHED               (0 << 31)
+#define GPIO_LATCH_PIN31_NotLatched               (0 << 31)
+#define GPIO_LATCH_PIN31_LATCHED                  (1 << 31)
+#define GPIO_LATCH_PIN31_Latched                  (1 << 31)
+/**@}*/
```
karlp: so we're at ~800 lines of "on/off << pin number"  Please no.

include/libopencm3/nrf5/common/gpio_common.h
```diff
@@ -0,0 +1,966 @@
...
+#define GPIO_DETECTMODE_DETECTMODE_Default        (0)
+#define GPIO_DETECTMODE_DETECTMODE_LDETECT        (1)
+/**@}*/
+
+/* --- GPIO_PIN_CNF[n] values ---------------------------------------------- */
+
+/** @defgroup gpio_pin_cnf[n] Configuration of GPIO pins
+@ingroup gpio_defines
+@{*/
+/* Pin direction */
```
karlp: please avoid these short defines here, they're polluting namespaces.

include/libopencm3/nrf5/common/gpio_common.h
```diff
@@ -0,0 +1,966 @@
...
+#define GPIO25                (1 << 25)
+#define GPIO26                (1 << 26)
+#define GPIO27                (1 << 27)
+#define GPIO28                (1 << 28)
+#define GPIO29                (1 << 29)
+#define GPIO30                (1 << 30)
+#define GPIO31                (1 << 31)
+#define GPIO_ALL              0xffffffff
+/*Pin number definitions for convenience*/
+#define pin(n)            n
```
karlp: no way, this is crazy.

include/libopencm3/nrf5/common/ppi_common.h
```diff
@@ -0,0 +1,736 @@
...
+/**@}*/
+
+/* --- Function prototypes ------------------------------------------------- */
+
+BEGIN_DECLS
+
+//funktion();
+//funktion();
+//funktion();
+//funktion();
```
karlp: junk all this dead comments.

include/libopencm3/nrf5/common/pwm_common.h
```diff
@@ -0,0 +1,527 @@
+/** @addtogroup timer_defines
```
karlp: wrong group again, please check how the generated doxygen looks and check the rest of these yourself.

include/libopencm3/nrf5/common/pwm_common.h
```diff
@@ -0,0 +1,527 @@
...
+#define PWM_PRESCALER_PRESCALER_DIV_4                (2)
+#define DIV_8                                        (3)
+#define PWM_PRESCALER_PRESCALER_DIV_8                (3)
+#define DIV_16                                       (4)
+#define PWM_PRESCALER_PRESCALER_DIV_16               (4)
+#define DIV_32                                       (5)
+#define PWM_PRESCALER_PRESCALER_DIV_32               (5)
+#define DIV_64                                       (6)
+#define PWM_PRESCALER_PRESCALER_DIV_64               (6)
+#define DIV_128                                      (7)
```
karlp: no, none of these DIV_ defines please.

karlp:
I've stopped part way through the PWM file, I won't look further in this pass.  It looks like you're adding too many ways of doign the same thing.  Just one is enough.  (And that should be the define way, not the struct way, just for consistency with the rest of the project)

Watch the namespaces, don't define all the duplicates with case changes.

You can feel very welcome to suggest examples PR that relies on another PR here, that's still a very useful way of seeing things in operation and even testing them.

lib/nrf5/common/uarte_common.c
```diff
@@ -0,0 +1,429 @@
...
+@param[in] uarte unsigned 32 bit. UARTE block register address base @ref
+uarte_reg_base
+@param[in] event unsigned 32 bit. predetermined event names @ref 
+uarte_events_interrupt
+@returns boolean
+*/
+
+bool uarte_get_event(uint32_t uarte, char event)
+{
+	if(event & CTS){return (UARTE_EVENTS_CTS(uarte) != 0);}
```
karlp: stylecheck here.

lib/nrf5/common/uart_common.c
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**@{*/
+
+#include <libopencm3/nrf5/uart.h>
+
+/*---------------------------------------------------------------------------*/
```
karlp: please don't include empty files like this, there's no point, just leave themuntil you need them.

lib/nrf5/common/temp_common.c
```diff
@@ -0,0 +1,41 @@
+/** @addtogroup temp_file
+
+@author @htmlonly &copy; @endhtmlonly 2018 
+Gerrit Maus <funk@maus.xyz>
+Luong Le <novercy@live.com>
+
+This library supports the TEMP in the NRF52 series.
```
karlp: this is the space where you have room to write what TEMP actually is.  otherwise you might as well have just left this out.

lib/nrf5/common/spim_common.c
```diff
@@ -0,0 +1,41 @@
+/** @addtogroup spim_file
+
+@author @htmlonly &copy; @endhtmlonly 2018 
+Gerrit Maus <funk@maus.xyz>
+Luong Le <novercy@live.com>
+
+This library supports the SPIM in the NRF52 series.
```
karlp: you have more room here, use it. don't just write comments that _appear_ to be commnets, but don't actually say anything.

lib/nrf5/common/saadc_common.c
```diff
@@ -0,0 +1,389 @@
...
+
+uint32_t saadc_sample_end(void)
+{
+	return SAADC_EVENTS_END;
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief SAADC ADC status.
+
+Return 1 if ADC is budy, conversion in progress.
```
karlp: typo

lib/nrf5/common/saadc_common.c
```diff
@@ -0,0 +1,389 @@
...
+{
+	SAADC_EVENTS_CH_LIMITH[channel] = 0;
+	SAADC_EVENTS_CH_LIMITL[channel] = 0;
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief SAADC Sampling process status.
+
+Return 1 if the sampling process is done and the ADC has filled up the Result
+buffer.
```
karlp: and what if it hasn't?  Is this just a bool?

vluongle:
Thanks @karlp for the review, 
Some of duplicate definitions were used for our private projects more flexible. Some others fit the database style with struct, so possibly I will keep this style. Unnecessary lines and some others duplicate definitions will be modified to looks similar to the libopencm3 style.

include/libopencm3/nrf5/common/temp_common.h
vluongle: TEMP_A doesn't need PTRIO, just an option to shorten functions later, since offset of Register A1 and A0 has a gap of 4 Bytes, and so on. If libopencm3 uses only Definition, then TEMP_A will be removed.

include/libopencm3/nrf5/2/memorymap.h
vluongle: Yes, some peripherals have the same address

include/libopencm3/nrf5/common/clock_common.h
vluongle: The duplicate defines will be removed

include/libopencm3/nrf5/common/gpio_common.h
vluongle: We have defined so that we can stick to database to write a code without seeing header-file. They will be removed for libopencm3-users.

include/libopencm3/nrf5/common/ppi_common.h
```diff
@@ -0,0 +1,736 @@
...
+#define PPI_TASKS_CHG4_DIS              MMIO32((PPI_BASE) + 0x024)
+#define PPI_TASKS_CHG5_EN               MMIO32((PPI_BASE) + 0x028)
+#define PPI_TASKS_CHG5_DIS              MMIO32((PPI_BASE) + 0x02C)
+
+/* Registers */
+#define PPI_CHEN                        MMIO32((PPI_BASE) + 0x500)
+#define PPI_CHENSET                     MMIO32((PPI_BASE) + 0x504)
+#define PPI_CHENCLR                     MMIO32((PPI_BASE) + 0x508)
+
+struct PPI_CH_TYPE{
```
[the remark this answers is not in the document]
vluongle: the underline can be used normally

include/libopencm3/nrf5/common/ppi_common.h
```diff
@@ -0,0 +1,736 @@
...
+#define PPI_CHG0                        MMIO32((PPI_BASE) + 0x800)
+#define PPI_CHG1                        MMIO32((PPI_BASE) + 0x804)
+#define PPI_CHG2                        MMIO32((PPI_BASE) + 0x808)
+#define PPI_CHG3                        MMIO32((PPI_BASE) + 0x80C)
+#define PPI_CHG4                        MMIO32((PPI_BASE) + 0x810)
+#define PPI_CHG5                        MMIO32((PPI_BASE) + 0x814)
+
+struct PPI_FORK_TYPE{
+	uint32_t TEP;
+};
```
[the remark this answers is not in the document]
vluongle: This is for register PPI_FORK_TYPE[x].TEP and not useless

include/libopencm3/nrf5/common/ppi_common.h
```diff
@@ -0,0 +1,736 @@
...
+#define PPI_FORK      ((volatile struct PPI_FORK_TYPE*)((PPI_BASE) + 0x910))
```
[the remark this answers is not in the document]
vluongle: nrf5-family using too many register-addresses like this, I think using pointer is better than definition in this case and the database also uses style like this

include/libopencm3/nrf5/common/ppi_common.h
vluongle: They will be removed

lib/nrf5/common/uarte_common.c
vluongle: I dont get stylecheck here, but this may because of "return false". It will be removed.

lib/nrf5/common/saadc_common.c
vluongle: The EVENTS_END (and other events) is a register so I prefer to leave the return value as same as defined type, otherwise might got stylecheck.

funkmaus:
I agree with @karlp that there are some conventions which have to be applied in order to get this PR mergeable. I also agree that there is some work left making those library functions to be more libopencm3-style like. But of course we should be aware of the differences between the nrf5x-families and e.g. the stm32fx-families (Task/Event system, C-style register names (see below), EasyDMA, …).

My two cents about this `PTRIO32` thing: Nordic Semiconductor makes exhaustive use of C-style register names within its datasheet. For example, in order to select the positive Input pin of ADC channel 0 you have to write (according to the datasheet's name) into SAADC register `CH[0].PSELP`. It just would be fine if this naming convention was supported by libopencm3. Libopencm3's general register name conventions `PeripheralName_RegisterName` are still complied: That specific register may be addressed by `SAADC_CH[0].PSELP`.

PR #997, opened 2018-12-07
stm32f0: adc: API call to clear EOS flag

patrickyeon:
Save me from directly manipulating registers in my code.

### DIFF
--- a/lib/stm32/f0/adc.c
+++ b/lib/stm32/f0/adc.c
@@ -277,6 +277,16 @@ bool adc_get_eoc_sequence_flag(uint32_t adc)
 	return ADC_ISR(adc) & ADC_ISR_EOSEQ;
 }
 
+/*---------------------------------------------------------------------------*/
+/** @ brief ADC Clear Regular End-Of-Conversion Sequence Flag
+ *
+ * @param[in] adc Unsigned int32. ADC base address (@ref adc_reg_base)
+ */
+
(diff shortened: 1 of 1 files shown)


karlp:
seems perfectly reasonable.  merged.  Eventually, these functions should move to the -v2 periph apis I guess.

patrickyeon:
@karlp sorry, reviving an old convo. Is there an intro/sketch of what is a -v1 api and what is a -v2, as examples to pattern new code off. Even better, a quick rundown of the philosophy/style of the -v2 API?

I'm working on a couple projects that integrate F0's, so I can roll in bringing the libopencm3 API up to date with this effort, if you give me a little bit of guidance.

karlp:
it's not a v2 api, it's the implementation of the support for the "second" generation adc peripheral found in f0, f3, l0, l4 and possibly other newer parts.  So the idea is, "don't add code to f0's adc.c, that's actually common for all parts with that adc peripheral"

the -v2 naming is (at least to some) easier to follow than the continually renaming files of "common_f01234l123" sort of thing.  -v2-multi and v2-single for the adc is because that's the sort of definining variation of the peripherals.  

basically, ST re-uses the peripherals heavily, so whenever possible, write code that supports the _peripheral_ not the particular target.

patrickyeon:
Ahh ok, I misunderstood. I'll see if I can follow what's working well enough to help there, thx.

PR #923, opened 2018-05-22
USB MSC: fix for scsi command READ_FORMAT_CAPACITIES and INQUIRY SERIAL NUMBER

onborodin:
Fix the subj. Original patch get there https://github.com/thirdpin/libopencm3

Before patch work with the device is not possible.
FreeBSD11 dmesg ouput
----
umass0 on uhub0
umass0: <Black Sphere Technologies MSC Demo, class 0/0, rev 1.10/2.00, addr 6> on usbus0
ugen0.3: <Black Sphere Technologies MSC Demo> at usbus0 (disconnected)
umass0: at uhub0, port 2, addr 6 (disconnected)
(da0:umass-sim0:0:0:0): got CAM status 0x44
(da0:umass-sim0:0:0:0): fatal error, failed to attach to device
----

After patch:
ugen0.3: <Black Sphere Technologies MSC Demo> at usbus0 (disconnected)
umass0: at uhub0, port 2, addr 16 (disconnected)
da0 at umass-sim0 bus 0 scbus1 target 0 lun 0
da0: <VendorID ProductID 0.00> s/n DEMO detached

The device works fine:

# ls -l /media/RAMDISK/
total 128
-rwxr-xr-x  1 root  wheel  65536  6 Dec  2012 ramdisk.dat
# cat /media/RAMDISK/ramdisk.dat
USB Mass Storage Class example. USB Mass Storage Class example. ...

Check with https://github.com/libopencm3/libopencm3-examples/tree/master/examples/stm32/f4/stm32f4-discovery/usb_msc on STM32F407

PR #1003, opened 2019-01-13
Implement necessary commands for Windows support

howard0su:
Implement read_format_capacities and add inquiry minimum support. Tested on STM32F103 and Windows 10.


karlp:
See also https://github.com/libopencm3/libopencm3/pull/923 for alternate implmentation

PR #1004, opened 2019-01-13
EFM32 I2C: fix base undeclared

dlharmon:
Fix this error

```
libopencm3/include/libopencm3/efm32/common/i2c_common.h:234:29: error: 'base' undeclared (first use in this function)
 #define I2C0_CTRL  I2C_CTRL(base)
```

resulting from this code

```
I2C0_CTRL = 0;
```


karlp:
merged thanks.  (I fixed up the commit log to make it display better)

PR #1010, opened 2019-01-22
stm32f4: support inclusion into assembly files

alexkazik:
This makes all stm32f4 includes compatible to be included from assembly.

### DIFF
--- a/include/libopencm3/stm32/common/crypto_common_f24.h
+++ b/include/libopencm3/stm32/common/crypto_common_f24.h
@@ -240,6 +240,8 @@ specific memorymap.h header before including this header file.*/
  */
 /**@{*/
 
+#if !defined(__ASSEMBLER__)
+
 enum crypto_mode {
 	ENCRYPT_TDES_ECB = CRYP_CR_ALGOMODE_TDES_ECB,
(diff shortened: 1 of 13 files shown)


alexkazik:
I want to use the defines: the addresses and constants.
Because I don't want to look in the manual and copy the address over and over.

I needed only two more periphs but went for all of stm32f4.

I wanted to offer to go through all stm32 but forgot to write it - shouldn't be that hard since they share some libs.
I can try all but that will take some time.

And I only want to help in case somebody else wants to use this library with assembler.

alexkazik:
Not actually, the generated file does only contain the addresses and bit offsets but not the data, e.g. for GPIO mode: Input, Output, AF, Analog.

But if you don't want this is your repo I can live with that, just wanted to help.

alexkazik:
For the perips I used so far, the one common code did the trick...

If you'd like the adaption of all header files I'll start (may take some time).
But I can also just use my private copy of it and be done.

karlp:
If this can't find another sponsor or an approach with less impact on the code base, I'll close this ~march 2020

karlp:
closing, no new sponsor, no new updates.

PR #1015, opened 2019-01-29
exti: prepare for stm32g0 exti.

grevaillot:
Hi,

Here's a first patchset for g0 merge, regarding exti common code :  basically ST moved again the AFIO/SYSCFG_EXTICR bits, made the gpio bank selection on 8 bits instead of previously 4, and splitted the EXTI_PR into EXTI_FPR (Failing Edge) and EXTI_RPR (Rising Edge).
I attempted to patch current common code to support these features and aligned devices header to have them export their EXTICR gpio bank selection mux bit sized defined via SYSCFG_EXTICR_SKIP (naming might not be totally obvious, but it was defined as it on f0, so i kept it for consistency).

BR

Guillaume

### DIFF
--- a/include/libopencm3/stm32/common/syscfg_common_l1f234.h
+++ b/include/libopencm3/stm32/common/syscfg_common_l1f234.h
@@ -50,6 +50,10 @@ specific memorymap.h header before including this header file.*/
 
 #define SYSCFG_CMPCR			MMIO32(SYSCFG_BASE + 0x20)
 
+/* --- SYSCFG_EXTICR Values -------------------------------------------------*/
+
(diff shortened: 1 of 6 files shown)


karlp:
The f0 SYSCFG_EXTICR_SKIP definition wasn't used anywhere, I'd rather see it renamed to something more obvious like FIELDSIZE or something rather than propagating that strange name around the place.  Otherwise I think this looks pretty good.

karlp:
Heh, I like how even though it's 8 bits now, the highest defined value is 5.  Nice one ST.  

Merged after rebase.  I added a bit more to some of the commit messages.  Thanks for this!

grevaillot:
thanks :)

PR #1016, opened 2019-01-29
gd32: add new chip series f1x0

Icenowy:
GD32F1X0 (X can be 3, 5, 7 and 9) is a series of Cortex-M3 MCUs by
GigaDevice, which features pin-to-pin package compatibility with
STM32F030 MCU line. F150 adds USB support to F130, and F170/F190 adds
CAN support.

Currently the code mainly targets GD32F130 and F150 chips. Some register
are different between F130/150 and F170/190, just like the difference
between STM32F1 Performance line and Connectivity line.

From the perspective of registers and memory map, GD32F1X0 seems like a
mixture between STM32F1 and STM32F0 (because it is designed to be
pin-to-pin compatible with F0, but with Cortex-M3 like F1). A bunch of
code are shared between STM32 and GD32, and these code are specially
processed to include the GD32 headers instead of STM32 headers when meet
GD32F1X0.


include/libopencm3/gd32/f1x0/flash.h
```diff
@@ -0,0 +1,116 @@
...
+
+/** @defgroup flash_options Option Byte Addresses
+@ingroup flash_defines
+@{*/
+#define FLASH_OPTION_BYTE_0             FLASH_OPTION_BYTE(0)
+#define FLASH_OPTION_BYTE_1             FLASH_OPTION_BYTE(1)
+#define FLASH_OPTION_BYTE_2             FLASH_OPTION_BYTE(2)
+#define FLASH_OPTION_BYTE_3             FLASH_OPTION_BYTE(3)
+#define FLASH_OPTION_BYTE_4             FLASH_OPTION_BYTE(4)
+#define FLASH_OPTION_BYTE_5             FLASH_OPTION_BYTE(5)
```
karlp: I guess you copied this, but I'd question whether these realyl provide any value.

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -0,0 +1,507 @@
...
+ * @date 18 August 2012
+ *
+ * LGPL License Terms @ref lgpl_license
+ *  */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2019 Icenowy Zheng <icenowy@aosc.io>
+ * Copyright (C) 2009 Uwe Hermann <uwe@hermann-uwe.de>
+ * Copyright (C) 2009 Federico Ruiz-Ugalde <memeruiz at gmail dot com>
```
karlp: which file did you copy to get this name in it it? they only contributed to f3 code aiui

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -0,0 +1,507 @@
...
+void rcc_set_usbpre(uint32_t usbpre);
+void rcc_set_prediv(uint32_t prediv);
+uint32_t rcc_system_clock_source(void);
+void rcc_clock_setup_in_hsi_out_64mhz(void);
+void rcc_clock_setup_in_hsi_out_48mhz(void);
+void rcc_clock_setup_in_hsi_out_24mhz(void);
+void rcc_clock_setup_in_hse_8mhz_out_24mhz(void);
+void rcc_clock_setup_in_hse_8mhz_out_72mhz(void);
+void rcc_clock_setup_in_hse_12mhz_out_72mhz(void);
+void rcc_clock_setup_in_hse_16mhz_out_72mhz(void);
```
karlp: don't copy these. I don't care if f1 uses them, no-one else does, and we're not going to expand their usage.

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -0,0 +1,507 @@
...
+#define RCC_AHBENR				MMIO32(RCC_BASE + 0x14)
+#define RCC_APB2ENR				MMIO32(RCC_BASE + 0x18)
+#define RCC_APB1ENR				MMIO32(RCC_BASE + 0x1c)
+#define RCC_BDCR				MMIO32(RCC_BASE + 0x20)
+#define RCC_CSR					MMIO32(RCC_BASE + 0x24)
+#define RCC_AHBRSTR				MMIO32(RCC_BASE + 0x28)
+#define RCC_CFGR2				MMIO32(RCC_BASE + 0x2c)
+#define RCC_CFGR3				MMIO32(RCC_BASE + 0x30)
+#define RCC_CR2					MMIO32(RCC_BASE + 0x34)
+
```
karlp: How much is actually different in this file?

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -0,0 +1,507 @@
...
+	RST_GPIOC	= _REG_BIT(0x28, 19),
+	RST_GPIOD	= _REG_BIT(0x28, 20),
+	RST_GPIOE	= _REG_BIT(0x28, 21),
+	RST_GPIOF	= _REG_BIT(0x28, 22),
+	RST_TSC		= _REG_BIT(0x28, 24),
+
+	/* APB2 peripherals */
+	RST_SYSCFG	= _REG_BIT(0x0C, 0),
+	RST_ADC		= _REG_BIT(0x0C, 9),
+	RST_ADC1	= _REG_BIT(0x0C, 9), /* Compatibility alias */
```
karlp: I'd actually drop the xxx_ADC defines, there's no reason to be backward compatible here.

lib/gd32/f1x0/gd32f1x0x4.ld
```diff
@@ -0,0 +1,31 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* Linker script for GD32F1x0x4, 16K flash, 4K RAM. */
```
karlp: we've stopped including any explicit linker scripts for parts now, and have a tticket filed to remove the partial set we have already, we should just skip this one too.

lib/stm32/common/flash_common_all.c
```diff
@@ -21,7 +21,11 @@
 
 /**@{*/
 
+#ifndef GD32F1X0
 #include <libopencm3/stm32/flash.h>
+#else
```
karlp: hrm, I'd kinda been thinking of just having the stm32/flash.h dispatch header include gd32 files, rather than having to chop into each of the c files like this too.  I guess this works as well. not really sure what's prettier.

lib/gd32/f1x0/rcc.c
```diff
@@ -0,0 +1,1034 @@
...
+
+void rcc_backupdomain_reset(void)
+{
+	/* Set the backup domain software reset. */
+	RCC_BDCR |= RCC_BDCR_BDRST;
+
+	/* Clear the backup domain software reset. */
+	RCC_BDCR &= ~RCC_BDCR_BDRST;
+}
+
```
karlp: what's actually different in this file? It looks like ~whole copy?

include/libopencm3/gd32/f1x0/rcc.h
Icenowy: The F1 rcc.h

include/libopencm3/gd32/f1x0/flash.h
Icenowy: With the name in GD32 user manual, they're:
OB_SPC(Security Protection Code)
OB_USER(User configuration)
OB_DATA[7:0]
OB_DATA[15:8]
OB_WP[7:0]
OB_WP[15:8]

include/libopencm3/gd32/f1x0/rcc.h
Icenowy: It seems f0 also uses rcc_clock_setup functions.

By the way I still think provide pre-defined clock initialization code with specific osc in and sysclk out is convenient.

lib/stm32/common/flash_common_all.c
Icenowy: Agree.

Okay, thanks.

lib/gd32/f1x0/rcc.c
Icenowy: It's based on F1 rcc.c, with PLL2/3 references dropped, and rcc_set_prediv() copied from F0 (because the way it configures prediv is similar to F0, not F1).

include/libopencm3/gd32/f1x0/rcc.h
Icenowy: By the way, I totally agree to drop 12Mhz and 16MHz input functions, because I can not test them now. All my boards with GD32F1x0 that is hackable are 8MHz in.

include/libopencm3/gd32/f1x0/flash.h
Icenowy: Oh... It seems reasonable to drop these macros.

I'll do it.

Thanks.

lib/gd32/f1x0/rcc.c
```diff
@@ -0,0 +1,658 @@
...
+ * The arguments provide the pll source, multipliers, dividers, all that's
+ * needed to establish a system clock.
+ * @param clock clock information structure
+ */
+void rcc_clock_setup_pll(const struct rcc_clock_scale *clock)
+{
+	/* Enable internal high-speed oscillator. */
+	rcc_osc_on(RCC_HSI);
+	rcc_wait_for_osc_ready(RCC_HSI);
+	rcc_set_sysclk_source(RCC_CFGR_SW_SYSCLKSEL_HSICLK);
```
karlp: if hsi is unused, do you turn it off again?  Is this little section even necessary? It makes it go a bit faster I guess.

lib/gd32/f1x0/rcc.c
```diff
@@ -0,0 +1,658 @@
...
+
+The prescale factor can be set to 1 (no prescale) for use when the PLL clock is
+48MHz, or 1.5 to generate the 48MHz USB clock from a 64MHz PLL clock.
+
+@note This bit cannot be reset while the USB clock is enabled.
+
+@param[in] usbpre Unsigned int32. USB prescale factor @ref rcc_cfgr_usbpre
+*/
+
+void rcc_set_usbpre(uint32_t usbpre)
```
karlp: the docs here imply there's some sort of value you can pass in, but you've written this as just a bool?  perhaps change the signature to make that a bit more obvious?  (what do other families do with this? are they all crap?)

lib/gd32/f1x0/gpio.c
```diff
@@ -0,0 +1,31 @@
...
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
karlp: You don't need these dummy files.  It was for documentation generation, but that's been changed.

lib/gd32/f1x0/flash.c
```diff
@@ -0,0 +1,153 @@
...
+ *
+ * @note: don't forget to lock it again when all operations are complete.
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2013 Frantisek Burian <BuFran@seznam.cz>
```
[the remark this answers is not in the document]
karlp: so... why does this file need to exist then?  can't it just use the f1 code as is?

lib/gd32/f1x0/rcc.c
karlp: Can we just break the f1 rcc.c into a common portion and an stm32f1 portion and share more of it then? there's a _lot_ of copied code here.

lib/gd32/f1x0/flash.c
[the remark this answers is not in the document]
Icenowy: In fact I just copied f0 code, as f1 have multi-bank.

Could I replace this file with just an #include to f0 code?

lib/gd32/f1x0/gpio.c
Icenowy: Okay. I will remove them.

lib/gd32/f1x0/rcc.c
Icenowy: These code is from STM32F1.

It's in fact wrong here, because the STM32F1 usbpre is 1-bit, but GD32F1x0 (as well as GD32F1) is 2-bit.

I will fix the code here, but for F1 I think further discussion should be done.

lib/gd32/f1x0/rcc.c
Icenowy: Verified that this part can be omitted.

I just modelled after the original F1 code, in which HSI is enabled first.

lib/gd32/f1x0/rcc.c
Icenowy: Okay. However how should the file be named and placed?

lgeek:
@Icenowy Maybe you could cherry pick my devices.data commit? https://github.com/lgeek/libopencm3/commit/472f8ec944d5cb24ab14953c7ceb96de1ad92bda

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -0,0 +1,526 @@
...
+	uint8_t usbpre; /* Only valid if HSE used */
+	bool use_hse; /* PLL source is HSE if set, HSI/2 if unset */
+	uint8_t pll_hse_prediv; /* Only valid if HSE used */
+	uint32_t ahb_frequency;
+	uint32_t apb1_frequency;
+	uint32_t apb2_frequency;
+};
+
+extern const struct rcc_clock_scale rcc_hsi_configs[RCC_CLOCK_HSI_END];
+extern const struct rcc_clock_scale rcc_hse8_configs[RCC_CLOCK_HSE8_END];
```
karlp: Why do these need to be two arrays when the "use_hse" bit is part of the struct?

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -0,0 +1,526 @@
...
+enum rcc_osc {
+	RCC_PLL, RCC_HSE, RCC_HSI, RCC_LSE, RCC_LSI
+};
+
+#define _REG_BIT(base, bit)		(((base) << 5) + (bit))
+
+/* V = value line F100
+ * N = standard line F101, F102, F103
+ * C = communication line F105, F107
+ */
```
karlp: This comment above is definitely irreelvant to the gd32 lines.  I'm dubious about how much of this is just copied from the f1 file.  can't we just include the f1 file and add sections?  Are there problematyic missing portions that f1 has that gd32 doens't?  It just seems an awful lot of copying for htings that are the "same"?

lib/gd32/f1x0/Makefile
```diff
@@ -0,0 +1,47 @@
...
+ARFLAGS		= rcs
+
+OBJS		= rcc.o flash.o
+
+OBJS            += \
+                   gpio_common_all.o gpio_common_f0234.o \
+		   rcc_common_all.o \
+		   flash_common_all.o flash_common_f.o flash_common_f01.o
+
+VPATH += ../:../../cm3:../common:../../stm32/common
```
karlp: this is where we could also point to stm32/f1 for instance, and just be careful with the file namings for files purely within the gd32 directory.  I'm not opposed to having to break up some of the f1 files (or f0 files whichever) into functional chunks to make it easier to compose the code more easily.

lib/gd32/f1x0/flash.c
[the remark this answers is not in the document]
karlp: As I mentioned in the makefile comment avboe, I'd kinda rather see just referring to the stm32 flash file directly, rather than getting into  #including .c files.

lib/gd32/f1x0/rcc.c
karlp: were you doing more here?

karlp:
I'm merging this as is, see if it gets users.  we can deal with the duplication and refactorings down the road. better in than out I guess.

Please file a followup ticket enabling the documentation for this target!

karlp:
merged via rebase in 330d5fd5

PR #1017, opened 2019-01-30
stm32l[01]: flash common: add flash_unlock_acr

grevaillot:
flash_unlock_acr allows to unlock RUN_PD bit from FLASH_ACR register. Relock is done automatically
when writing 0 to RUN_PD, so no flash_lock_acr method.

Code is common to l0 and l1 devices.


karlp:
merged in 7afd86db thanks!

PR #1018, opened 2019-01-30
stm32l0: rcc: add some utils.

grevaillot:
hi,

Here's a couple of patch regarding stm32 l0 rcc, adding some msi setup and peripheral clock source configuration.

The msi bit is similar to l1 afaik, but no shared rcc code yet, and can be found on l4, but with different config register, so as for most of rcc related code, no easy way to avoid "duplicating" code yet.

Regarding the rcc_set_periph_sel bits, i found nothing in the lib yet, despite that kind of logic being present on a lot of chips, so i'm open to suggestion to put some nicer higher level / generic API later, ie: rcc_set_peripheral_selection(uint32 peripheral, uint32 clock_selection), that could be used on other chips, instead of defining millions of functions.

### DIFF
--- a/include/libopencm3/stm32/l0/rcc.h
+++ b/include/libopencm3/stm32/l0/rcc.h
@@ -695,6 +695,11 @@ void rcc_clock_setup_pll(const struct rcc_clock_scale *clock);
 
 void rcc_set_msi_range(uint32_t msi_range);
 
+void rcc_set_lptim1_sel(uint32_t lptim1_sel);
+void rcc_set_lpuart1_sel(uint32_t lpupart1_sel);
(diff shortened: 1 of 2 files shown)


grevaillot:
Ok, i added a simple rcc_set_peripheral_clk_sel(perif, selection) that could replace all these rcc_set_*_sel(selection) and could be used as common api for other chips.

karlp:
I like it. I've been seeing this sea of peripheral clock selection opitions coming, with l0, l4, g0 and friends, and was already dreading it.  This looks like a perfectly feasible way of doing it, sans anything better coming along.

karlp:
merged thhanks!

PR #1020, opened 2019-02-05
stm32g0 support

grevaillot:
Okay, so, here's a first patchset for g0 support.. This as been tested on G071 nucleo board.

I'm not sure this is convenient to review/merge big drop like these, so this first set only contains core bits.. I can split it in smaller bits if needed.

Basically G0 is a cortex m0+ chip, between F0 and L0/L4, so not too different, but with a couple of variations, mostly around core peripherals...

Briefly :
EXTI has some slight modifications, but support has already been merged.
RCC, as usual is specific implementation wise, but is similar..
PWR does not really match usual api regarding PVD bits, as this chip has two thresholds.
FLASH does not really match existing common code feature split - and current code only implements minimal stuff (waitstate..)
GPIO is standard.


grevaillot:
ok, maybe i'll update the pr with a smaller patchset then :) ?

karlp:
actually looked reasonable, just been a super busy week :)

include/libopencm3/stm32/g0/gpio.h
```diff
@@ -0,0 +1,76 @@
...
+/* Register values                                                           */
+/*****************************************************************************/
+
+/** @defgroup gpio_speed GPIO Output Pin Speed
+@ingroup gpio_defines
+@{*/
+#define GPIO_OSPEED_LOW		0x0
+#define GPIO_OSPEED_MED		0x1
+#define GPIO_OSPEED_FAST	0x2
+#define GPIO_OSPEED_HIGH	0x3
```
karlp: HIHG and HIGHEST or something is probably better than having to try and guess what's higher, "high" or "fast" but we can fix that later.

include/libopencm3/stm32/g0/memorymap.h
```diff
@@ -0,0 +1,91 @@
...
+#define GPIO_PORT_B_BASE		(IOPORT_BASE + 0x00400)
+#define GPIO_PORT_C_BASE		(IOPORT_BASE + 0x00800)
+#define GPIO_PORT_D_BASE		(IOPORT_BASE + 0x00c00)
+#define GPIO_PORT_E_BASE		(IOPORT_BASE + 0x01000)
+#define GPIO_PORT_F_BASE		(IOPORT_BASE + 0x01400)
+
+/* ST provided factory calibration values @ 3.0V */
+#define ST_VREFINT_CAL			MMIO16((INFO_BASE + 0xAA))
+#define ST_TSENSE_CAL1_30C		MMIO16((INFO_BASE + 0xA8))
+#define ST_TSENSE_CAL2_130C		MMIO16((INFO_BASE + 0xCA))
```
karlp: really? 130C? gee, thanks ST.  (confirmed in the datasheet)

include/libopencm3/stm32/g0/rcc.h
```diff
@@ -0,0 +1,948 @@
...
+#include <libopencm3/stm32/pwr.h>
+
+/* --- RCC registers ------------------------------------------------------- */
+
+#define RCC_CR_OFFSET			0x00
+#define RCC_CR					MMIO32(RCC_BASE + RCC_CR_OFFSET)
+#define RCC_ICSCR_OFFSET		0x04
+#define RCC_ICSCR				MMIO32(RCC_BASE + RCC_ICSCR_OFFSET)
+#define RCC_CFGR_OFFSET			0x08
+#define RCC_CFGR				MMIO32(RCC_BASE + RCC_CFGR_OFFSET)
```
karlp: why this style?  it's not done anywhere else, and I can't say I like it.

include/libopencm3/stm32/g0/rcc.h
```diff
@@ -0,0 +1,948 @@
...
+@sa rcc_pllcfgr_pllr
+@{*/
+#define RCC_PLLCFGR_PLLR_DIV(x)		((x)-1)
+#define RCC_PLLCFGR_PLLR_DIV2		1
+#define RCC_PLLCFGR_PLLR_DIV3		2
+#define RCC_PLLCFGR_PLLR_DIV4		3
+#define RCC_PLLCFGR_PLLR_DIV5		4
+#define RCC_PLLCFGR_PLLR_DIV6		5
+#define RCC_PLLCFGR_PLLR_DIV7		6
+#define RCC_PLLCFGR_PLLR_DIV8		7
```
karlp: do we really need both versions?  there's no backwards compat here.

include/libopencm3/stm32/g0/rcc.h
```diff
@@ -0,0 +1,948 @@
...
+@sa rcc_pllcfgr_pllq
+@{*/
+#define RCC_PLLCFGR_PLLQ_DIV(x)		((x)-1)
+#define RCC_PLLCFGR_PLLQ_DIV2		1
+#define RCC_PLLCFGR_PLLQ_DIV3		2
+#define RCC_PLLCFGR_PLLQ_DIV4		3
+#define RCC_PLLCFGR_PLLQ_DIV5		4
+#define RCC_PLLCFGR_PLLQ_DIV6		5
+#define RCC_PLLCFGR_PLLQ_DIV7		6
+#define RCC_PLLCFGR_PLLQ_DIV8		7
```
karlp: same here

include/libopencm3/stm32/g0/rcc.h
```diff
@@ -0,0 +1,948 @@
...
+#define RCC_PLLCFGR_PLLP_DIV23		22
+#define RCC_PLLCFGR_PLLP_DIV24		23
+#define RCC_PLLCFGR_PLLP_DIV25		24
+#define RCC_PLLCFGR_PLLP_DIV26		25
+#define RCC_PLLCFGR_PLLP_DIV27		26
+#define RCC_PLLCFGR_PLLP_DIV28		27
+#define RCC_PLLCFGR_PLLP_DIV29		28
+#define RCC_PLLCFGR_PLLP_DIV30		29
+#define RCC_PLLCFGR_PLLP_DIV31		30
+#define RCC_PLLCFGR_PLLP_DIV32		31
```
karlp: I _realllly_ don't see why we need all these.

include/libopencm3/stm32/g0/rcc.h
```diff
@@ -0,0 +1,948 @@
...
+@{*/
+#define RCC_PLLCFGR_PLLM_DIV(x)		((x)-1)
+#define RCC_PLLCFGR_PLLM_NODIV		0
+#define RCC_PLLCFGR_PLLM_DIV2		1
+#define RCC_PLLCFGR_PLLM_DIV3		2
+#define RCC_PLLCFGR_PLLM_DIV4		3
+#define RCC_PLLCFGR_PLLM_DIV5		4
+#define RCC_PLLCFGR_PLLM_DIV6		5
+#define RCC_PLLCFGR_PLLM_DIV7		6
+#define RCC_PLLCFGR_PLLM_DIV8		7
```
karlp: same

include/libopencm3/stm32/g0/rcc.h
```diff
@@ -0,0 +1,948 @@
...
+	RST_USART1 = _REG_BIT(RCC_APBRSTR2_OFFSET, 14),
+	RST_SPI1 = _REG_BIT(RCC_APBRSTR2_OFFSET, 12),
+	RST_TIM1 = _REG_BIT(RCC_APBRSTR2_OFFSET, 11),
+	RST_SYSCFG = _REG_BIT(RCC_APBRSTR2_OFFSET, 0),
+};
+
+struct rcc_clock_scale {
+	enum rcc_osc sysclock_source;
+		
+	// PLL as sysclock source cfg
```
karlp: wrong comment style

include/libopencm3/stm32/g0/rcc.h
```diff
@@ -0,0 +1,948 @@
...
+void rcc_enable_pllp(bool enable);
+void rcc_enable_pllq(bool enable);
+void rcc_enable_pllr(bool enable);
+
+void rcc_set_ppre(uint32_t ppre);
+void rcc_set_hpre(uint32_t hpre);
+void rcc_set_hsisys_div(uint32_t hsidiv);
+void rcc_set_mcopre(uint32_t mcopre);
+
+// add lsco stuff
```
karlp: drop the comment

lib/stm32/g0/rcc.c
```diff
@@ -0,0 +1,557 @@
...
+#include <libopencm3/stm32/pwr.h>
+#include <libopencm3/stm32/flash.h>
+#include <libopencm3/cm3/assert.h>
+
+/* Set the default clock frequencies after reset. */
+uint32_t rcc_ahb_frequency =  16000000;
+uint32_t rcc_apb1_frequency = 16000000;
+
+const struct rcc_clock_scale rcc_clock_config[RCC_CLOCK_CONFIG_END] = {
+	[RCC_CLOCK_CONFIG_LSI_32KHZ] = {
```
karlp: how portable is thhis style? isn't it for creating sparse arrays?  I mean, it hhelps you find them, but... still.

lib/stm32/g0/rcc.c
```diff
@@ -0,0 +1,557 @@
...
+{
+	uint8_t shift;
+	uint32_t mask;
+
+	switch (periph) {
+		case ADC1_BASE:
+			shift = RCC_CCIPR_ADCSEL_SHIFT;
+			mask = RCC_CCIPR_ADCSEL_MASK;
+			break;
+
```
karlp: kill all the blank lines here.

karlp:
you've got a lot of // style comments still too.

lib/stm32/g0/rcc.c
grevaillot: well, hm, I rechecked and this seem to be valid c99 (designated initializers) - Given the number of entries in that array, using them seemed safer.

include/libopencm3/stm32/g0/rcc.h
grevaillot: ack

include/libopencm3/stm32/g0/rcc.h
grevaillot: indeed... and i've got them like that like all around ... sorry.

include/libopencm3/stm32/g0/rcc.h
grevaillot: Indeed. I prefer the macro but current code base has both. I'll remove them.

include/libopencm3/stm32/g0/rcc.h
grevaillot: Ah! I attempted to do some clever stuff to to reduce amount of somewhat more or less duplicated code in rcc between devices, but nothing came out nicely and forgot about them.. I'll get rid of these offset defines.

grevaillot: hmm. or not. it's been a while now - it's also done like in include/libopencm3/stm32/l4/rcc.h, but i used it for all the definitions. I'll rest

grevaillot:
ok, this should merge with the gd stuff - luckily there was no documentation to merge :)

include/libopencm3/stm32/g0/gpio.h
grevaillot: Indeed. I based on the other devices naming (L4, not to name it), otherwise the refman register definition is very low/low/high/very high :

```
00: Very low speed
01: Low speed
10: High speed
11: Very high speed
```
L4 reference manual uses the same naming, so (apparently), it seemed "ok" to keep compatibility here.

include/libopencm3/stm32/g0/rcc.h
karlp: right, it's only for the ones that are used in the bit macros, so just for periphheral enable/reset...

the ```	RST_GPIOE = _REG_BIT(RCC_AHB2RSTR_OFFSET, 4),``` sort of stuff.

include/libopencm3/stm32/g0/gpio.h
karlp: ok, so... yay, compat with l4, boo, not the ref man, and not good choices :)  Leave it for now, so it merges "standard" and we can maybe redo both of them at a later date then.

karlp:
I've pulled this locally and rebased it, it's next in line.

karlp:
This has now landed.  I re-ordered a few things (pwr and flash actually needed to be before rcc to be buildable) but it's ~same as this PR.  Thanks for this!

grevaillot:
great, thanks; i'll push a second patchset soon, but i need to have a
second look at adc part, a saw some commits here :)

PR #1021, opened 2019-02-08
Use gcc-ar for archiving, to support Link Time Optimisation

deece:
When building with CFLAGS=-flto, we get the following errors:
      AR      libopencm3_sam3a.a
      /home/deece/src/onewire-softdevice/libopencm3/../toolchain/gcc-arm-none-eabi-8-2018-q4-major/bin/arm-none-eabi-ar: gpio_common_all.o: plugin needed to handle lto object

The easiest way to address this is to use gcc-ar rather than
binutils ar, as this supplies the necessary paths to the
archiver to handle LTO objects.

This change is transparent to non-LTO builds (gcc-ar passes
through to binutils ar with the same arguments).

Signed-off-by: Alastair D'Silva <alastair@d-silva.org>


deece:
Dupe of #850

PR #1022, opened 2019-02-09
stm32f1: adc: fixed deprecated/broken example code

szczys:
Doc fix: Code wouldn't compile. I've updated two deprecated functions and fixed a define that didn't exist.


karlp:
I feel like there was another ticket for exactly this, but I can't find it.

karlp:
merged in 0fd4f74e thanks

PR #1024, opened 2019-02-19
swm050: new MCU family

Icenowy:
SWM050 is a series of MCU made by Foshan Synwit Tech. It contains a
Cortex-M0 CPU core, 8KiB of Flash and 1KiB of SRAM. The only peripherals
are GPIO, Timer and WDT. There's only two parts in this series, with
either TSSOP-8 or SSOP-16 packages.

This commit introduces the interrupt vector and GPIO support for them.


include/libopencm3/swm050/gpio.h
```diff
@@ -0,0 +1,85 @@
...
+
+#define GPIO_INTEN			MMIO32(GPIO_BASE + 0x30)
+#define GPIO_INTMASK			MMIO32(GPIO_BASE + 0x34)
+#define GPIO_INTLEVEL			MMIO32(GPIO_BASE + 0x38)
+#define GPIO_INTPOLARITY		MMIO32(GPIO_BASE + 0x3c)
+#define GPIO_INTSTATUS			MMIO32(GPIO_BASE + 0x40)
+#define GPIO_INTRAWSTATUS		MMIO32(GPIO_BASE + 0x44)
+#define GPIO_INTEOI			MMIO32(GPIO_BASE + 0x48)
+
+#define SWD_SEL				MMIO32(SYSTEM_CON_BASE + 0x80)
```
karlp: Should this define exist? it's got a prefix that doesn't match this file, and it's a duplicate of the next three defines.  (likewise, why do you need all four of these?)

include/libopencm3/swm050/irq.json
```diff
@@ -0,0 +1,21 @@
...
+	"gpioa1",
+	"gpioa2",
+	"gpioa3",
+	"gpioa4",
+	"gpioa5",
+	"gpioa6",
+	"gpioa7",
+	"gpioa8",
+	"gpioa9"
+    ],
```
karlp: wow, really? those are the only irqs?

lib/swm050/gpio.c
```diff
@@ -0,0 +1,173 @@
...
+Set the direction of one or more pins of GPIO.
+
+@param[in] gpios Unsigned int16. Pin identifiers @ref gpio_pin_id
+	     If multiple pins are to be changed, use bitwise OR '|' to separate
+	     them.
+@param[in] dir Unsigned int8. Pin direction @ref gpio_dir
+*/
+void gpio_dir(uint16_t gpios, uint8_t dir)
+{
+	if (dir == GPIO_INPUT)
```
karlp: {} here please.  there's a make checkstyle target that should help a bit

lib/swm050/gpio.c
```diff
@@ -0,0 +1,173 @@
...
+/** @brief Set the direction of a Group of Pins
+
+Set the direction of one or more pins of GPIO.
+
+@param[in] gpios Unsigned int16. Pin identifiers @ref gpio_pin_id
+	     If multiple pins are to be changed, use bitwise OR '|' to separate
+	     them.
+@param[in] dir Unsigned int8. Pin direction @ref gpio_dir
+*/
+void gpio_dir(uint16_t gpios, uint8_t dir)
```
karlp: are there more than two states here? would a bool be enough for this? rather than uint8?

lib/swm050/gpio.c
```diff
@@ -0,0 +1,173 @@
...
+void gpio_sel_af(uint16_t gpios, bool af_en)
+{
+	if (gpios & GPIO0)
+		GPIO_SEL = (GPIO_SEL & (~0x3)) | (af_en ? 0x1 : 0x0);
+	if (gpios & GPIO1)
+		GPIO_SEL = (GPIO_SEL & (~0xc)) | (af_en ? 0x4 : 0x0);
+	if (gpios & GPIO2)
+		GPIO_SEL = (GPIO_SEL & (~0x30)) | (af_en ? 0x10 : 0x0);
+	if (gpios & GPIO7)
+		GPIO_SEL = (GPIO_SEL & (~0xc000)) | (af_en ? 0x4000 : 0x0);
```
karlp: do only thouse four pins support AF? (I've not looked at the manuals for this part, but that code looks odd)

include/libopencm3/swm050/gpio.h
```diff
@@ -0,0 +1,85 @@
...
+#define GPIO_SEL			MMIO32(SYSTEM_CON_BASE + 0x80)
+#define GPIO_PULLUP			MMIO32(SYSTEM_CON_BASE + 0x80)
+#define GPIO_INEN			MMIO32(SYSTEM_CON_BASE + 0x80)
```
karlp: yeah, ar eyou sure about these? the functions you have for these use them, but I don't see how input enable is also pullup enable?

lib/swm050/gpio.c
```diff
@@ -0,0 +1,173 @@
...
+
+/*---------------------------------------------------------------------------*/
+/** @brief Enable the SWD debugging port
+
+Enable or disable the SWD debugging port. When SWD debugging port is enabled,
+GPIO and AF of the SWD pins will be both unavailable.
+
+@param[in] en Bool. Whether SWD is enabled
+*/
+void swd_en(bool en)
```
karlp: I feel like this should probably be in a different file.  We try and keep the namespaces somewhat clear for each file too.  alternatively, prefix this with gpio_

karlp:
This generally looks good, apart form those addresses that seem to be copy paste errors, thanks!

include/libopencm3/swm050/irq.json
Icenowy: Yes. This is a quite small chip, and what I get is in TSSOP-8 package.

Icenowy: Of course, for utilizing all GPIOs, a SSOP-16 package is also available.

include/libopencm3/swm050/gpio.h
Icenowy: oh sorry, copy-pasta error, should be 0x30, 0x80, 0x90, 0xe0.

For the SWD_SEL logic, the two pins have 3 status possible: SWD, AF and GPIO. When SWD_SEL is 1, SWD is available; when SWD_SEL is 0 and GPIO_SEL (in the datasheet named PORTA_SEL, but this chip has only Port A) have the AF of the corresponding pin set, AF is available, otherwise GPIO is available.

lib/swm050/gpio.c
Icenowy: No. Bool is enough, just difficult to decide whether true is input or false is input.

lib/swm050/gpio.c
Icenowy: Oh sorry, I forgot that libopencm3 coding style is a little different than Linux one.

lib/swm050/gpio.c
Icenowy: Yes, other pins can only do GPIO.

Although the datasheet is Chinese only, the pin setup picture has no Chinese inside, so I copied it here.

image

lib/swm050/gpio.c
Icenowy: This is useful for enabling the GPIO under SWD pins.

Maybe it should be named something like gpio_switch_from_swd ?

lib/swm050/gpio.c
karlp: gpio_output() or gpio_input() would determine the direction bool, doesn't really matter which

lib/swm050/gpio.c
karlp: ok, cool, just the sort of thing thatlooks odd :)

include/libopencm3/swm050/gpio.h
```diff
@@ -0,0 +1,86 @@
...
+void gpio_set(uint16_t gpios);
+void gpio_clear(uint16_t gpios);
+uint16_t gpio_get(uint16_t gpios);
+void gpio_toggle(uint16_t gpios);
+
+void gpio_input(uint16_t gpios);
+void gpio_output(uint16_t gpios);
+void gpio_sel_af(uint16_t gpios, bool af_en);
+void gpio_pullup(uint16_t gpios, bool en);
+void gpio_in_en(uint16_t gpios, bool en);
```
karlp: what's the difference between gpio_input() and gpio_in_en?   feel these names are a bit vague.

gpio_set_input?
gpio_set_otuput
gpio_set_af
gpio_enable_pullup
gpio_???

lib/swm050/gpio.c
```diff
@@ -0,0 +1,190 @@
...
+
+Enable or disable the input function of one or more pins of GPIO. Disabling
+the input function of pins can decrease the power usage of the MCU.
+
+@param[in] gpios Unsigned int16. Pin identifiers @ref gpio_pin_id
+	     If multiple pins are to be changed, use bitwise OR '|' to separate
+	     them.
+@param[in] en Bool. Whether input function is enabled
+*/
+void gpio_in_en(uint16_t gpios, bool en)
```
karlp: I'm afraid it's still not clear to me what the difference is between gpio_input and gpio_in_en.  I can _see_ thaht they're different registers, but I don't understand from the documentation when I need each one.  Particularly this one.

karlp:
It's just the gpio_in_en() function that's holding this up now. thanks for your patience

Szalacinski:
The datasheet says that PORTA_INEN (GPIO_INEN in the commit), which gpio_in_en() corresponds to, is a register to completely disable input for power savings.  Seeing as I'm probably reading a machine translated version, which may be inaccurate, it may be that it disables the pin entirely, rather than setting the input or output functions, with GPIO_XDIR (GPIO_DIR in the commit).  I think they may have used 输入 (input) to refer to the pin itself.  If this is not the case, then my only guess as to why they did not automatically trigger a power-saving function when the pin is set to output would imply that it may be slower to disable the input rather than switch between input or output, poor planning on the manufacturer's part, or that we are missing some sort of programmer's manual that would further clarify this register.  If there are any tests that you would like me to run, I just received an SWM050 dev board in the mail today from smdprutser.nl

karlp:
@Szalacinski and that's exactly the sort of information that needs to be in the doxygen so that people can know what function they need to be using for what.

karlp:
but I can maybe reword that a little when I merge it.  I'll take it for now.

include/libopencm3/swm050/gpio.h
```diff
@@ -0,0 +1,86 @@
...
+/** @defgroup gpio_dir GPIO Pin Direction
+@ingroup gpio_defines
+@{*/
+#define GPIO_INPUT			0x0
+#define GPIO_OUTPUT			0x1
+/**@}*/
+
+#define GPIO_DATA			MMIO32(GPIO_BASE + 0x0)
+#define GPIO_DIR			MMIO32(GPIO_BASE + 0x4)
+#define GPIO_EXT			MMIO32(GPIO_BASE + 0x4c)
```
karlp: one last thing, which pdf did you get these values from?  I'm looking at and it has "PORTA_EXT" at 0x50, with "INTEOI" at 0x4c, and 0x48 as "reversed" (presumably reserved)

Szalacinski: [http://caxapa.ru/thumbs/870237/Foshan-Synwit-Tech-SWM050I2P7-65_C111818.pdf](url)

Szalacinski: I'm seeing the same thing

karlp: Well, the summary table has the other address, the descriptions has the address as you have them... joy  I've kept it as  you've got here, but would love to see it confirmed, whether it matches the table or the descriptions in reality :)

karlp:
I've just merged this, I feel it should get some leeway as a new target.  However, in general, 

- register names should match the datasheet names
- files should handle a single peripheral.  The SYSCON stuff should really be in a separate namespace, not lumped in as gpio.

But all this can wait til it has users. it's good to get it in for now.  thanks for your patience while I worked through this!

PR #1025, opened 2019-02-19
doxygen config generation

grevaillot:
Hi,

This PR suggests a way to get rid of plenty of copypasted doxygen stuff, by generating doc root doxygenlayout file, as well as devices stuff, based on main Makefile $(TARGETS) and template files.

This also allows to build only one device doc easily (ie: make TARGETS=stm32f0 doc), making documentation fixing less painfull, and make doc folder device agnostic.

We currently loose device "fancy" naming as device name is guessed (toUpper()..) from folder name - so currently, devices like efm32hg are not known as Happy Gecko anymore, but I currently dont see any elegant way to do that while keeping the doc folder device agnostic. Maybe we could define that in device's makefile ? I'm open to suggestions..


scripts/gendoxylayout.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# This library is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library. If not, see <http://www.gnu.org/licenses/>.
+
+from lxml import etree
```
karlp: lxml is a non-core library right?  and it's one that has to be compiled?

karlp: if I'm reading right, the expat binding is the only one standard.  https://docs.python.org/2/library/xml.html (and same for v3)  If we use that, it's less deps to install.  I like the idea of this PR, but I don't really like adding deps for this

grevaillot: arf, It looked python standard here - I'll have a look at that.

grevaillot:
Ok, I just pushed a modified tool using xml.etree.ElementTree - apparently ok on python2/3 out of the box.
ElementTree stuff is not as convenient to use as lxml, so it's a bit less nice ..

karlp:
excellent thanks.  tested with both python2.7 and 3. something.  used it to turn ondocs for the gd32 nice and quickly too.

karlp:
(merrged in bdc38eed1 and the one before)

grevaillot:
nice, cool :)

PR #1027, opened 2019-02-21
CM3: fix SCS_DWT_LSR and SCS_DWT_LAR definitions

cuvoodoo:
SCS_DWT_BASE as base address used by SCS_DWT_LSR and SCS_DWT_LAR
is not defined, but DWT_BASE is.

### DIFF
--- a/include/libopencm3/cm3/scs.h
+++ b/include/libopencm3/cm3/scs.h
@@ -316,9 +316,9 @@
 #define SCS_DWT_PCSR		MMIO32(DWT_BASE + 0x18)
 
 /* CoreSight Lock Status Register for this peripheral */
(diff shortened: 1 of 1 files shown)


cuvoodoo:
actually the LSR and LAR registers are not part of SCS_DWT, but CoreSight (see ARM CoreSight Architecture Specification v3.0 B2.1.5 Programmers’ Model Quick Reference).
SCS_DWT_LSR and SCS_DWT_LAR should be removed, and LSR and LAR should be defined in a new file along all other CoreSight registers.

karlp:
There's a bit of legacy here from things when they all ended up in scs.h, but thye could definitely be moved into some of the other files, there's already a dwt.h for instance.

PR #1030, opened 2019-03-28
Add a couple of ld scripts for stm32f4

Peque:
There is currently an `stm32f405x6.ld` script, but there is no STM32F405X6 as far as I know. Should it be removed? Or am I missing something?


karlp:
Not adding any new ones, will eventually remove the existing ones.  Please use the DEVICE= linker script generator, or provide your own little header, and include the cortex-m-generic.ld yourself.

Peque:
@karlp Okay, I will have a look at that! :+1:

PR #1031, opened 2019-03-28
stm32f4: add HSI clock scale configurations

Peque:
I defined `rcc_hsi_3v3` as an exact copy of `rcc_hse_16mhz_3v3`. Please, tell me if there is a better way to avoid duplicated code.


karlp:
right way would be to modify the setup function to add a clock source parameter.

PR #1032, opened 2019-03-28
stm32f4: add HSI clock setup function

Peque:


karlp:
This is more like it, but really, just add it to the struct.  It's already done like that on some other familes

Peque:
@karlp Thanks for your (very quick) response. :blush:

Would you mind pointing out an example among those families? Otherwise I will look for it, but it may be quicker. :innocent:

karlp:
If you're not looking at other families when you're introducing APIs, you need to start :)  We want things to be consistent where possible, not just arbitrary differences across them.  f3 has a struct with a pll source field, and is a good example of the desired direction.

```
f1: separate functions per config
f0: separate functions per config
f2/f4: hse function only
f3: nice single function
f7: hse and hsi functions separate
l0: nice single function
l1: nice single function
l4: no function at all yet.
```

f3, l0, l1 are good examples, I'd rather see f2/f4 move to that style.  f1 and f0 are horrible apis and will eventually be burnt on a fire.

karlp:
See also the same comments on https://github.com/libopencm3/libopencm3/pull/857

karlp:
See also the same comments on https://github.com/libopencm3/libopencm3/pull/857 and https://github.com/libopencm3/libopencm3/pull/696

Peque:
@karlp Couple of questions:

- L1 uses `.pll_source` field; F3 uses `.pllsrc`. Any preferred name?
- L1 defines `RCC_CFGR_PLLSRC_HSI_CLK` and `RCC_CFGR_PLLSRC_HSE_CLK`; F3 defines `RCC_CFGR_PLLSRC_HSI_DIV2` and `RCC_CFGR_PLLSRC_HSE_PREDIV`. Which naming style should I prefer? (being a HSE *or* HSI macro I would rather not append `DIV3` or `PREDIV`, but maybe I am missing something)
- Current F4 clock setup function is named `rcc_clock_setup_hse_3v3`, I guess it is okay to replace the `_hse` with `_pll` and leave only a single `rcc_clock_setup_pll_3v3` function, or would you rather keep it for backwards compatibility?
- Should I remove the `_3v3` parts from the module too? Have not seen them in other families and there does not seem to be any other option in F4, so why should it be there? (when it could be assumed the default is 3v3)

karlp:
1. No preference on the name other than to not make a third one.
2. they're different because the pll sources available are different.  Compare the reference manuals for the two parts.  F4 only has HSI or HSE, so just use those too.  
3. leave the existing one for now, but drop the 3v3 as well from your new one., that's certainly out of scope too now.    It should automatically do the right thing based on the voltage_scale enums in the struct.

Peque:
@karlp Something like this?

Do not merge, have not tested it on the board, just to know if that is what you are looking for.

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -790,6 +795,7 @@ struct rcc_clock_scale {
 	uint32_t apb2_frequency;
 };
 
+extern const struct rcc_clock_scale rcc_hsi_configs[RCC_CLOCK_3V3_END];
```
karlp: I'm not really enthused by having a bunch of separate lists, especially when they all seem to use the same end address, but yeah, it's one way of doing it.

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -1095,6 +1101,7 @@ void rcc_set_main_pll_hsi(uint32_t pllm, uint32_t plln, uint32_t pllp,
 void rcc_set_main_pll_hse(uint32_t pllm, uint32_t plln, uint32_t pllp,
 			  uint32_t pllq, uint32_t pllr);
 uint32_t rcc_system_clock_source(void);
+void rcc_clock_setup_pll(const struct rcc_clock_scale *clock);
 void rcc_clock_setup_hse_3v3(const struct rcc_clock_scale *clock);
```
karlp: you can add an attribute deprecated here if you like, see f3 rcc.c, or you can add @deprecated doxygen to the comments to mark it in the docs.  (you can even do both!) :)

lib/stm32/f4/rcc.c
```diff
@@ -49,13 +49,102 @@ uint32_t rcc_ahb_frequency = 16000000;
...
+		.hpre = RCC_CFGR_HPRE_DIV_NONE,
+		.ppre1 = RCC_CFGR_PPRE_DIV_4,
+		.ppre2 = RCC_CFGR_PPRE_DIV_2,
+		.voltage_scale = PWR_SCALE1,
+		.flash_config = FLASH_ACR_DCEN | FLASH_ACR_ICEN |
+				FLASH_ACR_LATENCY_3WS,
+		.ahb_frequency  = 120000000,
+		.apb1_frequency = 30000000,
+		.apb2_frequency = 60000000,
+	},
```
karlp: The three configs above here are not worth having in the library, IMO.

lib/stm32/f4/rcc.c
```diff
@@ -49,13 +49,102 @@ uint32_t rcc_ahb_frequency = 16000000;
...
+				FLASH_ACR_LATENCY_5WS,
+		.ahb_frequency  = 180000000,
+		.apb1_frequency = 45000000,
+		.apb2_frequency = 90000000,
+	},
+};
+
 const struct rcc_clock_scale rcc_hse_8mhz_3v3[RCC_CLOCK_3V3_END] = {
 	{ /* 48MHz */
 		.pllm = 8,
```
karlp: I know they're here, so we're not getting rid of them right now, but they're pointless dangling garbage that no-one uses.

lib/stm32/f4/rcc.c
```diff
@@ -812,4 +935,9 @@ void rcc_clock_setup_hse_3v3(const struct rcc_clock_scale *clock)
 	rcc_osc_off(RCC_HSI);
 }
 
+void __attribute__((deprecated)) rcc_clock_setup_hse_3v3(const struct rcc_clock_scale *clock)
```
karlp: yep :)

karlp:
yeah, that looks pretty good.

include/libopencm3/stm32/f4/rcc.h
Peque: What would you propose? A function instead? Is there an example in other families?

include/libopencm3/stm32/f4/rcc.h
Peque: Added the `@deprecated` coment: 65a00fe49ec5f65f9a838d03d8a0ba1b12ff8cec

lib/stm32/f4/rcc.c
Peque: I can delete all of them in a new commit (i.e.: all the 120/84/48 MHz configurations, because they existed before for multiple HSE setups). Want me to?

lib/stm32/f4/rcc.c
Peque: Allow me to mark this as resolved. :innocent:

lib/stm32/f4/rcc.c
Peque: Allow me to mark this as resolved. :innocent:

Peque:
@karlp Friendly ping (waiting for feedback on the review comments). :innocent:

lib/stm32/f4/rcc.c
karlp: No, I see now why you had it.  I can clean that up later.  this was totally the right thing to do consistency wise.

Peque: Resolving this conversation then, if you don't mind.

include/libopencm3/stm32/f4/rcc.h
Peque: Resolving this conversation, if you don't mind.

include/libopencm3/stm32/f4/rcc.h
Peque: @karlp Should we mark this as resolved then? May I proceed with squashing the fix-up commits?

karlp: nah, it's all good, I've merged and squished them locally, just doign some other tests and it will probably land within the hour :)

karlp:
This has now landed.  I added a few extra bits afterwards to make the deprecated stuff work nicely.  Thanks for this!

Peque:
@karlp Thank you for your time and for your guidance/feedback throughout this pull request. :heart:

PR #1035, opened 2019-04-04
docs: HACKING: Fixed dead link to Linux kernel coding style

Masterkraft0r:
The link to the Linux kernel coding style you provided is dead, so I fixed that.


karlp:
Thanks for pointing this out.  that link will die rapidly too, however, so I've updated it to use https://www.kernel.org/doc/html/latest/process/coding-style.html, which, while not static, is static enough giving how conservative the kernel is.

PR #1037, opened 2019-04-11
Connect up additional libraries for STM32F7 series

rschlaikjer:
Update includes & objects for 
- IWDG
- CRC
- Syscfg (header only)
- CAN (header didn't have include guard, but wasn't included in link list)
- Ethernet

All tested on STM32F745 and 750 cores.


karlp:
Merged thanks.  (I reworded the titles to include the stm32f7 tag)

PR #1040, opened 2019-04-27
stm32: Fix adc_calibrate_async() 

mjoldfield:
In adc_calibrate_async() we should only set the ADCAL bit and leave the rest alone. 

The old code set the entire register.

### DIFF
--- a/lib/stm32/common/adc_common_v2.c
+++ b/lib/stm32/common/adc_common_v2.c
@@ -155,7 +155,7 @@ void adc_power_off(uint32_t adc)
  */
 void adc_calibrate_async(uint32_t adc)
 {
-	ADC_CR(adc) = ADC_CR_ADCAL;
+	ADC_CR(adc) |= ADC_CR_ADCAL;
 }
 
 /**


karlp:
.... kinda :)  The only bit that is not "rs" is the ADCVREGEN bit, but you can only write a zero to it when the adc is disabled, at which point you can't set the calibration flag.

So, while it appears odd, it's actually completely safe, and _marginally_ faster.

Were you having problems that you fixed with this code? or just "fixed it" while looking at something else?

mjoldfield:
Yes, I was having problems on the STM32L431RCT6.

The code below works if you only set the bit, but fails if you set the whole register. In particular the
adc_read_regular() call returns 0. The ellipses hide other bits of code which run for at least 50ms.

```
  rcc_periph_clock_enable(adc_clk);
  RCC_CCIPR   |= (3UL << 28);  // select system clock and not PLLADC
  rcc_periph_clock_enable(gpio_clk);

...

  ADC_CR(adc) &= ~ADC_CR_DEEPPWD;
  ADC_CR(adc) |=  ADC_CR_ADVREGEN;

...

  ADC_CR(adc) |= ADC_CR_ADCAL;

  while(adc_is_calibrating(adc));

  adc_power_on(adc);

  channels[0] = channel;
  adc_set_regular_sequence(adc, 1, channels);

  adc_start_conversion_regular(adc);

...

  while (!adc_eoc(adc));

  *result = adc_read_regular(adc);
```

karlp:
right yes, vreg gets turned off again.

karlp:
merged in fc8a6aa7 thanks

PR #1041, opened 2019-04-28
Added F7 ADC support

matthewlai:
According to AN4660 (F4 to F7 migration guide), the only ADC difference is in the mapping of EXTSEL values.

Added support for F7 by refactoring everything in F4 ADC except for EXTSEL definitions into adc_common_v3, and have both F4 and F7 depend on that and add the EXTSEL values.


karlp:
Hi,

This is in principle good, however, a) It's not v3, that would be misleading, it's the "extended" v1 peripheral, which splits the common/global registers, has configurable resolutions, and includes the overrun flags.  I'd pretty much just suggest calling the file adc_common_v1_multi, even though it's a different "multi" than in the v2 peripheral (sequence capable or not)  but definitely _v1_<something>

but the bigger one is b) this is _also_ shared very heavily with L1.  L1/f4/f7 are all ~same compared to f1,  which is the other "v1" peripheral.

matthewlai:
I've renamed it to adc_common_v1_multi.

I took a quick look at L1 and it looks like at least the register addresses have quite a few differences from v1_multi?

Eg. JOFRx registers have different addresses, SQRx have different addresses, same for JDRx. The EXTSEL parts are different, too. It doesn't seem like it would be helpful to support it with v1_multi?

karlp:
JOFR and SQR are different just because there's more of them, the definitions of them are all the same though.    I'm going to pull this locally and see what I can do.

karlp:
I see what you mean though .)  I think I'm going to keep the multi and pull in l1, and _also_ keep a common f4/7 file too....

karlp:
Have a look at the linked PR ifyou can, and I'll try and merge the whole lot if you are ok with it.

karlp:
merged via the other PR now thanks.

PR #1042, opened 2019-04-30
stm32l4: Correct memorymap and add the existing CAN library

iqyx:
Name of the CAN peripheral in the memorymap was corrected to be able to include the existing CAN library. Compiles ok, tested on STM32L432.


karlp:
merged in e50ce6a8 thanks

PR #1044, opened 2019-05-10
Reference PR for f7 adc...

karlp:
@matthewlai here's what I've put together taking your base commits unifying f4 and f7, and adding in L1, and perhaps making it easier for f37x and f2 in the future...

See https://github.com/libopencm3/libopencm3/pull/1041 for original discussion.

This contains your code, but extended a lot.  It's not had any testing on parts though, just shuffling code around.  It all compiles and documents properly, but hasn't been checked on any hardware, just a lot of checks against ref manuals.  (Even found an old bug in L1 code)


matthewlai:
Awesome thanks! Looks good to me, but also don't have a board ready to test at the moment.

vector91:
Hi!
If it's ok for you, I can do some tests on a nucleo 767zi in the coming days

karlp:
More testing always welcome :)  I'm trying to resurrect some old adc tests and make them more useful too.

vector91:
I tested a bit single, continuous and scan mode, with and without DMA and all looks ok.
I didn't test injected channels, discontinuous mode and the multi ADC modes.
I will try to test them next week :)

PR #1048, opened 2019-05-14
PROPOSAL: find common headers for doxygen

H2OBrain:
Problem:
- The scripts/gendoxylist script does not detect headers which are not in the chip-series include folder,  when no c-sources reference them
Solution:
- Create a dummy source-code which includes all the header files in the manufacturer AND series folder
- Extract all header references by running the precompiler on the generated source code
Problems with the solution:
- Compilation requires some flags which should probably come as arguments, rather than a guessed (eg. gcc -D__ARM_ARCH_7M__ -DSTM32F7)

Etc:
- Maybe remove the "All headers for core/platform.." part and change `grep -e "${COMMON_PATH}/[^/]*.h"`
    - to `grep -e "${SERIES_PATH}/[^/]*.h" -e "${COMMON_PATH}/[^/]*.h"`
    - or even to `grep -e "${SERIES_PATH}/[^/]*.h" -e "${MANUFACTURER_PATH}/[^/]*.h" -e "${COMMON_PATH}/[^/]*.h"` to include "manufacturer"-header files too

### DIFF
--- a/scripts/gendoxylist
+++ b/scripts/gendoxylist
@@ -17,6 +17,16 @@ for FN in $(find ${UNTRACKED_DIRS} -name '*.h'); do
 	printf "INPUT += ../%s\n" "$FN" >> ${ODIR}/${ONAME}
 done
 
+# Find header dependencies in the manufacturer-folder (eg. libopencm3/stm32)
+# and ../common folder (eg. libopencm3/stm32/common)
+printf "# Headers lying around..\n" >> ${ODIR}/${ONAME}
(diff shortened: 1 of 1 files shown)


karlp:
well, making a dummy source code file is how things have always worked, you could just make the .c file and commit it with less magic, and less issues with compiler flags and so on.  Really, if you are at the stage of having common shared files, you could/should be at the stage of having some common code too, and then all this is moot.

H2OBrain:
Okay, I will use dummy source files then

H2OBrain:
easy.. i was just having fun with bash :)

PR #1050, opened 2019-05-19
Updated rules

h2obrain:
The first commit adds double fpu for stm32f7 targets which support double fpu

The second one normalizes LIBNAME from libopencm3_lpc43xx to libopencm3_lpc43xxm4
- lpc43xx was the only target for which LIBNAME did not match the lib path
- with this fix partial compilation of libopencm3 is possible from Makefile


karlp:
What does "normalized" mean for the lpc43?

for the f7 one, you're changing the flags that user code will be compiled with, but it won't change what the library is compiled with.  The only way around this is to have two libs built, one for each lib flags.

H2OBrain:
lpc43: all other libs are named after the path to their Makefile. Like lib/stm32/f4/Makefile becomes libopencm3_stm32f4.a, but lib/lpc43xx/m4/Makefile becomes libopencm3_lpc43xx.a. By adding this fix, my extension to the [example/rules.mk](https://github.com/H2OBrain/libopencm3-examples/commit/58e8057dca62acf6fac389d64735e359ba4f0dfc#diff-bfebe34154a0dfd9fc7b447fc9ed74e9) would work without any fixes. Also the lpc43xx examples would find their lib.

stm32f7: it looks like the lib is compatible as long as it is not compiled with -msoft-float, but i could not find any official answer to this

PR #1052, opened 2019-05-21
.gitignore: ignore autogenerated include/libopencmsis/gd32/

h2obrain:


karlp:
merged via rebas in 58536780 thanks!

PR #1053, opened 2019-05-22
mk/genlink-config: provide LIBNAME in devices autogeneration

h2obrain:
This adds .a to LIBDEPS and always defines LIBNAME.
I do not know how to find out when LIBNAME is opencm3_$(genlink_family). None of the examples need this.


karlp:
Ths missing .a was a bug, now fixed, I'm not sure I follow what else you were trying to do here?

h2obrain:
I would like to get the LIBNAME-variable out of genlink-config no matter if libopencm3 was compiled. If LIBNAME was already defined when the genlink-config script is included, I do not want it to be overwritten.
In the current examples it will try to build $(OPENCM3_DIR)/lib/lib.a when you type make in the example-subfolder before building libopencm3, which is a bit confusing for people not able to read the README.md like myself :)

karlp:
if you are defining LIBNAME, you have already decided you don't want to use the genlink stuff.  There is a readme, there is a template.  I really don't want to go around changing things, "just because" you want it done differently, for.... unclear reasons.  Is there some sort of actual _use case_ you are unable to do?  All I can see is, "please change, because of.... something"

karlp:
(talk to me like I'm five please, I'm obviously not understanding what you're trying to do here yet)

h2obrain:
i would like to remove all the switches and replace lines 55 to 65 with `LIBNAME = opencm3_$(genlink_family)`, but for some reason, there is this special(?) case where `LIBNAME = opencm3_$(genlink_subfamily)`, so (as a cheap fix) I check for the `$(OPENCM3_DIR)/lib/libopencm3_$(genlink_subfamily).a` file first

i personally do not want this `ifeq ($(LIBNAME),)` switch. i only added it to account for this comment (which I forgot to adapt):
# only append to LDFLAGS if the library file exists to not break builds
# where those are provided by different means

PR #1054, opened 2019-05-24
Add some ITM/TPIU/SCS utility methods useful for enabling SWO debug

rschlaikjer:
These methods making setting up SWO output a little more readable. The example code has been tested to work on the STM32F302 core, and should work on other cortex-M3/4/7 devices that embed a full ITM/TPIU.


ChuckM:
Odd that the gadget code is failing as this doesn't seem to touch that.

rschlaikjer:
I seem to get the same error building master - looks like this merge may have caused the issue?
https://travis-ci.org/libopencm3/libopencm3/builds/535754972

rschlaikjer:
Looks like this fixes the build problem: https://github.com/libopencm3/libopencm3/pull/1055

include/libopencm3/cm3/itm.h
```diff
@@ -89,6 +98,59 @@
...
+#define ITM_PORT22 (1 << 22)
+#define ITM_PORT23 (1 << 23)
+#define ITM_PORT24 (1 << 24)
+#define ITM_PORT25 (1 << 25)
+#define ITM_PORT26 (1 << 26)
+#define ITM_PORT27 (1 << 27)
+#define ITM_PORT28 (1 << 28)
+#define ITM_PORT29 (1 << 29)
+#define ITM_PORT30 (1 << 30)
+#define ITM_PORT31 (1 << 31)
```
karlp: no, no please no :) We don't need this sort of thing.

lib/cm3/itm.c
```diff
@@ -0,0 +1,156 @@
...
+/**
+@section  itm_example Example ITM setup for async SWO
+
+Example: Async SWO output enabled, Manchester format, 8MHz:
+@code
+  // Enable trace subsystem
+  scs_set_trace_enabled(true);
+
+  // Use Manchester code for asynchronous transmission
+  tpiu_set_trace_output_mode(TPIU_SPPR_ASYNC_MANCHESTER);
```
karlp: If you're writing examples, a) still use C style comments, and b) use an example that's useful to people, so that's not likely to be manchester.

lib/cm3/itm.c
```diff
@@ -0,0 +1,156 @@
...
+  if (enable) {
+    ITM_TCR |= ITM_TCR_ITMENA;
+  } else {
+    ITM_TCR &= ~ITM_TCR_ITMENA;
+  }
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief Enable a given set of stimulus ports (ITM_PORTN) */
+void itm_enable_stimulus_ports(uint32_t ports) { ITM_TER[0] |= ports; }
```
karlp: please write functions out properly.

lib/cm3/itm.c
```diff
@@ -0,0 +1,156 @@
...
+/** @brief Directly configure the mask of enabled ITM ports (ITM_PORTN) */
+void itm_set_stimulus_port_mask(uint32_t mask) { ITM_TER[0] = mask; }
+
+/*---------------------------------------------------------------------------*/
+/** @brief Write a byte to the given ITM stimulus port */
+void itm_send_blocking_u8(uint8_t port, uint8_t datum) {
+  // Wait for stimulus port to become available
+  while (!(ITM_STIM8(port) & ITM_STIM_FIFOREADY)) {
+  }
+  // Write datum
```
karlp: most of these comments seem superfluous, and are using c++ style as well.

lib/stm32/f7/Makefile
```diff
@@ -55,13 +55,16 @@ OBJS		+= flash_common_all.o flash_common_f.o flash_common_f24.o flash.o
 OBJS		+= fmc_common_f47.o
 OBJS		+= gpio_common_all.o gpio_common_f0234.o
 OBJS		+= i2c_common_v2.o
+OBJS		+= itm.o
 OBJS		+= iwdg_common_all.o
-OBJS		+= ltdc_common_f47.o 
+OBJS		+= ltdc_common_f47.o
```
karlp: unrelated whitespace changes should be dropped

karlp:
I'm not really against this sort of thing, but I fairly strongly feel that this gives people the wrong impression, namely that this is code you include in your own application.  People do this, it works, but it's..... just so backwards!  It takes up code space in your application, hard limits you to the specific settings you chose.

You should let your debugger interface do this for you directly, which is how openocd, orbuculum, BMP all do it.  At the _least_ it should be strongly worded that there is virtually never a good reason to be using this code.

rschlaikjer:
Hmm, perhaps I have been labouring under the wrong impression then! At least from BMP, I was unaware of a way to enable and configure ITM / TRACESWO output on a target that didn't otherwise have firmware to deal with it. If it is possible to do that, then perhaps the bulk of this code can indeed be discarded, and what's left (mostly just named wrappers around writing to ITM registers) may not be worth keeping.

karlp:
BMP isn't very clear on this itself, and it's "using swo" doc actualyl has code to do this statically, but also mentions you can do this via gdb memory accesses.  Orbuculum, which is the tool I use for parsing the swo outputs, has pretty good docs, and provides a gdbinit file withh a bunch of gdb functions that can do ~all of these steps at will.  https://github.com/orbcode/orbuculum

Openocd, it's just built into the server directly.  THey don't have _allll_ the variations available as commands, but you can things like the below in your config file
```
tpiu config internal swodump.log uart off 32000000
```
and pass that to orbuculum's -f file input.   Openocd provides commands like "itm port <port> <on|off>" and similar, with port 0 being enabled by default.  You can do ~same as the orbuculum gdb functions with openocd too, or you can provide them in tcl functions directly, and then you don't need gdb at all.

openocd doc is at http://openocd.org/doc/html/Architecture-and-Core-Commands.html#index-tpiu-config

karlp:
Do you still want to add anything here? Should we keep this open or close it?

rschlaikjer:
I'll close this - thanks for the tips on orbuculum, and I think you're right that most of this doesn't really need to be in code here.

karlp:
Please feel very free to follow up if you want any help with tracing, it's awesome and I wish more people knew how "easy" it was to use.  #libopencm3 on irc.freenode.org, and I believe libopencm3/discuss on gitter too if you want some more interactive help with it.

PR #1055, opened 2019-05-24
Add .a suffix to LIBDEPS var in genlink-config

rschlaikjer:
It looks like commit 96d094af13dab67709a252a39da164bde4d637e0
introduced a build failure to the gadget for the tilm4f120xl target,
which appears to be the only one to use genlink.

Adding back the `.a` suffix to the rule appears to fix the problem.


karlp:
Yeha, this was my bad, it's fixed now.

PR #1059, opened 2019-06-08
doc: no need to call sed -s, unbreak make doc on os x

grevaillot:
Not sure how/why i pushed that, absolutely no need to call sed with -s
parameter. Fix build on os x / systems running non gnu sed.


karlp:
merged in b7fc86c7 thanks!

PR #1060, opened 2019-06-11
Add DAC support for the STM32L4

ervanalb:
Tested on STM32 NUCLEO-L432 board


include/libopencm3/stm32/l4/memorymap.h
```diff
@@ -62,6 +62,7 @@
 #define USB_PMA_BASE			(PERIPH_BASE_APB1 + 0x6c00)
 #define POWER_CONTROL_BASE		(PERIPH_BASE_APB1 + 0x7000)
 #define DAC1_BASE			(PERIPH_BASE_APB1 + 0x7400)
+#define DAC_BASE			DAC1_BASE
```
ervanalb: I could have just renamed `DAC1_BASE` but I did it this way so as to not break anybody's code that is relying on `DAC1_BASE` existing. Happy to switch it back if you prefer.

karlp: We've done similar things on F0 that only has one dac as well.  If l4 only has the one dac, then this is perfectly fine here too.

karlp: Though, if there was no user of DAC before, can we just always have DAC1?

karlp: Actually, looking at the rest of the library, I think DAC1 should be dropped.  until/if they really do bring out second dac.  Everyone else uses DAC_BASE.

karlp:
I've merged this without the DAC1_BASE, as mentioned above.  thanks for this!

PR #1061, opened 2019-06-11
stm32: dma cselr enhancement

grevaillot:
Hi,

Here are a couple of patches regarding dma channel selection register, (DMA_CSELR) present on l0, l4 and f09x. I moved the register defintion out of device dma header to a new dma_common_cselr.h, added shift/mask register usage macros and dma_set_channel_request function.


lib/stm32/common/dma_common_csel.c
```diff
@@ -0,0 +1,46 @@
...
+
+/*---------------------------------------------------------------------------*/
+/** @brief DMA Channel Set Request Selection
+
+Set DMA request mapping selection for given channel. Refer to datasheet for channel
+request mapping tables.
+
+@param[in] dma unsigned int32. DMA controller base address: DMA1 or DMA2
+@param[in] channel unsigned int8. Channel number:  1-7 for DMA1 or 1-5 for DMA2
+@param[in] request unsigned int8. DMA request mapping.
```
karlp: My only comment here, because you're doing such good stuff in general, is that the type information here in the doxygen is redundant.  the types are in the signature, and it just takes up space here in the comments that can be better used for the description.

lib/stm32/l0/Makefile
```diff
@@ -43,6 +43,7 @@ OBJS            += gpio_common_all.o gpio_common_f0234.o rcc_common_all.o
 OBJS		+= adc_common_v2.o
 OBJS		+= crs_common_all.o
 OBJS		+= dma_common_l1f013.o
+OBJS		+= dma_common_csel.o
```
karlp: You can put all the dma objects on the same line. (same below)

grevaillot:
Thanks for the review, I updated the PR,

karlp:
this has merged now, via e06898d9 eb0bc099 and 7d344b18 tahnks!

PR #1062, opened 2019-06-11
stm32: rng enhancements

grevaillot:
Hi,

Here are a couple of patches i had on my tree regarding rng handling, adding a couple of helpers and handling potential errors in a nicer way, based on RM recomandations.

### DIFF
--- a/lib/stm32/common/rng_common_v1.c
+++ b/lib/stm32/common/rng_common_v1.c
@@ -44,6 +44,20 @@ void rng_enable(void)
     RNG_CR |= RNG_CR_RNGEN;
 }
 
+/** Enable the Random Number Generator error interrupt.
+*/
+void rng_interrupt_enable(void)
+{
+    RNG_CR |= RNG_CR_IE;
+}
+
+/** Disable the Random Number Generator error interrupt.
+*/
+void rng_interrupt_disable(void)
+{
+    RNG_CR &= ~RNG_CR_IE;
+}
+
 /** Randomizes a number (non-blocking).
  * Can fail if a clock error or seed error is detected. Consult the Reference
  * Manual, but "try again", potentially after resetting the peripheral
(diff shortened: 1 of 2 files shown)


lib/stm32/common/rng_common_v1.c
```diff
@@ -73,17 +87,29 @@ bool rng_get_random(uint32_t *rand_nr)
...
         uint32_t rv;
         bool done;
         do {
-                if (RNG_SR & RNG_SR_SECS) {
-                        rng_disable();
-                        rng_enable();
+
+                if (RNG_SR & RNG_SR_SEIS) {
+                        RNG_SR = RNG_SR & ~RNG_SR_SEIS;
+                        for (int i = 12; i != 0; i--)
```
karlp: we still want {} here.

grevaillot: my bad, fixed

lib/stm32/common/rng_common_v1.c
```diff
@@ -94,10 +94,16 @@ uint32_t rng_get_random_blocking(void)
...
-                        rng_disable();
-                        rng_enable();
+
+                if (RNG_SR & RNG_SR_SEIS) {
+                        RNG_SR = RNG_SR & ~RNG_SR_SEIS;
+                        for (int i = 12; i != 0; i--) {
+                                rv = RNG_DR;
+                        }
+                        RNG_CR &= ~RNG_CR_RNGEN;
+                        RNG_CR |= RNG_CR_RNGEN;
```
karlp: My only concern with this series is this one.  I'm really not sure that "both" is a good compromise.  I kinda feel like if you do the disable/enable, there was no point in the readout, and the readout of 12 is in the g0 manual as a speed up?  _or_ it's a different implementation? (more doubtful)  I just... feel weird doing both.

karlp: eh, I'm not using the peripheral.  I'm going to go with your judgement on this.

karlp:
merged all in e2ac1a63..a8a92b4c thanks.  still feels weird doing both approaches, but better than the prior code certainly, and you're using it, so that counts for more! thanks!

PR #1064, opened 2019-06-13
stm32g0: support++

grevaillot:
Hi,

Here's some more g0 support, making it a bit more usefull - adding iwdg/usart/spi/i2c/rng/crc and timer.
Not much new stuff here, a couple of new regs and bits for Timer and RNG but that's all.

I need to review adc and dma/dmamux stuff, but that will wait the next patchset

### DIFF
--- a/lib/stm32/g0/Makefile
+++ b/lib/stm32/g0/Makefile
@@ -44,6 +44,7 @@ OBJS += usart_common_all.o usart_common_v2.o
 OBJS += spi_common_all.o spi_common_v1_frf.o
 OBJS += i2c_common_v2.o
 OBJS += rng_common_v1.o
+OBJS += crc_common_all.o
 
 VPATH +=../:../../cm3:../common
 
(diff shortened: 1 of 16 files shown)


ChuckM:
+1

lib/stm32/g0/Makefile
```diff
@@ -39,6 +39,7 @@ OBJS += flash.o flash_common_all.o
 OBJS += gpio_common_all.o gpio_common_f0234.o 
 OBJS += pwr.o
 OBJS += rcc.o rcc_common_all.o
+OBJS += iwdg_common_all.o
```
karlp: this is actually alphabetically sorted now, let's try and keep it that way, easier to find things.

karlp:
Looks good, I will sort the .o names when I merge it, no need to repush anything.

karlp:
merged, there's a few things like TIM5 not existing, but those are problems with the timer_common-all common code, not this patch, so ... "out of scope"  thanks for this!

PR #1065, opened 2019-06-14
stm32: l0: timer/lptimer: doc+

grevaillot:
So, i had that for a while in a tree here : improve doc for l0 lptim (in timer.h), and add MASK defines - let's push it before someone else does it :)

### DIFF
--- a/include/libopencm3/stm32/l0/timer.h
+++ b/include/libopencm3/stm32/l0/timer.h
@@ -37,8 +37,9 @@ LGPL License Terms @ref lgpl_license
 
 #include <libopencm3/stm32/common/timer_common_all.h>
 
-/* Timer 2/21/22 option register (TIMx_OR) */
+/**@{*/
 
+/** Timer 2/21/22 option register (TIMx_OR) */
 #define TIM_OR(tim_base)		MMIO32((tim_base) + 0x50)
 #define TIM2_OR				TIM_OR(TIM2)
 #define TIM21_OR			TIM_OR(TIM21)
@@ -114,6 +115,8 @@ LGPL License Terms @ref lgpl_license
 #define LPTIM1_ARR			LPTIM_ARR(LPTIM1_BASE)
 #define LPTIM1_CNT			LPTIM_CNT(LPTIM1_BASE)
 
+/** @defgroup lptim_isr LPTIM_ISR Interrupt and Status Register
+@{*/
 #define LPTIM_ISR_CMPM			(1 << 0)
 #define LPTIM_ISR_ARRM			(1 << 1)
 #define LPTIM_ISR_EXTTRIG		(1 << 2)
@@ -121,7 +124,10 @@ LGPL License Terms @ref lgpl_license
 #define LPTIM_ISR_ARROK			(1 << 4)
 #define LPTIM_ISR_UP			(1 << 5)
 #define LPTIM_ISR_DOWN			(1 << 6)
+/**@}*/
 
+/** @defgroup lptim_icr LPTIM_ICR Interrupt Clear Register 
+@{*/
 #define LPTIM_ICR_CMPMCF		(1 << 0)
 #define LPTIM_ICR_ARRMCF		(1 << 1)
 #define LPTIM_ICR_EXTTRIGCF		(1 << 2)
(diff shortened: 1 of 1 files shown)


include/libopencm3/stm32/l0/timer.h
```diff
@@ -138,31 +138,53 @@ LGPL License Terms @ref lgpl_license
 #define LPTIM_IER_UPIE			(1 << 5)
 #define LPTIM_IER_DOWNIE		(1 << 6)
 
+/* --- LPTIM_CFGR values ----------------------------------------------------- */
```
karlp: If you're going to add a "blah values" section, you might as well make it a doxygen group.  We've done that in other places, so there's a group for all the values of each register.  Just this ----- line doesn't really do much.

include/libopencm3/stm32/l0/timer.h
```diff
@@ -138,31 +138,53 @@ LGPL License Terms @ref lgpl_license
...
+
+/* CKSEL: Select internal (0) or external clock source (1) */
```
karlp: you need a /** to make doxygen pick it up

include/libopencm3/stm32/l0/timer.h
```diff
@@ -138,31 +138,53 @@ LGPL License Terms @ref lgpl_license
...
 #define LPTIM_CFGR_CKSEL		(1 << 0)
 
 #define LPTIM_CFGR_CKPOL_SHIFT		1
+#define LPTIM_CFGR_CKPOL_MASK		0x03
 #define LPTIM_CFGR_CKPOL		(3 << LPTIM_CFGR_CKPOL_SHIFT)
+/** @defgroup lptim_cfgr_ckpol LPTIM_CFGR CKPOL Clock Polarity
+* @ingroup timer_defines
```
karlp: You can skip the @ingroup when it's naturally inside that group.  You only need to explicitly do an @ingroup if you want to put it "elsewhere"

include/libopencm3/stm32/l0/timer.h
```diff
@@ -182,30 +209,47 @@ LGPL License Terms @ref lgpl_license
...
+/* @def LPTIM_CFGR_TRIGEN LPTIM_CFGR TRIGEN Trigger enable and polarity
+* @ingroup timer_defines
+@{*/
 #define LPTIM_CFGR_TRIGEN_SW		(0 << LPTIM_CFGR_TRIGEN_SHIFT)
 #define LPTIM_CFGR_TRIGEN_RISING	(1 << LPTIM_CFGR_TRIGEN_SHIFT)
 #define LPTIM_CFGR_TRIGEN_FALLING	(2 << LPTIM_CFGR_TRIGEN_SHIFT)
 #define LPTIM_CFGR_TRIGEN_BOTH		(3 << LPTIM_CFGR_TRIGEN_SHIFT)
+/**@}*/
 
+/* TIMOUT: Timeout enable */
```
karlp: Same for all of these, they need the second **

include/libopencm3/stm32/l0/timer.h
grevaillot: hmm, do you have some good doc example/guideline to follow to keep things consistent ? I aligned with timer_common_all.h :)

include/libopencm3/stm32/l0/timer.h
grevaillot: ahhh, indeed. thanks

include/libopencm3/stm32/l0/timer.h
grevaillot: hmm, that does not seem to work here without explicit ingroup :( - not sure how/why right now, but same thing is done in timer_common_all.h. I'll have a look.

grevaillot:
ok, i fixed the defines comment to have them picked up and grouped registers to have something nicer, but i still need to add some ingroup @timer_defines to have them in the right place.

include/libopencm3/stm32/l0/timer.h
```diff
@@ -182,30 +218,52 @@ LGPL License Terms @ref lgpl_license
 /* 5 is reserved */
 #define LPTIM_CFGR_TRIGSEL_EXT_TRIG6	(6 << LPTIM_CFGR_TRIGSEL_SHIFT)
 #define LPTIM_CFGR_TRIGSEL_EXT_TRIG7	(7 << LPTIM_CFGR_TRIGSEL_SHIFT)
+/**@}*/
 
 #define LPTIM_CFGR_TRIGEN_SHIFT		17
+#define LPTIM_CFGR_TRIGSEL_MASK		0x07
 #define LPTIM_CFGR_TRIGEN		(3 << LPTIM_CFGR_TRIGEN_SHIFT)
+/* @defgroup LPTIM_CFGR_TRIGEN LPTIM_CFGR TRIGEN Trigger enable and polarity
```
karlp: hah, this is what was giving me troubles with groupings, it's missing the second *  I've fixed it locally, no need to repush anything.

karlp:
this has landed already, I fixed up the extr abits locally when I merged it. thanks!

PR #1066, opened 2019-06-17
.gitignore: ignore autogenerated include/libopencmsis/swm050/

grevaillot:
.gitignore: ignore autogenerated include/libopencmsis/swm050/


karlp:
merged via rebase thanks!

PR #1069, opened 2019-06-18
stm32l0: consistent makefile.

grevaillot:
i found out adding lptim basic support that l0 Makefile has not been aligned : let's do it before adding new stuff :)


karlp:
nice thanks, not sure how I missed it :)  too many files in too many windows I guess.

PR #1070, opened 2019-06-18
stm32: extract l0 lptimer stuff from timer.h to common lptimer.h, enable it for f4,f7 and g0 too.

grevaillot:
lptimer peripheral is present on f4,f7,l0,l4,g0 and prob others. Extract content from stm32l0 timer.h and make it usable by other chips.

Issue : currently, and as far as i understand, common headers without common code are not indexed by doxygen - gendoxylist only pickups chip's header, and headers extracted from .d files. Am I wrong ?


include/libopencm3/stm32/lptimer.h
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/stm32/memorymap.h>
+
+#if defined(STM32F4)
+#       include <libopencm3/stm32/f4/lptimer.h>
+#elif defined(STM32L0)
+#       include <libopencm3/stm32/l0/lptimer.h>
+#elif defined(STM32l4)
```
karlp: needs to be an uppercase L here I believe

karlp:
You mention f7, but yoru dispatch header only has f4?

include/libopencm3/stm32/lptimer.h
```diff
@@ -0,0 +1,36 @@
...
+#       include <libopencm3/stm32/l4/lptimer.h>
+#elif defined(STM32G0)
+#       include <libopencm3/stm32/g0/lptimer.h>
+#else
```
karlp: (I'm not going to get upset about it, but this commit is first, and includes these l4/g0 files that don't exist until the subsequent commit, if you want perfect style bonus points, you enable them here as you add the files)

grevaillot:
Ok, i saw your answer on irc - i was looking for #1048 :) 

I've got some lptimer code indeed, just not exactly ready for PR, But i'm not sure i'll have some time to prepare/clean it. I wanted to push that first, but guess this can wait a bit.

And, Hmm, yeah, definitely some fixes before merge ;) - I never used the draft function i was not even sure the pr would have been readable by all actually, it went a bit too fast :)

include/libopencm3/stm32/lptimer.h
```diff
@@ -22,8 +22,16 @@
 #include <libopencm3/cm3/common.h>
 #include <libopencm3/stm32/memorymap.h>
 
-#if defined(STM32L0)
+#if defined(STM32F4)
+#       include <libopencm3/stm32/f4/lptimer.h>
```
karlp: I'm still not sure I saw an f4 file?  (but yes, this is all much nicer so far)

grevaillot: yeah - actually there's no lptim_base referenced in current f4 memorymap, and i got lost looking at RM (I never worked on that chip), and pushed to my branch before leaving .. - I'll update the PR after fixing that.

grevaillot:
closing that one, will push a new PR with some lptimer basic support

Issue #957, opened 2018-08-24
CortexM0: NVIC priority registers support only word addressing


mattande:
nvic_set_priority() handles all IPR register accesses using byte addressing, however on Cortex M0 all NVIC registers must only be accessed using word addressing.

See Cortex M0 TRM 5.1. About the NVIC:
All NVIC registers are only accessible using word transfers. Any attempt to read or write a halfword or byte individually is Unpredictable.

karlp:
so, what does "unpredictable" actually do in practise?  (just curious, as you must have run into this)

karlp:
oh right, I merged thsi one, and only commented on the other one, instead of here.  thanks for cleaning up behind me!

PR #1072, opened 2019-06-19
CortexM0: IPR and SHPR are only word addressable

mattande:
Fix for issue #957 

For ARMv6M, the IPR and SHPR registers are accessible only when
adddressed with a 32bit word read or write.

Currently in libopencm3 all NVIC interrupt priority register accesses
are made using an 8bit read or write, which results in the hardware
ignoring the write or always returning 0 on read.

Address this by introducing NVIC_IPR32() and SCS_SHPR32() macro and
conditional implementation of nvic_set_priority when building for
cortex-m0.

See ARMv6M developer documentation:
IPR: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0497a/Cihgjeed.html
SHPR: http://infocenter.arm.com/help/index.jsp?topic=/com.arm.doc.dui0497a/CIAGECDD.html

### DIFF
--- a/include/libopencm3/cm3/nvic.h
+++ b/include/libopencm3/cm3/nvic.h
@@ -86,9 +86,14 @@
 
 /* IPR: Interrupt Priority Registers */
 /* Note: 240 8bit Registers */
-/* Note: 32 8bit Registers on CM0 */
+/* Note: 8 32bit Registers on CM0, requires word access */
+#if defined(__ARM_ARCH_6M__)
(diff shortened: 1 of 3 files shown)


karlp:
What's wonderful about these is that ARM DDI 0419 rev C, and also rev D both mention that these registers have no restrictions on them.  Only the web versions you linked mention this.  Thanks ARM

karlp:
This has merged via d8579dde thanks!

PR #1074, opened 2019-06-26
stm32f0, stm32f1, gd32f1x0: dedup RCC code

Icenowy:
The RCC source code has many duplications in STM32F0x, STM32F1x and
GD32F1x0.

Extract all the common source codes to a rcc_common_f01.c source code.

The APIs of these chips' RCC is made consistent in this progress.


include/libopencm3/stm32/f1/rcc.h
```diff
@@ -233,18 +233,19 @@
...
 @ingroup STM32F1xx_rcc_defines
 
 @{*/
-#define RCC_CFGR_SW_SYSCLKSEL_HSICLK		0x0
-#define RCC_CFGR_SW_SYSCLKSEL_HSECLK		0x1
-#define RCC_CFGR_SW_SYSCLKSEL_PLLCLK		0x2
+#define RCC_CFGR_SW_HSI				(0 << RCC_CFGR_SW_SHIFT)
+#define RCC_CFGR_SW_HSE				(1 << RCC_CFGR_SW_SHIFT)
+#define RCC_CFGR_SW_PLL				(2 << RCC_CFGR_SW_SHIFT)
+
```
karlp: I generally prefer to keep these defines unshifted.  You've got the PLL factors unshifted, the original bits here were unshifted, and you've changed all of them to be shifted, is there any reason I can't see for that?

karlp: (To clarify, my general rule is that single bits are defined shifted, but fields, _especially_ when the field value is relevant, should all be unshiftted)

lib/stm32/common/rcc_common_f01.c
```diff
@@ -0,0 +1,589 @@
...
+
+/**@{*/
+
+#include <libopencm3/cm3/assert.h>
+#include <libopencm3/stm32/rcc.h>
+
+/* Set the default clock frequencies */
+uint32_t rcc_ahb_frequency = 8000000; /* 8MHz after reset */
+uint32_t rcc_apb1_frequency = 8000000; /* 8MHz after reset */
+#ifdef RCC_HAS_APB2
```
karlp: can this just be based on a #if defined RCC_APB2_ENR or similar for instance? as is done below here with PLL2 and PLL3?

karlp:
rcc_backupdomain_reset is now gone?  I don't have a strong opinion on that, if it's not considered useful, but the docs still refer to it in @ref links.

include/libopencm3/stm32/f1/rcc.h
Icenowy: I just followed the current F0 code.

But of course I can unshift them, including F0.

Icenowy:
@karlp backup domain is now a RST_ enum item.

include/libopencm3/stm32/f1/rcc.h
karlp: Yes, I'd rather you copied the f1 code on this, where it's unshifted.

karlp:
ok, the thing with backupdomain reset is that there is doxygen references that refer to it, and the function still exists on other families.  I think adding it to the reset macros is clever, but it might be something a little out of scope of what you're doing here maybe?

Icenowy:
@karlp I'm not considering other STM32's now (mainly because I don't have them).

Looks like I need a rebase here.

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -338,7 +338,12 @@
 
 #define RCC_BDCR_BDRST				(1 << 16)
 #define RCC_BDCR_RTCEN				(1 << 15)
-/* RCC_BDCR[9:8]: RTCSEL */
+#define RCC_BDCR_RTCSEL_SHIFT			8
+#define RCC_BDCR_RTCSEL				(3 << RCC_BDCR_RTCSEL_SHIFT)
```
karlp: Classicaly, this gets the suffix _MASK

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -338,7 +338,12 @@
...
+#define RCC_BDCR_RTCSEL_NOCLK			(0 << RCC_BDCR_RTCSEL_SHIFT)
+#define RCC_BDCR_RTCSEL_LSE			(1 << RCC_BDCR_RTCSEL_SHIFT)
+#define RCC_BDCR_RTCSEL_LSI			(2 << RCC_BDCR_RTCSEL_SHIFT)
+#define RCC_BDCR_RTCSEL_HSE			(3 << RCC_BDCR_RTCSEL_SHIFT)
```
karlp: And these would then be unshifted

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -393,6 +398,16 @@
 #define RCC_CFGR3_ADCSW				(1 << 8)
 #define RCC_CFGR3_CECSW				(1 << 6)
 
+/* --- RCC_CR2 values ------------------------------------------------------ */
+
+#define RCC_CR2_HSI14CAL_SHIFT			8
+#define RCC_CR2_HSI14CAL			(0xFF << RCC_CR2_HSI14CAL_SHIFT)
```
karlp: this would then be ```RCC_CR2_HSI14CAL_MASK 0xff```

include/libopencm3/gd32/f1x0/rcc.h
```diff
@@ -393,6 +398,16 @@
...
+#define RCC_CR2_HSI14TRIM_SHIFT			3
+#define RCC_CR2_HSI14TRIM			(31 << RCC_CR2_HSI14TRIM_SHIFT)
```
karlp: "31" here is unusual, it would normally be hex, and normally be _MASK and unshifted.

include/libopencm3/stm32/f0/rcc.h
```diff
@@ -563,6 +564,7 @@ int rcc_css_int_flag(void);
 void rcc_set_sysclk_source(enum rcc_osc clk);
 void rcc_set_usbclk_source(enum rcc_osc clk);
 void rcc_set_rtc_clock_source(enum rcc_osc clk);
+uint32_t rcc_rtc_clock_enabled_flag(void);
```
karlp: couldn't this just be bool? (it is a boolean function right?)

karlp:
I feel like there's a lot going on here for a single commit, if you cna split it, awesome, otherwise I'll wade through it.

PR #1076, opened 2019-07-01
Add STM32L4+ support

Meumeu:
This is mostly based on STM32L4, except the SDMMC controller is on the AHB bus instead of the APB bus, and there is a LCD TFT controller and a MIPI DSI host like STM32F4 and STM32F7.


grevaillot:
nice - Fyi. I've got some support for dmamux to finish on g0, l'll have a look at l4+'s peripheral, i think it's pretty much the same.

include/libopencm3/stm32/adc.h
```diff
@@ -36,6 +36,8 @@
 #       include <libopencm3/stm32/l1/adc.h>
 #elif defined(STM32L4)
 #       include <libopencm3/stm32/l4/adc.h>
+#elif defined(STM32L4PLUS)
+#       include <libopencm3/stm32/l4/adc.h>
```
karlp: I don't like this approach. I think it's why you landed in problems with the VPATH.  I don't like the idea of l4 plus doing it's own thing for shared code vs all the other stm32's shared code.

include/libopencm3/stm32/l4plus/dsi.h
```diff
@@ -0,0 +1,31 @@
+/** @defgroup dsi_defines DSI Defines
+ * @brief <b>Defines Constants and Macros for the STM32F7xx Display Serial
+ *           Interface Host and Wrapper</b>
+ * @ingroup STM32F7xx_defines
```
karlp: you must be careful when copying and pasting.

include/libopencm3/stm32/l4plus/ltdc.h
```diff
@@ -0,0 +1,31 @@
+/** @defgroup ltdc_defines LTDC Defines
+ * @brief <b>Defined Constants and Types for the STM32F7xx LCD TFT Display
+ *           Controller</b>
+ * @ingroup STM32F7xx_defines
```
karlp: here too

include/libopencm3/stm32/l4plus/memorymap.h
```diff
@@ -0,0 +1,141 @@
...
+#define TIM8_BASE			(PERIPH_BASE_APB2 + 0x3400)
+#define USART1_BASE			(PERIPH_BASE_APB2 + 0x3800)
+#define TIM15_BASE			(PERIPH_BASE_APB2 + 0x4000)
+#define TIM16_BASE			(PERIPH_BASE_APB2 + 0x4400)
+#define TIM17_BASE			(PERIPH_BASE_APB2 + 0x4800)
+#define SAI1_BASE			(PERIPH_BASE_APB2 + 0x5400)
+#define SAI2_BASE			(PERIPH_BASE_APB2 + 0x5800)
+#define DFSDM_BASE			(PERIPH_BASE_APB2 + 0x6000)
+#define LCD_TFT_BASE			(PERIPH_BASE_APB2 + 0x6800)
+#define LTDC_BASE			(PERIPH_BASE_APB2 + 0x6800) /* compat. with f4 */
```
karlp: don't duplicate the address, just do LTDC_BASE LCD_TFT_BASE

include/libopencm3/stm32/l4plus/rcc.h
```diff
@@ -0,0 +1,996 @@
+/** @defgroup rcc_defines RCC Defines
+ *
+ * @ingroup STM32L4xx_defines
+ *
+ * @brief <b>Defined Constants and Types for the STM32L4xx Reset and Clock
+ * Control</b>
```
karlp: and here

include/libopencm3/stm32/ltdc.h
```diff
@@ -24,6 +24,8 @@
 #       include <libopencm3/stm32/f4/ltdc.h>
 #elif defined(STM32F7)
 #       include <libopencm3/stm32/f7/ltdc.h>
+#elif defined(STM32L4PLUS)
+#       include <libopencm3/stm32/l4plus/ltdc.h>
 #else
-#       error "LCD-TFT only defined for STM32F4"
+#       error "LCD-TFT only defined for STM32F4, STM32F7 and STM32L4+"
```
karlp: safeer to just "not defined for this family" or similar rather than trying to list them all each time

karlp:
In general, please don't add fix commits in a series that adds the code with the errors. You submit code that is correct, fix commits only come if the bad code is already committed.

If you want to share code between these families, you need to share it, not just reach into the other one and hope it will work.  Use the same style we're already using for common code, and if you really feel it should be separate, you can make a common_l4 directory

Meumeu:
> In general, please don't add fix commits in a series that adds the code with the errors. You submit code that is correct, fix commits only come if the bad code is already committed.
> 
> If you want to share code between these families, you need to share it, not just reach into the other one and hope it will work. Use the same style we're already using for common code, and if you really feel it should be separate, you can make a common_l4 directory

Yes, I saw the errors that needed to be fixed after I created the PR, I'll clean up the history when I'm done. For the shared code, you're right, I'll put them in xyz_common_l4.h/.c files

karlp:
Please don't rename files like include/libopencm3/stm32/common/dma2d_common_f47.h → include/libopencm3/stm32/common/dma2d_common_f47_l4.h  You can already see how bad that approach was by how many files you have to rename to even try and keep them in sync.  As a standalone commit you can use a functional name or peripheral version, like you'll with things like i2c_v1 and v2, and adc, and adc_multi and so on, but the blanket renaming just makes churn that's harder to merge and track and keep uptodate.

include/libopencm3/stm32/common/ltdc_common_f47_l4.h
```diff
@@ -7,8 +7,8 @@
  *
  * @date 5 December 2014
  *
- * This library supports the LCD TFT display controller (LTDC) in the STM32F4xx
- * and STM32F7xx series of ARM Cortex Microcontrollers by ST Microelectronics.
+ * This library supports the LCD TFT display controller (LTDC) in the STM32F4xx,
+ * STM32F7xx, and STM32L4xx series of ARM Cortex Microcontrollers by ST Microelectronics.
```
karlp: If you want reviewers to have sane chance of reviewing this, please break this up into smaller steps. You're trying to add L4+.  This is fixing comments in a file that's already used by L4 and already wrong, and will also be used by L4+.  It's _correct_ to fix it, don't get me wrong, but it's got to be in separate commits (see also my comment about renaming files)

include/libopencm3/stm32/common/pwr_common_l4.h
```diff
@@ -0,0 +1,174 @@
+/** @addtogroup pwr_defines PWR Defines
+ *
+ * @ingroup STM32L4xx_defines
```
karlp: this is going to break documentation for l4+ unless I'm very mistaken.

include/libopencm3/stm32/common/pwr_common_l4.h
```diff
@@ -0,0 +1,174 @@
...
+ *
+ * @brief <b>Defined Constants and Types for the STM32L4xx Power Control</b>
```
karlp: What I've started doing with EFM32 is to not include the partname here :) just say "for the Power Control module"

include/libopencm3/stm32/l4/flash.h
```diff
@@ -1,21 +1,21 @@
 /** @defgroup flash_defines FLASH Defines
- *
- * @ingroup STM32L4xx_defines
  *
  * @brief <b>Defined Constants and Types for the STM32L4xx Flash Control</b>
  *
+ * @ingroup STM32L4xx_defines
+ *
```
karlp: you've moved this for no reason, it's just churn

include/libopencm3/stm32/l4/flash.h
```diff
@@ -1,21 +1,21 @@
...
+ * @date 2 July 2019
  *
  * LGPL License Terms @ref lgpl_license
- *  */
+ */
 /*
  * This file is part of the libopencm3 project.
  *
- * Copyright (C) 2016 Benjamin Levine <benjamin@jesco.karoo.co.uk>
+ * Copyright (C) 2019 Guillaume Meunier <guillaume.meunier@centraliens.net>
```
karlp: So as I understand it, you moved this entire file to the common_l4_flash file, where it kept benajamin's copyright (good) and you've then put your own copyright on the empty file remaining?  I would that is, at least poor form, and somewhat disingenuous, but it's probably still _legal_ to copyright the #include line.  I'd lean towards dropping the claim entirely, (Benjamin's moves to the common file) or leaving benjamin's.  Certainly until you've added more meaningful content to this file.

include/libopencm3/stm32/l4/syscfg.h
```diff
@@ -6,16 +6,14 @@
  *
  * @version 1.0.0
  *
- * @date 28 December 2017
+ * @date 2 July 2019
```
karlp: I'm not convinced this is correct either.  l4/syscfg has definitely existed since dec 2017.

include/libopencm3/stm32/l4plus/rcc.h
```diff
@@ -0,0 +1,1127 @@
...
+#define RCC_APB1SMENR1			MMIO32(RCC_BASE + RCC_APB1SMENR1_OFFSET)
+#define RCC_APB1SMENR2_OFFSET		0x7c
+#define RCC_APB1SMENR2			MMIO32(RCC_BASE + RCC_APB1SMENR2_OFFSET)
+#define RCC_APB2SMENR_OFFSET		0x80
+#define RCC_APB2SMENR			MMIO32(RCC_BASE + RCC_APB2SMENR_OFFSET)
+#define RCC_CCIPR			MMIO32(RCC_BASE + 0x88)
+#define RCC_BDCR			MMIO32(RCC_BASE + 0x90)
+#define RCC_CSR				MMIO32(RCC_BASE + 0x94)
+#define RCC_CRRCR			MMIO32(RCC_BASE + 0x98)
+#define RCC_CCIPR2			MMIO32(RCC_BASE + 0x9c)
```
karlp: is none of this shared with l4?  You can have partial sharing, it doesn't have to be just entire files shared

lib/stm32/f7/Makefile
```diff
@@ -46,15 +46,15 @@ OBJS += crc_common_all.o crc_v2.o
 OBJS += dac_common_all.o
 OBJS += desig.o
 OBJS += dma_common_f24.o
-OBJS += dma2d_common_f47.o
-OBJS += dsi_common_f47.o
+OBJS += dma2d_common_f47_l4.o
+OBJS += dsi_common_f47_l4.o
```
karlp: see how all these changes stack up when you start renaming files? :)

lib/stm32/l4plus/pwr.c
```diff
@@ -0,0 +1,85 @@
+/** @defgroup pwr_file PWR peripheral API
+ *
+ * @ingroup peripheral_apis
+ *
+ * @brief <b>libopencm3 STM32L4xx Power Control</b>
+ *
+ * @version 1.0.0
+ *
+ * @author @htmlonly &copy; @endhtmlonly 2016 Benjamin Levine <benjamin@jesco.karoo.co.uk>
```
karlp: If this file is benjamin's from 2016, shouldn't it all be shared?  are the functions here actually all new work just for l4plus?  In that case you probably _do_ want to change dates and copyrights here.

include/libopencm3/stm32/l4/flash.h
Meumeu: Yes, I moved the entire file. I can remove the copyright line as it's only a single #include (in the other files as well).

include/libopencm3/stm32/l4plus/rcc.h
Meumeu: Most of the #defines can be shared, the enums will have to stay separated obviously as some peripherals moved from APB to AHB.

lib/stm32/l4plus/pwr.c
Meumeu: pwr_enable_backup_domain_write_protect and pwr_disable_backup_domain_write_protect are from Benjamin, pwr_set_vos_scale is based on his version but ST added scale 1 boost on top of scale 1 and scale 2.

include/libopencm3/stm32/common/ltdc_common_f47_l4.h
Meumeu: I've started splitting it into several commits at the same time I rebased on the last master branch to fix the merge conflict.

lib/stm32/l4plus/pwr.c
karlp: so shouldn't those two functions go into the shared file then?

karlp:
Have just gotten back to this, will re-review all updates. sorry about the delay :(

karlp:
I saw that this had new commits, but I see it's just rebased.  I'm not at _all_ convinced that this massive copy paste is appropriate.  What's _actually_ different? is it just nvic, part of rcc?  It seems like there's been just wholesale moving l4 code to common, and then copy/pasting headers?

PR #1078, opened 2019-07-02
doc: stm32: adc: upgrade common_v2 documentation

grevaillot:
I took some time to fixup a bit adc common_v2 header documentation (g0 uses that peripheral), to have something nicer.. 

so : add register grouping, fixup comment have them pickedup by doxygen, align style and masks. I also added some empty group stubs (ie, adc_cgfr2), to allow some addgroups from device's adc header.

### DIFF
--- a/include/libopencm3/stm32/common/adc_common_v2.h
+++ b/include/libopencm3/stm32/common/adc_common_v2.h
@@ -36,138 +36,181 @@ specific memorymap.h header before including this header file.*/
 #define LIBOPENCM3_ADC_COMMON_V2_H
 
 /* ----- ADC registers  -----------------------------------------------------*/
-/* ADC interrupt and status register */
+/** ADC interrupt and status register */
 #define ADC_ISR(adc)		MMIO32((adc) + 0x00)
-/* Interrupt Enable Register */
+/** Interrupt Enable Register */
 #define ADC_IER(adc)		MMIO32((adc) + 0x04)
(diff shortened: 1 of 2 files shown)


include/libopencm3/stm32/common/adc_common_v2.h
```diff
@@ -36,138 +36,179 @@ specific memorymap.h header before including this header file.*/
...
 
-/* DMAEN: Direct memory access enable */
+/** DMAEN: Direct memory access enable */
 #define ADC_CFGR1_DMAEN		(1 << 0)
 
+/**@}*/
+
+/* ADC_SMPR Values -----------------------------------------------------------*/
+/** @defgroup adc_smpr SMPR ADC sample time register
+@{*/
```
karlp: you open a group here without closing it?  and really, if you're not putting things in, shouldn't we just drop them?

grevaillot: oops, fixed - pushed too quick, updated pr with closed group.

My idea here was: we define a register, let's define also the maching doxygen group too, and avoid doing it in each device's adc.h when defining register bit fields. ppl can then just @addtogroup, like i did in g0 adc. But it can look a bit weird, indeed, when no single bit it shared between chips. Up to you, i can drop that.

karlp:
merged tahnks!  we'll try this predef groups out, see how it goes

PR #1079, opened 2019-07-02
stm32g0: add adc.

grevaillot:
v2 "single" peripheral with a couple of tweaks :
 - added registers to configure two additionnal advanced analog watchdog.
 - different adc sampling time time based on channel groups.
 - 8 steps adc sequence injection, using chselr/chselrmode.

A note on the rm explains that after every configuration change to ADC_CFGR1's
SCANDIR or CHSELRMOD or CHSELR register, user need to check that configuration
is applied before any other modification / adc conversion start.. 

This makes adc_set_reqular and adc_set_injected sequence a bit painfull to read, but, well..

### DIFF
--- a/include/libopencm3/stm32/g0/adc.h
+++ b/include/libopencm3/stm32/g0/adc.h
@@ -1,0 +1,325 @@
+/** @defgroup adc_defines ADC Defines
+ *
+ * @ingroup STM32G0xx_defines
+*
+ * @author @htmlonly &copy; @endhtmlonly 2019 Guillaume Revaillot <g.revaillot@gmail.com>
+ *
+ * @brief <b>Defined Constants and Types for the STM32STM32G0xx Analog to Digital Converter</b>
+ *
+ * @version 1.0.0
+ *
+ * LGPL License Terms @ref lgpl_license
+ *  */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**@{*/
+
+#ifndef LIBOPENCM3_ADC_H
+#define LIBOPENCM3_ADC_H
+
+#include <libopencm3/stm32/common/adc_common_v2.h>
+#include <libopencm3/stm32/common/adc_common_v2_single.h>
+
+/** @defgroup adc_reg_base ADC register base addresses
+ *@{*/
+#define ADC1				ADC1_BASE
+/**@}*/
+
+/** @defgroup adc_channel ADC Channel Numbers
+ *@{*/
+#define ADC_CHANNEL_TEMP	12
+#define ADC_CHANNEL_VREF	13
+#define ADC_CHANNEL_VBAT	14
+/**@}*/
+
+/* ----- ADC registers  -----------------------------------------------------*/
+/** ADC_AWD1TR Watchdog 1 Threshold register */
+#define ADC_AWD1TR(adc)		MMIO32((adc) + 0x20)
+/** ADC_AWD2TR Watchdog 2 Threshold register */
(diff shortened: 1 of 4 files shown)


include/libopencm3/stm32/g0/adc.h
```diff
@@ -0,0 +1,327 @@
...
+@{*/
+
+#define ADC_CCR_PRESC_MASK		(0xf)
+#define ADC_CCR_PRESC_SHIFT		(18)
+/** @defgroup adc_ccr_presc ADC clock prescaler
+ *@{*/
+#define ADC_CCR_PRESC_NODIV		(0x0)
+#define ADC_CCR_PRESC_DIV1		(0x1)
+#define ADC_CCR_PRESC_DIV2		(0x2)
+#define ADC_CCR_PRESC_DIV6		(0x3)
```
karlp: 6? what bizarre dividers

include/libopencm3/stm32/g0/adc.h
```diff
@@ -0,0 +1,327 @@
...
+
+/* SMP1 ADC Channel Sample Time selection */
+#define ADC_SMPR_SMPSEL_SHIFT			0x8
+#define ADC_SMPR_SMPSEL_MASK			0x7ffff
+#define ADC_SMPR_SMPSEL_CHANNEL_SHIFT(channel)	((channel) + ADC_SMPR_SMPSEL_SHIFT)
+#define ADC_SMPR_SMPSEL_CHANNEL_MASK			(1)
+/** @defgroup adc_smpr_smpsel ADC Sample Time selection
+@{*/
+#define ADC_SMPR_SMPSEL_SMP1	0x0
+#define ADC_SMPR_SMPSEL_SMP2	0x1
```
karlp: oh boi, just read the refman on this stuff.  quite different from the other families :)

lib/stm32/g0/adc.c
```diff
@@ -0,0 +1,251 @@
...
+ * sequence of channels to be converted. ADCSTART must be de-asserted before
+ * sequence setup.
+ * Note that on STM32G0, the sequence is limited to 8 channels, and
+ * only the first 14 ADC channels are selectable.
+ *
+ * @param[in] adc ADC base address (@ref adc_reg_base)
+ * @param[in] length Number of channels in the group, range 0..8
+ * @param[in] channel Set of channels in sequence (0..14 or @ref adc_channel)
+ */
+void adc_set_injected_sequence(uint32_t adc, uint8_t length, uint8_t channel[])
```
karlp: I understand why you went to use this api for the "other" way of doing sequences on g0, (just the first 14, and only 8 channels) but the "injected" name has a very specific meaning on the other families, and I think it's misleading to try and use that name here.

I don't know what _to_ do, but these two sequence limitions are quite awkward

grevaillot: hmm, It looked like the f3 injected stuff - i'll have a look at the refman later.
I pushed a patch without the "injected" stuff meanwhile, to merge the remaining stuff.

grevaillot:
hi,

I removed the (not) "injected" bits in last push to get that merge, went to holiday and did not really pushed for - would you prefer to have that back with proper naming ? They are calling that discontinous mode, so i'd go for s/adc_set_injected_sequence/adc_set_discontinous_sequence ?

jaxxzer:
I'm having trouble with adc here: https://github.com/jaxxzer/stm32-boilerplate/blob/g0/src/example/adc/main.c

The adc value is always reading zero. @grevaillot have you tested this? Can you take a quick look?

jaxxzer:
@grevaillot have you had a chance to take another look at this? I'll have some time this week to try.

grevaillot:
hi !
I overlooked your message here - I'm currently on medical leave so i can not have a look right now, but it should be working :) - IIRC i think you need to take care of adc clock source to have something working, but if sampling ends and result is 0 I guess ADC is properly clock feeded.

jaxxzer:
I missed setting up CHSELR, thanks!

grevaillot:
great !

grevaillot:
> I removed the (not) "injected" bits in last push to get that merge, went to holiday and did not really pushed for - would you prefer to have that back with proper naming ? They are calling that discontinous mode, so i'd go for s/adc_set_injected_sequence/adc_set_discontinous_sequence ?

Ping, anything i can do to have these bits (without the set injected/discontinous part) merged ? :)

karlp:
as we spoke on irc, if you can fix the doxygen warnings so that the ADC stuff turns up properly, I can merge this.  just run ```make doc``` locally and check that the g0 page looks "right" so, no broken definitions for adc stuff at the bottom, and the adc peripheral docs having docs for both the common-v2 code and the specific adc code for g0.  You can check doc/stm32g0/doxygen_stm32g0.log for the warnings/errors doxygen spits out

karlp:
merged in 38b45c87861e5461283609f3aa2df31b141faae6 thanks

PR #1086, opened 2019-07-27
F4 rcc

jhalmen:
I'm working on a project with the f401. i've come across a few things i thought worth changing / adding.

short explanation for the current commits:
* the removed functionality can be accessed through `rcc_osc_on(RCC_PLLSAI);` respectively `rcc_is_osc_ready(RCC_PLLSAI);`
* add short wrapper to configure the i2s pll available on the stm32f401

### DIFF
--- a/include/libopencm3/stm32/f4/rcc.h
+++ b/include/libopencm3/stm32/f4/rcc.h
@@ -1080,6 +1080,7 @@ void rcc_osc_on(enum rcc_osc osc);
 void rcc_osc_off(enum rcc_osc osc);
 void rcc_css_enable(void);
 void rcc_css_disable(void);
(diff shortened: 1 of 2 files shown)


karlp:
Looks perfectly reasonable to me.  Are those removed functions also present on f7 or anywhere where we could drop them too?

jhalmen:
i just checked, and i can't seem to find any.
but then again, f4 is the only family where i could find *any* pllsai functions

OpenUAS:
Any reason not to merge?

karlp:
This has now merged.  I changed the commit titles to match style, I changed the commit comments to include the replacement functions, I changed the documentation on range of values, as f405 et al also have this register, with different range of values.

PR #1088, opened 2019-08-06
Linker: Add definition for .heap to let malloc work

ChuckM:
This adds a symbol (heap_start) at the end of allocated
RAM for the start of heap.

### DIFF
--- a/ld/linker.ld.S
+++ b/ld/linker.ld.S
@@ -132,6 +132,10 @@ SECTIONS
 		. = ALIGN(4);
 		_ebss = .;
 	} >ram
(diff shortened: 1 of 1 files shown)


karlp:
so, how is this different from _ebss? I don't get who would tag sections as .heap?  I feel like we discussed this somewhere, but it never got written down?  What am I missing here?

karlp:
Any more info?  I'm still not sure why _ebss isn't enough? is it just that you had some code that was expecting a "heap_start" symbol?  Please re-open if you have more info and this is still important

PR #1092, opened 2019-08-27
stm32g0: fix typos in memorymap, impacting tim1 and tim15-17.

grevaillot:
bad copypaste, and unfortunatelly these timers where not tested when I enabled support for g0..


karlp:
well, really, the typo was that you kept the APB1/APB2 split, when the g0 only talks about having "apb"  I guess fixing this is "right" short term, even if it's confusing really to keep this style.

karlp:
merged for now, we can work out naming later.

grevaillot:
indeed.. i pushed a pr to fix that too.

PR #1094, opened 2019-08-28
stm32g0: memorymap: get rid of apb1/apb2 reference, device only has only one apb.

grevaillot:
I apparently based memorymap.h on previously written header without noticing that g0 has only one apb despite a big hole in the memory space and addresses matching usual apb1/apb2 split..


karlp:
cool, this was easier than I'd hoped.  seems it was already msotly taken care of: https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/stm32/g0/rcc.h#L617-L622

karlp:
merged in 998e647d thanks

PR #1095, opened 2019-08-28
stm32g0: use proper registers for gpio peripherals clock sleep enable.

grevaillot:


karlp:
indeed. merged  in 8a1cfa8c after dropping an extra space. thanks!

PR #1101, opened 2019-09-10
stm32:L4:flash: Add erase bank2 flash function.

balanceTWK:
I have compared several different flash reference manuals for the L4 series, which I think is necessary.

😀


karlp:
thanks very much for this, sorry about the delay even replying :(  I've fixed the codestyle for you, and merged it in 6fc1ff22.  You can use "make stylecheck" to test style changes locally fwiw.

PR #1103, opened 2019-09-21
SWM050: Finishes GPIO, IAP flash, sysclock, sleep/stop, and the sysctl memory map.

Szalacinski:
Updates the main memory map and the makefile.
Adds the SWM050 to devices.data, so that a linker script can be automatically generated.

The only things left to be done with the SWM050 are the TIMERSE and WDT peripherals, which are in progress and mostly complete.


Szalacinski:
I just realized that there is still an issue that was brought up in #1024, which is that SYSCON needs to be in its own file.  Should I go ahead and fix this or just wait for my next PR that adds TIMERSE and WDT support?

include/libopencm3/swm050/clk.h
```diff
@@ -0,0 +1,60 @@
...
+
+/* Various Clock Defines */
+/** @defgroup clk_misc_defs Various Clock Defines
+
+@{*/
+#define DEFAULT_BOOT_FREQ	(9000000U)
+#define CYCLES_PER_US		(sclk_freq/1000000)
+#define CYCLES_PER_MS		(sclk_freq/1000)
+#define M36			(36000000U)
+#define M18			(18000000U)
```
karlp: all of these are inappropriately named, they should all be CLK_xxxx something.  We've also never offered macros like CYCLES_PER_US and CYCLES_PER_MS for any other family, so I don't really want to start trying it now.  It's also misleading with the way cortex-m pipelines work.  Consider which of these are even really needed in the core header.

include/libopencm3/swm050/clk.h
```diff
@@ -0,0 +1,60 @@
...
+@{*/
+#define CLK_36MHZ		(0x1)
+#define CLK_18MHZ		(0x0)
+/*@}*/
+
+BEGIN_DECLS
+
+uint32_t clk_speed(bool mhz, uint16_t div);
+uint32_t clk_init(bool mhz, uint16_t div);
+//void clk_en(bool en);
```
karlp: don't include commented out code.

include/libopencm3/swm050/clk.h
```diff
@@ -0,0 +1,60 @@
...
+/** @defgroup clk_speed Base Clock Speeds
+@{*/
+#define CLK_36MHZ		(0x1)
+#define CLK_18MHZ		(0x0)
+/*@}*/
+
+BEGIN_DECLS
+
+uint32_t clk_speed(bool mhz, uint16_t div);
+uint32_t clk_init(bool mhz, uint16_t div);
```
karlp: I really hope the difference in these becomes apparent later :)

include/libopencm3/swm050/flash.h
```diff
@@ -0,0 +1,42 @@
...
+/**@{*/
+#ifndef LIBOPENCM3_FLASH_H
+#define LIBOPENCM3_FLASH_H
+
+#include <libopencm3/cm3/common.h>
+
+BEGIN_DECLS
+
+uint32_t flash_write(uint32_t *dest, uint32_t *src, uint8_t cnt);
+uint32_t flash_read(uint32_t *src, uint32_t *dest, uint8_t cnt);
```
karlp: is this a swm050 pecularity that flash isn't memory mapped?

include/libopencm3/swm050/gpio.h
```diff
@@ -47,52 +44,74 @@
...
 
 /* GPIO direction definitions */
 /** @defgroup gpio_dir GPIO Pin Direction
 @{*/
-#define GPIO_INPUT			0x0
-#define GPIO_OUTPUT			0x1
-/**@}*/
+#define GPIO_INPUT			(0x0)
+#define GPIO_OUTPUT			(0x1)
+/*@}*/
```
karlp: drop all the gratuitous changes in this file please.  No need for bracketing, and no need for dropping doxygen comments even if it happens to work sometimes with some versions of doxygen

include/libopencm3/swm050/clk.h
Szalacinski: On boot, the internal oscillator needs time to stabilize, but only needs to happen once, so any changes to the multiplier and divider do not require the wait time.

include/libopencm3/swm050/gpio.h
```diff
@@ -47,52 +44,74 @@
...
+
+/* GPIO polarity definitions */
+/** @defgroup gpio_pol GPIO Polarity
+@{*/
+#define GPIO_LOW			(0x0)
+#define GPIO_HIGH			(0x1)
```
karlp: you should make this something like GPIO_POL_ or GPIO_POLARITY_ I would think, and same for the irq triggers, so you know which one you're using where.  There's a hacking file that says how to name defines.

include/libopencm3/swm050/gpio.h
```diff
@@ -47,52 +44,74 @@
...
+#define GPIO_UNMASKED			(0x0)
+#define GPIO_MASKED			(0x1)
+/*@}*/
+
+/* GPIO Registers */
 /** @defgroup gpio_registers GPIO Registers
 @{*/
 /** Data register */
-#define GPIO_DATA			MMIO32(GPIO_BASE + 0x0)
+#define ADATA				MMIO32(GPIO_BASE + 0x0)
```
karlp: definitely don't do this.  Consistency across the families is important, and particularly when you get things like "INTMASK" you really need to know that that's the GPIO_INTMASK, not the TIMER_INTMASK or something else.

include/libopencm3/swm050/flash.h
Szalacinski: Yes.  I'm fairly certain it is a separate peripheral.  I tried some writes to flash that replied with the correct status code, but they didn't show up on a dump of the main flash.  If it is memory-mapped, it is not documented where.

include/libopencm3/swm050/gpio.h
```diff
@@ -47,52 +44,74 @@
...
-/** GPIO Alternat function selection register */
-#define GPIO_SEL			MMIO32(SYSTEM_CON_BASE + 0x80)
+/** GPIO Alternate function selection register */
+#define PORTA_SEL			MMIO32(SYSTEM_CON_BASE + 0x80)
 /** GPIO Pull up register */
-#define GPIO_PULLUP			MMIO32(SYSTEM_CON_BASE + 0x90)
+#define PORTA_PULLUP			MMIO32(SYSTEM_CON_BASE + 0x90)
 /** GPIO Input enable register */
-#define GPIO_INEN			MMIO32(SYSTEM_CON_BASE + 0xe0)
+#define PORTA_INEN			MMIO32(SYSTEM_CON_BASE + 0xe0)
```
karlp: these could potentially be changed to GPIO_PORTA_XXXX but I'd probably say to just leave them as is for now.

include/libopencm3/swm050/gpio.h
```diff
@@ -106,11 +125,19 @@ void gpio_input(uint16_t gpios);
 void gpio_output(uint16_t gpios);
 void gpio_sel_af(uint16_t gpios, bool af_en);
 void gpio_pullup(uint16_t gpios, bool en);
-void gpio_in_en(uint16_t gpios, bool en);
+void gpio_input_en(uint16_t gpios, bool en);
```
karlp: I know it's a new file, but changing apis should generally be accompanied with the explanation somewhere for it.

include/libopencm3/swm050/sysctl.h
```diff
@@ -0,0 +1,50 @@
...
+#ifndef LIBOPENCM3_SYSCTL_H
+#define LIBOPENCM3_SYSCTL_H
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/swm050/memorymap.h>
+
+/** @defgroup sysctl_register SYSCTL Registers
+ * @note System configuration registers
+ * @{*/
+/** Clock dividers for TIMERSE and SCLK */
+#define SYS_CFG_0			MMIO32(SYSCTL_BASE + 0x0)
```
karlp: if you're using SYSCTL_BASE in the sysctl file, then the defines should be SYSCTL_xxx

lib/swm050/Makefile
```diff
@@ -33,7 +33,7 @@ TGT_CFLAGS      += $(DEBUG_FLAGS)
 TGT_CFLAGS	+= $(STANDARD_FLAGS)
 ARFLAGS		= rcs
 
-OBJS += gpio.o
+OBJS += gpio.o clk.o flash.o pwr.o
```
karlp: this is fine, but to avoid conflicts merging this sort of thing, I've moved most other families to one file/common set per line, in alphabetical order, so
```
OBJS += clk.o
OBJS += flash.o
OBJS += gpio.o
....
```

lib/swm050/clk.c
```diff
@@ -0,0 +1,117 @@
...
+#include <libopencm3/swm050/sysctl.h>
+
+
+/*---------------------------------------------------------------------------*/
+/** @brief Change the system clock multiplier and divider
+
+Change system clock speed.
+
+@param[in] mhz Base clock speed @ref clk_speed
+	     The base clock speed, before the clock divider
```
karlp: I've got to say, a bool here feels like a very non-obvious argument type to a user.  espectially when the description is "the speed"  Could you explain this a bit more?  Maybe just to me, maybe it's obvious to the end user, but it feels like a clunky api like this.

lib/swm050/clk.c
```diff
@@ -0,0 +1,117 @@
...
+		SYS_DBLF &= ~BIT0;
+		sclk_tmp = M18;
+	}
+	
+	if (div <= 1) {
+		SYS_CFG_0 |= BIT0;
+	} else {
+		uint32_t div_masked = (div & ~0xFFFFFC01);
+		SYS_CFG_0 = (SYS_CFG_0 & 0xFFFFFC00) | div_masked;
+		sclk_tmp /= div_masked;
```
karlp: honestly, I'd drop the return of the clock speed.  the user is going to know what they were setting, and you're forcing a division on a cortex m0 here for them, just for a return value that will, in all likelihood, be ignored.

lib/swm050/clk.c
Szalacinski: There are 2 possible states for the clock multiplier, stored in a bit.  the clk_speed defines provide a way for the user to choose one of these.  It is an odd type, but there are only 2 possible options.

lib/swm050/clk.c
```diff
@@ -0,0 +1,117 @@
...
+	if ((mhz == CLK_18MHZ) && (div <= 1)) {
+		return M18;
+	}	
+
+	return clk_speed(mhz, div);
+}
+
+/* Function should not be used on the SWM050, but can be used
+         on other Synwit ARM MCUs that support external oscillators.
+	 Disables the internal oscillator. */
```
karlp: just drop this code until you send a PR for other synwit parts please.

lib/swm050/clk.c
```diff
@@ -0,0 +1,117 @@
...
+	     Passes value to mhz in @ref clk_setupr
+
+@param[in] div Clock divider
+	     Passes value to div in @ref clk_setup
+
+@return The current clock value in Hz.
+
+*/
+uint32_t clk_init(bool mhz, uint16_t div)
+{
```
karlp: wat.  Am I going to have to go and dig up the manual for this thing again?  If you have to use this function, why doyou need the other exposed as a public api?  How did you figure out the nop length? can you use a memory barrier?  is there a flag to check?

lib/swm050/flash.c
```diff
@@ -0,0 +1,95 @@
...
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**@{*/
+#include <libopencm3/swm050/flash.h>
+
+#define IAP_WR	(void *)(0x1000AB)
+#define IAP_E	(void *)(0x100127)
```
karlp: aren't these things a user might want exposed from flash.h too?

lib/swm050/flash.c
```diff
@@ -0,0 +1,95 @@
...
+*/
+uint32_t flash_write(uint32_t *dest, uint32_t *src, uint8_t cnt)
+{
+	return iap_write_read(dest, src, cnt, 1);
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief Read from the user flash
+
+Reads words from the 0.5k user flash area.
```
karlp: ooooh k :)

lib/swm050/flash.c
karlp: I see what you're doing with them now, but they could be commented internally as poitners to ROM APIs?

lib/swm050/gpio.c
```diff
@@ -187,11 +183,129 @@ is enabled, GPIO and AF of the SWD pins will be both unavailable.
...
+
+@param[in] gpios Pin identifiers @ref gpio_pin_id
+	     If multiple pins are to be changed, use bitwise OR '|' to separate
+	     them.
+*/
+void gpio_int_en(uint16_t gpios)
+{
+	for (uint8_t i = 0; 0 <= 9; ++i) {
+		if (gpios & (1 << i)) {
+			nvic_enable_irq(NVIC_GPIOA0_IRQ + i);
```
karlp: traditionally on other families, we've only enabled the irqs in the peripheral blocks, and user code hsa been responsible for nvic layer stuff.  I'd like to see that either preserved, or a discussion on it across the whole tree.

lib/swm050/gpio.c
```diff
@@ -187,11 +183,129 @@ is enabled, GPIO and AF of the SWD pins will be both unavailable.
...
+void gpio_int_mask(uint16_t gpios, bool masked)
+{
+	INTMASK_A = masked ? (INTMASK_A | gpios) : (INTMASK_A & ~gpios);
+}
+
+
+/*---------------------------------------------------------------------------*/
+/** @brief Sets whether the pins are edge triggered or level triggered
+
+Sets whether the pins are edge triggered or level triggered.  Edge-triggered
```
karlp: as a general doxygen tip, you can drop the @brief, and just remember it uses the first sentence up to the .  Can make things simpler to write and read.

karlp:
I've not run the doxygen generation itself to look for warnings and errors yet, I'll do that before I merge regardless.  Looks mostly good though.

lib/swm050/flash.c
Szalacinski: There isn't any documented use for these outside of these 3 functions.  Do you think they should be exposed?

include/libopencm3/swm050/gpio.h
Szalacinski: Would GPIO_ADATA be good?  Some of the chips in the SWM family have registers that go beyond port A.

include/libopencm3/swm050/gpio.h
Szalacinski: It was changed to distinguish it more from the interrupt enable function.  Should I note that in the doxygen or just change it back?

lib/swm050/gpio.c
Szalacinski: Fixed.

lib/swm050/clk.c
Szalacinski: Fixed.

lib/swm050/clk.c
Szalacinski: Fixed.

lib/swm050/Makefile
Szalacinski: Fixed.

include/libopencm3/swm050/sysctl.h
Szalacinski: Fixed.

include/libopencm3/swm050/gpio.h
Szalacinski: Fixed.

include/libopencm3/swm050/gpio.h
Szalacinski: Fixed.

include/libopencm3/swm050/clk.h
Szalacinski: Fixed.

include/libopencm3/swm050/clk.h
Szalacinski: Fixed.

include/libopencm3/swm050/gpio.h
Szalacinski: Fixed.

include/libopencm3/swm050/gpio.h
Szalacinski: Fixed.

lib/swm050/flash.c
Szalacinski: Fixed.

include/libopencm3/swm050/clk.h
karlp: I think that's crazy having to have two functions just for that. this can't be the best way of doing things.

include/libopencm3/swm050/flash.h
karlp: if I understand from the later .c file (and the datasheet), this is not "flash" in the sense of the "flash.c" functions in any other families.  This is for a special  0.5k "user flash" region, separate from the normal data section.  I'm not sure what call it, and I know synwit just calls it "data flash" vs "program flash" so *shrugs* but it seemed confusing.

lib/swm050/clk.c
karlp: still, the values aren't "truthy" and "falsey" they're an enum of length 2.  I at least find the idea of a bool for this argument confusing and non-obvious.

include/libopencm3/swm050/clk.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef LIBOPENCM3_CLK_H
+#define LIBOPENCM3_CLK_H
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/swm050/memorymap.h>
+
+/* Various Clock Defines */
+/** @defgroup clk_misc_defs Various Clock Defines
+
+@{*/
+#define DEFAULT_BOOT_FREQ	(9000000U)
```
karlp: what is this used for? in other families, that have used a global to store the clock speed, it just gets set to that, and edited, this define doesn't really seem terribly useful

Szalacinski: That's the default boot frequency according to the datasheet.  I can see why it could be removed, though.

lib/swm050/clk.c
```diff
@@ -0,0 +1,95 @@
...
+		SYSCTL_SYS_DBLF |= BIT0;
+	} else {
+		SYSCTL_SYS_DBLF &= ~BIT0;
+	}
+	
+	if (div <= 1) {
+		SYSCTL_SYS_CFG_0 |= BIT0;
+	} else {
+		uint32_t div_masked = (div & ~0xFFFFFC01);
+		SYSCTL_SYS_CFG_0 = (SYSCTL_SYS_CFG_0 & 0xFFFFFC00) | div_masked;
```
karlp: have a look at how functions like this are written https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/l4/rcc.c#L281-L324 and see if you can avoid these custom and diverse masks

include/libopencm3/swm050/clk.h
Szalacinski: Do you think that a static bool should be used in one merged function to keep track of whether the function has already been run once?

include/libopencm3/swm050/clk.h
Szalacinski: Fixed

lib/swm050/clk.c
Szalacinski: Fixed.

lib/swm050/clk.c
Szalacinski: Fixed.

include/libopencm3/swm050/clk.h
Szalacinski: Fixed.

lib/swm050/clk.c
Szalacinski: The nop length was based on information from Synwit's site.  There is no flag to check.

include/libopencm3/swm050/gpio.h
```diff
@@ -57,42 +54,64 @@
...
+/* SYSCON Registers */
 /** @defgroup syscon_register SYSCON Registers
  * @note These registers are really part of the SYSCON system control space
  * @{*/
 /** SWD Enable register */
 #define SWD_SEL				MMIO32(SYSTEM_CON_BASE + 0x30)
-/** GPIO Alternat function selection register */
-#define GPIO_SEL			MMIO32(SYSTEM_CON_BASE + 0x80)
+/** GPIO Alternate function selection register */
+#define PORTA_SEL			MMIO32(SYSTEM_CON_BASE + 0x80)
```
karlp: SYSCON_PORTA_SEL, and really, shoudl be moved to it's own file.  Same for SYSCON_PORTA_PULLUP and SYSCON_PORTA_INEN

lib/swm050/gpio.c
```diff
@@ -187,11 +185,123 @@ is enabled, GPIO and AF of the SWD pins will be both unavailable.
...
+
+When interrupts are masked, it prevents them from being received, which is a
+quicker way to turn on and off GPIO interrupts (after calling gpio_int_en()).
+
+@param[in] gpios Pin identifiers @ref gpio_pin_id
+	     If multiple pins are to be changed, use bitwise OR '|' to separate
+	     them.
+	     
+@param[in] gpios Pin mask selection @ref gpio_int_masked
+	     Whether to mask or unmask pins.
```
karlp: param doesn't match argument

lib/swm050/gpio.c
```diff
@@ -187,11 +185,123 @@ is enabled, GPIO and AF of the SWD pins will be both unavailable.
...
+/** @brief Sets the pins as external interrupts, rather than normal GPIO
+
+Enable interrupts on the selected pins.  If you want to quickly
+switch on and off interrupts, use gpio_int_mask() after calling this.
+
+@param[in] gpios Pin identifiers @ref gpio_pin_id
+	     If multiple pins are to be changed, use bitwise OR '|' to separate
+	     them.
+*/
+void gpio_int_en(uint16_t gpios)
```
karlp: I _really_ want this to be gpio_irq_enable or gpio_int_enable if you must.  Also, there's no method to ever clear these bits.  I know that you say that using the mask function is "faster" but there's still no way to turn them off.

lib/swm050/gpio.c
```diff
@@ -187,11 +185,123 @@ is enabled, GPIO and AF of the SWD pins will be both unavailable.
...
+interrupt bits must be cleared by software.
+
+@param[in] gpios Pin identifiers @ref gpio_pin_id
+	     If multiple pins are to be changed, use bitwise OR '|' to separate
+	     them.
+
+@param[in] type Trigger Type @ref gpio_trig_type
+	     Level or edge triggered
+*/
+void gpio_int_type(uint16_t gpios, bool type)
```
karlp: this isnt' a bool anymore is it?

lib/swm050/gpio.c
```diff
@@ -187,11 +185,123 @@ is enabled, GPIO and AF of the SWD pins will be both unavailable.
...
+Sets whether the interrupt is triggered by a high or low level/edge.
+
+@param[in] gpios Pin identifiers @ref gpio_pin_id
+	     If multiple pins are to be changed, use bitwise OR '|' to separate
+	     them.
+
+@param[in] pol Polarity @ref gpio_pol 
+	     High or low level/edge
+*/
+void gpio_int_pol(uint16_t gpios, bool pol)
```
karlp: same here,

lib/swm050/gpio.c
```diff
@@ -187,11 +185,123 @@ is enabled, GPIO and AF of the SWD pins will be both unavailable.
...
+{
+	GPIO_INTPOLARITY_A = pol ? (GPIO_INTPOLARITY_A | gpios) : (GPIO_INTPOLARITY_A & ~gpios);
```
karlp: did "make stylecheck" not complain about this?  this isn't saving any code, it just makes the lines harder to read

include/libopencm3/swm050/gpio.h
```diff
@@ -106,11 +125,19 @@ void gpio_input(uint16_t gpios);
...
+void gpio_input_en(uint16_t gpios, bool en);
 
 void gpio_sel_swd(bool en);
 
+void gpio_int_en(uint16_t gpios);
+void gpio_int_mask(uint16_t gpios, bool masked);
+void gpio_int_type(uint16_t gpios, bool level);
+void gpio_int_pol(uint16_t gpios, bool pol);
+uint16_t gpio_int_stat(void);
+uint16_t gpio_int_raw_stat(void);
```
karlp: can these both be _status instead of _stat?

lib/swm050/pwr.c
```diff
@@ -0,0 +1,55 @@
...
+/** @brief Go into sleep mode
+
+Goes to sleep and wakes up on when GPIO pin 9 is pulled low.  Please give
+yourself enough time to connect the debugger before calling this, in order to
+not get locked out of the MCU.
+
+*/
+void pwr_sleep(void)
+{
+	SYSCTL_SYS_CFG_2 |= BIT4;
```
karlp: no, BIT4 should be defined to what it is.

lib/swm050/pwr.c
```diff
@@ -0,0 +1,55 @@
...
+
+Completely stops the MCU.  No SWD, nothing.  No documentation exists for this.
+Please use pwr_sleep() if you want to save power.  If you absolutely need to use
+this, please make sure that you give yourself enough time to connect via SWD and
+stop execution.  Included for completeness.
+
+*/
+void pwr_stop(void)
+{
+	SYSCTL_SYS_CFG_2 = 0x30;
```
karlp: I don't think this function should exist honestly.  the register and the bit definition exist for masochists.

karlp:
Merged after rebase.  I put the irq_enable back to int, it needs to be all or nothing, and int is fine too.  other families seem to have a mix of irq, int _and_ interrupt. so *shrugs*  lets let this target move forward at least.

Issue #234, opened 2014-01-09
stm32l1 medium+ and high density parts have wrong unique id address


karlp:
Should be 0x1FF800CC
This was fixed for most parts in https://github.com/libopencm3/libopencm3/pull/211 but the m+ and h density parts of the L1 family bizarrely have a different base address.  And libopencm3 doesn't have any way of separating this parts at this time.

PR #1105, opened 2019-09-25
Refactored desig to support F7 with different desig reg addresses

matthewlai:
Refactored desig to support F7 having different reg addresses depending on device ID in DBGMCU


lib/stm32/common/desig_common_all.c
```diff
@@ -19,35 +19,22 @@
...
-	len = (2 * sizeof(dev_id_buf) < string_len) ?
-		2 * sizeof(dev_id_buf) : string_len - 1;
+	len = (2 * sizeof(uid_buf) < string_len) ?
+		2 * sizeof(uid_buf) : string_len - 1;
 
 	for (i = 0; i < len; i += 2) {
-		string[i]     = chars[(device_id[i / 2] >> 4) & 0x0F];
-		string[i + 1] = chars[(device_id[i / 2] >> 0) & 0x0F];
+		string[i]     = chars[(uid[i / 2] >> 4) & 0x0F];
+		string[i + 1] = chars[(uid[i / 2] >> 0) & 0x0F];
```
karlp: this is all gratuitous in this function, you can just drop it.

lib/stm32/common/desig_common_all.c
```diff
@@ -66,15 +53,18 @@ void desig_get_unique_id_as_string(char *string,
...
-	serial[2] = id[9];
-	serial[3] = id[8] + id[0];
-	serial[4] = id[7];
-	serial[5] = id[6];
+	serial[0] = uid[11];
+	serial[1] = uid[10] + uid[2];
+	serial[2] = uid[9];
+	serial[3] = uid[8] + uid[0];
+	serial[4] = uid[7];
+	serial[5] = uid[6];
```
karlp: same here, it might be "nice" but it just makes churn for no real particular reason in unrelated code.

lib/stm32/f7/desig.c
```diff
@@ -0,0 +1,71 @@
...
+      flash_size_base = (uint32_t*)DESIG_FLASH_SIZE_BASE_451;
+      break;
+    case 0x0452:
+      flash_size_base = (uint32_t*)DESIG_FLASH_SIZE_BASE_452;
+      break;
+  }
+
+  if (!flash_size_base) {
+    /* We don't know the address for this device. Hang here to help debugging. */
+    for(;;);
```
karlp: perhaps nicer to return 0?  we have a cm3_assert you could use too, cm3_assert_not_reached iirc, or similar.

karlp:
I like what you've done here.  we should continue with it, to fix existing bugs such as https://github.com/libopencm3/libopencm3/issues/234 (and iirc, it's also a problem within some other families too, not just l1 and f7)

lib/stm32/common/desig_common_all.c
matthewlai: I thought it would be nice to change the names here, since "device ID" actually means something else (the device ID field in DBGMCU). I feel it would be especially confusing to anyone tracing this code now since that other device ID is actually used to decide the address now. I can revert them if you want, though. Is there a benefit to not changing this? It's an internal name that no one should really care about?

lib/stm32/common/desig_common_all.c
matthewlai: This is just to make them match the above.

lib/stm32/f7/desig.c
matthewlai: I thought about returning 0, but thought it would be nicer to fail loudly rather than silently. What do you think?

cm3_assert_not_reached() is nice. I'll change it tonight when I'm back home.

lib/stm32/common/desig_common_all.c
karlp: yeah, I mean, whatever, it's just generally bad form to make unrelated changes.  it makes the diffs much bigger and harder to review, and we end up wasting time on this sort of cruft instead of the issues itself.

matthewlai: Ah I had these changes in a separate commit before, but squashed them for the PR. Will keep them separate next time.

lib/stm32/f7/desig.c
matthewlai: Done.

karlp:
merged thanks! (I fixed your whitespace for you)

PR #1109, opened 2019-10-03
devices.data: add missing stm32f302?e*

kidonsky:


karlp:
Thanks for this, I added the d variant, and indeed, went and updated all the stm32f3 variants too.

PR #1111, opened 2019-10-12
stm32l0: add CRC support

nmrossomando:
Compiles in the common CRC code and adds a header.

Tested on an STM32L021


karlp:
merged in 77d96a34 thanks!

PR #1113, opened 2019-10-17
NVIC/Cortex-M0: IPR and SCS_SHPR access must be words!

UweBonnes:


karlp:
this has whitespace problems at least.  will look at the rest later  Fixes https://github.com/libopencm3/libopencm3/issues/957

lib/cm3/nvic.c
```diff
@@ -161,6 +163,23 @@ void nvic_set_priority(uint8_t irqn, uint8_t priority)
 		/* Device specific interrupts */
 		NVIC_IPR(irqn) = priority;
 	}
+#else
+        volatile uint32_t *block;
+        unsigned int index;
+	if (irqn < NVIC_IRQ_COUNT) {
```
karlp: is this relying on it being unsigned and that the negatives will appear to be higher?

lib/cm3/nvic.c
```diff
@@ -143,6 +143,7 @@ uint8_t nvic_get_irq_enabled(uint8_t irqn)
  *
  * There are 4 priority levels only, given by the upper two bits of the
  * priority byte, as required by ARM standards. No grouping available.
+ * Only word access allowed!
```
karlp: given that people are using the api for this, this comment is meaningless for them.

lib/cm3/nvic.c
UweBonnes: Same code as original code

lib/cm3/nvic.c
UweBonnes: Unclear wording. Is with regard to the Cortex-M registers, not the API access.

UweBonnes:
I am happy with #1072 to. Should I rework my patch or do you regard #1072 better,

karlp:
re whitespace, you get "make stylecheck" and try not to make things worse.  I'll look at both of them tonight.

karlp:
merged the first implentation in d8579dde thanks!

PR #1118, opened 2019-10-26
Enables DEVICES_DATA per project overrides

### LINKED ISSUE #1117: Bootloader and genlink.py / devices.data
I'm working on a project that will have a bootloader on the first few kb of ROM, and the main program sitting in the rest of the memory, similar to [the example bootloader](https://github.com/libopencm3/libopencm3/wiki/USB-DFU).

Building the bootloader is straightforward, as its start ROM address matches what is at `ld/devices.data`. But building the main program, would require, based on the same device:

- Push `ROM_OFF` further, by the max size of the bootloader.
- Decrease `ROM` by the size of the bootloader.

This can be accomplished by adding a new device to `ld/devices.data` for the main program, eg:

```
stm32f411?e*_main_program stm32f4 ROM_OFF=0x08000800 ROM=510K RAM=128K
stm32f411?e* stm32f4 ROM=512K RAM=128K
stm32f4 END ROM_OFF=0x08000000 RAM_OFF=0x20000000 CPU=cortex-m4 FPU=hard-fpv4-sp-d16
```

This however, has 2 problems. `scripts/genlink.py` creates duplicate defines:

```
<command-line>:0:0: warning: "_ROM" redefined
<command-line>:0:0: note: this is the location of the previous definition
<command-line>:0:0: warning: "_ROM_OFF" redefined
<command-line>:0:0: note: this is the location of the previous definition
```

and upstream `ld/devices.data` would need to have project specific stuff (_main_program).

The solution I see to better support this is:

- Change `scripts/genlink.py` to make the last defined value the one to be used. This allows overrides and takes away the duplicaiton.
- Change `mk/genlink-rules.mk` and `mk/genlink-config.mk` to support custom per project `DEVICES_DATA`, that can work on top of upstream `ld/devices.data`.

I can send a PR with these changes, but I'd love to hear some opinions if there's a better approach to support bootloader + main program build.

fornellas:
- `scripts/genlink.py`:  allows for defines to be overridden, as opposed to just duplicating them.
- `mk/genlink-config.mk`: allows `DEVICES_DATA` to be externally defined.

I tested this at my new project https://github.com/fornellas/3dpkbd2, seems to work. There's `bootloader/` and `keyboard/`: each `Makefile` points to a different device, which translates to different ROM addresses, by using a `custom.devices.data`. It is worth confirming if this change makes sense, as despite the fact that `bootloader.bin` is <1k, the jump to main application only works, if it is offset by 16k from the start of the ROM (fails with anything less).

Closes #1117.

### DIFF
--- a/mk/genlink-config.mk
+++ b/mk/genlink-config.mk
@@ -22,7 +22,9 @@ $(warning no DEVICE specified for linker script generator)
 endif
 
 LDSCRIPT	= generated.$(DEVICE).ld
+ifeq ($(DEVICES_DATA),)
 DEVICES_DATA = $(OPENCM3_DIR)/ld/devices.data
+endif
 
 genlink_family		:=$(shell $(OPENCM3_DIR)/scripts/genlink.py $(DEVICES_DATA) $(DEVICE) FAMILY)
 genlink_subfamily	:=$(shell $(OPENCM3_DIR)/scripts/genlink.py $(DEVICES_DATA) $(DEVICE) SUBFAMILY)
--- a/scripts/genlink.py
+++ b/scripts/genlink.py
@@ -33,7 +33,7 @@ mode = sys.argv[3].upper()
 
 device = {
     'info': {},
-    'defs': [],
+    'defs': {},
     'family': [],
 }
 
@@ -73,7 +73,8 @@ with open(data_file_path, 'r') as data_file:
                 device['info'][k.lower()] = v
                 continue
 
-            device['defs'].append((k, v))
+            if k not in device['defs']:
+                device['defs'][k] = v
 
         # if parent is +, there's more data for this pattern
         if parent == '+':
@@ -102,8 +103,8 @@ if mode in ('CPPFLAGS', 'DEFS'):
 
 # defines
 if mode == 'DEFS':
-    if len(device['defs']) > 0:
-        defs = ' '.join('-D_%s=%s' % d for d in device['defs'])
+    if device['defs']:
+        defs = ' '.join('-D_{}={}'.format(k, v)for k, v in device['defs'].items())
         sys.stdout.write(' ' + defs)
 
 # device family


karlp:
Your 16k limit is simply the erase size of the first sector of the F411's memory map.

mk/genlink-config.mk
```diff
@@ -22,7 +22,9 @@ $(warning no DEVICE specified for linker script generator)
 endif
 
 LDSCRIPT	= generated.$(DEVICE).ld
+ifeq ($(DEVICES_DATA),)
 DEVICES_DATA = $(OPENCM3_DIR)/ld/devices.data
+endif
```
karlp: just DEVICES_DATA ?= is the same thing here...

karlp:
I'm really not at all convinced that enabling an override of the devices.data file entirely is useful.  I can see methods you might want to add extra line items, but repalcing the whole thing? I don't see the use case.

One thing to bear in mind is that the first match wins, so you just have to list your own things first, if you are trying to customize this.

Finally, I'm not sure the goal of the linker script generator was ever to support arbitrary layouts of user provided memory.  Originalyl I believe the scope was just to avoid the project shipping hundreds and hundreds of explicit linker files ourselves.  While we _could_ script our way into providing a way of customizing linker scripts, there's already a way of doing that..... using linker scripts :)  I'm not saying we _can't_ add this sort of PR, I'm just really not seeing the enhancement it provides vs just providing your own memory map files and using the common section definitions as I mentioned in my comment on https://github.com/libopencm3/libopencm3/issues/1117

I can put together a sample project of that style if it's not clear to you, but it avoids scripting and learning _another_ custom language and format (customizing the devices.data files)

fornellas:
All fair points. Not sure I missed something, but when I tried creating my project based on the template
https://github.com/libopencm3/libopencm3-template/blob/master/my-project/Makefile
I ran into the situation of not being able to both override LDSCRIPT **and** reuse the goodies (eg compiler flags) from
https://github.com/libopencm3/libopencm3/blob/master/mk/genlink-config.mk.

I also noticed the Makefile for the exapmles
https://github.com/libopencm3/libopencm3-examples/blob/master/examples/rules.mk
and the template
https://github.com/libopencm3/libopencm3-template/blob/master/rules.mk
are not the same, this is also potentially something that could live at libopencm3 repo, so all projects can share.

fornellas:
> I'm really not at all convinced that enabling an override of the devices.data file entirely is useful.

My intent was to [build on top](https://github.com/fornellas/3dpkbd2/blob/master/common.mk#L10) of the existing one, not override:
```
$(shell cat ${PROJECT_DEVICES_DATA} ${OPENCM3_DEVICES_DATA} > ${DEVICES_DATA})
```
So, the first match (`PROJECT_DEVICES_DATA`) wins.

I posted a comment on #1117 detailing what we get extra, TLDR, all compiler frags from `DEVICE` being set, as opposed to having to manually set them if we have a custom `LDSCRIPT`.

karlp:
I still feel you've missed my point.  You're now requiring users to learn the syntax of the devices.data file, so that they can script memory sizes and locations.  This is _already available_ via ld script language itself.

As for the compiler flags, that's a separate issue altogether.  aiui, they're even split into different include files for exactly that reason.

fornellas:
Let's drop this idea then, I'll sketch something for the compiler flags + custom LDSCRIPT later, thanks.

karlp:
Let's step back a bit and try and work out what problems we're really trying to solve.  Are you on irc at all? we can chat about it more back and forth there, it's a bit faster than github messages.

PR #1128, opened 2019-11-09
Add more USB endpoint defines

fornellas:
The [USB Endpoint Descriptor](https://www.beyondlogic.org/usbnutshell/usb5.shtml#EndpointDescriptors) `bEndpointAddress` attribute has things encoded in its bits:

> Bits 0..3b Endpoint Number.
> Bits 4..6b Reserved. Set to Zero
> Bits 7 Direction 0 = Out, 1 = In (Ignored for Control Endpoints)

This PR add a macro to allow the definition of the endpoint address using these components. Eg:

```c
#define HID_ENDPOINT_NUMBER 1
#define HID_ENDPOINT_IN_ADDR USB_ENDPOINT_ADDR_GEN(USB_ENDPOINT_DIR_IN, HID_ENDPOINT_NUMBER)
```

### DIFF
--- a/include/libopencm3/usb/usbstd.h
+++ b/include/libopencm3/usb/usbstd.h
@@ -226,6 +226,9 @@ struct usb_endpoint_descriptor {
 /* USB bEndpointAddress helper macros */
 #define USB_ENDPOINT_ADDR_OUT(x) (x)
 #define USB_ENDPOINT_ADDR_IN(x) (0x80 | (x))
+#define USB_ENDPOINT_DIR_OUT (0<<7)
+#define USB_ENDPOINT_DIR_IN (1<<7)
+#define USB_ENDPOINT_ADDR_GEN(direction, number) (direction | (number && 0xF))
 
 /* USB Endpoint Descriptor bmAttributes bit definitions - Table 9-13 */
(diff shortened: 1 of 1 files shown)


fornellas:
Ping. Is this good to merge or is there a feeling that it requires changes? Thanks.

karlp:
I hadn't gotten around to looking at it yet honestly, usb is always a headache.  But, what's wrong with just
```
#define HID_ENDPOINT_IN_ADDR USB USB_ENDPOINT_ADDR_IN(HID_ENDPOINT_NUMBER)
```
This uses the existing defines, (the ones _immediately_ before the ones you added, I'm curious how you overlooked them) and I'm really not sure what the extra value of your new ones is?

fornellas:
I don't think there's anything wrong with your approach. It is just for someone who's not super familiar with all USB stuff (I'm learning as I go here), I felt things could be a bit more explicit with this PR. Well, your call which way to go I suppose.

karlp:
Well, there's no documentation for your method either, so given that this is just "same, but different" I'm inclined to close it.  I'd rather have one defined way rather than many similar but different ways to do things honestly.

fornellas:
Fair :-D

PR #1134, opened 2019-11-21
lib: add stm32f103 low density devices

FFY00:
Add linker script for stm32f103x4 and stm32f103x6

Signed-off-by: Filipe Laíns <lains@archlinux.org>


karlp:
As discussed on irc, we're no longer attempting to provide discrete files for every available combination of devices.  The linker script generator is the recommended path forwards.

PR #1135, opened 2019-11-21
stm32l4: Fix typo impacting uart4 and uart5

larchuto:
UART4 and UART5 are mistakenly defined to USART4 and USART5 where it should be UART4 and UART5.


karlp:
merged thanks!

PR #1136, opened 2019-11-25
stm32g0: add syscfg header.

grevaillot:
add stm32g0 syscfg header.

### DIFF
--- a/include/libopencm3/stm32/g0/syscfg.h
+++ b/include/libopencm3/stm32/g0/syscfg.h
@@ -1,0 +1,325 @@
+/** @defgroup syscfg_defines SYSCFG Defines
+ *
+ * @ingroup STM32G0xx_defines
+ *
+ * @brief <b>Defined Constants and Types for the STM32G0xx System Configuration controller </b>
+ *
+ * @version 1.0.0
+ *
+ * @author @htmlonly &copy; @endhtmlonly 2019
+ * Guillaume Revaillot <g.revaillot@gmail.com>
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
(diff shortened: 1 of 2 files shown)


include/libopencm3/stm32/g0/syscfg.h
```diff
@@ -0,0 +1,325 @@
...
+/**@}*/
+
+/** SYSCFG_CFGR1_IR_POL IR output polarity selection */
+#define SYSCFG_CFGR1_IR_POL		(1 << 5)
+
+/** SYSCFG_CFGR1_PA12_RMP PA12 pin remapping */
+#define SYSCFG_CFGR1_PA12_RMP		(1 << 4)
+
+/** SYSCFG_CFGR1_PA11_RMP PA11 pin remapping */
+#define SYSCFG_CFGR1_PA11_RMP		(1 << 3)
```
fivdi: Very nice. These will come in handy for pin remapping on an 8 pin stm32g031j6

grevaillot: Well, I need to recheck, but i think that on 8 pins chips, pins are actually bonded to multiple gpio.

grevaillot:
Thanks.

include/libopencm3/stm32/g0/syscfg.h
fivdi: > but i think that on 8 pins chips, pins are actually bonded to multiple gpio.

Yes, this is correct. The remapping is an additional twist to the story.

I tested this PR yesterday and it worked. The default mapping for an stm32g031j6 is for PA11 and PA12 to be mapped to physical pins 5 and 6 respectively. By setting SYSCFG_CFGR1_PA11_RMP and SYSCFG_CFGR1_PA12_RMP PA9 and PA10 can be mapped to physical pins 5 and 6 instead of PA11 and PA12.

PR #1137, opened 2019-11-26
Makefile: Properly declare dependencies for generated files

DanielG:
Currently the generated file dependencies are not tracked by make, instead
the files are always regenerated. Furthermore when only building a single
target by running, say:

    $ make lib/stm32/f0

all headers are generated instead of only the ones necessary for this
target.

This patch adds dependency tracking for generated files to solve both these
problems.


karlp:
yes. that's true.  I'm very unconvinced that the amount of makefile you have to add for this is worth it though.

DanielG:
I'm happy enough to do some convincing :)

So the entire reason for this is that I finally want to get rid of the separate compilation step for libopencm3 itself when it's embedded into other projects. Every time I come back to doing something with lcm3 I get annoyed by having to hack something together to make that work.

In this iteration of my "make libopencm3 compile automatically" I've come up with adding this to the libopencm3-template Makefile:

```
LIBNAME = opencm3_stm32f0
[...]
$(OPENCM3_DIR)/lib/lib$(LIBNAME).a:
	make -C libopencm3 lib/stm32/f0
``` 

That works pretty well, except for two problems:

1) All GENHDRs runs on every `make` call in the parent project and
2) I have to set `LIBNAME` manually because `genlink-config.mk` for some ungodly reason tries to check if the library already exists and refuses to set LIBNAME if not.

I spent a couple of hours today trying to fix 1). My first try was to turn the `$(LIB_DIRS):` rule into a pattern rule like `lib/%.build:` so I can directly declare dependencies on the generated files like so:

```
lib/%.build: include/libopencm3/%/nvic.h include/libopencm3/%/irqhandlers.h lib/%/vector_nvic.c
```

Turns out that doesn't work because some targets *cough* lm4f *cough* don't have a irq.json to generate these files and make gets upset.

Ok, so let's try some conditionals:

```
lib/%.build: $(if $(wildcard include/libopencm3/%/irq.json), include/libopencm3/%/nvic.h include/libopencm3/%/irqhandlers.h lib/%/vector_nvic.c,)
```

That works fine in principle but there doesn't appear to be any way to pull these `%` patterns into a variable so then you end up with these paths splattered a couple of times across the Makefile which sucks.

If you would prefer that over foreach and eval I'm happy to go back to the original version which is certainly easier to understand but has even more duplication than the current version.

DanielG:
Ah, I'm an idiot. Doing most of the work in the target makefiles makes this a lot easier.

I managed to get rid of all the duplication and evil .. eeer `eval`. This should be much easier to swallow :)

Makefile
```diff
@@ -22,10 +22,6 @@ PREFIX		?= arm-none-eabi-
 STYLECHECK      := scripts/checkpatch.pl
 STYLECHECKFLAGS := --no-tree -f --terse --mailback
 
-space:=
-space+=
-SRCLIBDIR:= $(subst $(space),\$(space),$(realpath lib))
```
karlp: where does this get set now?

karlp:
ok, so, you've still got a pile of makefile hacking that I can't even follow.  Are we sure it wouldn't be vastly easier to just fix the genlink-config.mk?  I don't really see why it's refusing to set vars hwne the file isn't found, and the commit that adds it c025dc032  isn't enlightening.  You could then get 99% of what you need (building the lib automatically if it didn't exist) without any worries at all, even if there is some non-ideal rebuilds if you actually needed to rebuild.

Makefile
DanielG: This has been set in the target Makefiles for a while already, so for example lib/stm32/f0/Makefile.

DanielG:
If you point me to the Makefile bits that bother you most I'm happy to add some comments. I hate make as much as the next guy but after spending some time on it it all seems obvious in retrospect so I just need some input on what needs to be explained.

karlp:
So to back up a little, I want to try and address your _underlying_ concerns

> So the entire reason for this is that I finally want to get rid of the separate compilation step for libopencm3 itself when it's embedded into other projects. Every time I come back to doing something with lcm3 I get annoyed by having to hack something together to make that work.

and

> I have to set LIBNAME manually because genlink-config.mk for some ungodly reason tries to check if the library already exists and refuses to set LIBNAME if not.

Fiddling with the makefiles to avoid generating headers when you rebuild the library seems like a side issue, (real, I agree it's not perfect)  particularly for all _uses_ of the library, you only make the library once.  Lets fix _this_ stuff that affects people

DanielG:
There you are, please see the commit message for more commentary.

I've tested the new LIBNAME mechanism agains each and every entry in devices.data so that should all be good.

There are a couple of odd cases that currently don't have any corresponding libraries so with the old mechanism the don't set `LIBNAME` but now they just use the family suffix like everything else:
- stm32ts
- stm32w
- efm32zg
- rm46l852

And then there is lpc43xx_m0 which just used to be plain wrong, it had `libopencm3_lpc43xx.a` previously when obviously it should have been `libopencm3_lpc43xx_m0.a`. I went ahead and fixed that.

DanielG:
Changes:
- Rebased on master

72274d3caf09a5b4a6646e65bf376c91b8fa2ef1 seems to have incoorporated most but not all of my `Makefile: Use relative paths to refer to dirs above target` commit but git was able to resolve it cleanly (neat I didn't know it could do that).

@karlp can I get another review on this?

lib/Makefile.include
```diff
@@ -23,6 +23,8 @@ ifneq ($(V),1)
 Q := @
 endif
 
+TOPDIR = $(SRCLIBDIR)/..
```
karlp: I don't understand what this commit is still for?  It just adds another variable, but nothing that uses it?

lib/Makefile.include
```diff
@@ -23,7 +23,7 @@ ifneq ($(V),1)
 Q := @
 endif
 
-TOPDIR = $(SRCLIBDIR)/..
+TOPDIR := $(SRCLIBDIR)/..
```
karlp: this is changing the line you added in the previous commit. One of them is wrong :)

lib/Makefile.include
```diff
@@ -34,6 +34,20 @@ STANDARD_FLAGS ?= -std=c99
 
 all: $(SRCLIBDIR)/$(LIBNAME).a
 
+GENFILES = \
+	$(TOPDIR)/include/libopencm3/$(TARGET)/nvic.h \
+	$(TOPDIR)/include/libopencmsis/$(TARGET)/irqhandlers.h \
+	$(TOPDIR)/lib/$(TARGET)/vector_nvic.c
```
karlp: now we can't have any other generated files?  this doesn't seem scalable or easy for anyone else to ever work on.  (And if you _are_ making lists, use += instead of \ line extensions so that adding entries doesn't cause unrelated changes.

lib/Makefile.include
```diff
@@ -34,6 +34,20 @@ STANDARD_FLAGS ?= -std=c99
...
+
+IRQ_DEFN = $(TOPDIR)/include/libopencm3/$(TARGET)/irq.json
+
+$(GENFILES) : $(IRQ_DEFN)
+	@printf "  GENHDR   $(TARGET) %s\n" '$<';
+	$(Q)cd $(TOPDIR) && ./scripts/irq2nvic_h ./$(subst $(TOPDIR)/,,$<)
+
+# Real dependencies come in via *.d files
+$(OBJS): | $(if $(wildcard $(IRQ_DEFN)), $(GENFILES),)
```
karlp: pipe deps? is this a game to try and use the most obscure corners of makefiles possible?

lib/Makefile.include
DanielG: Yeah, you're right. Looks like it can just be removed.

lib/Makefile.include
DanielG: ack, `+=` is obviosly better here.

But I don't understand your point about not being able to "habe other generated files"? If new generated (header) files come around we just have to add them to the list.

lib/Makefile.include
DanielG: In principle I can use a regular prerequisite relation here, but the problem is that would be an overapproximation and cause all object files to be rebuilt if any of the generated headers change regardless of them actually using the header or not -- hardly satisfying. So instead I use order-only and let the `*.d`'s do the real work. I'll add a comment about that.

karlp: would it be easier here to change the irq generator to have an argument for the type of file to make?  it would run more times, but would make it easier to have "simple" "boring" make rules for each file and clean up things with $(RM) instead of having to run the script with "--cleanup" ?

lib/Makefile.include
karlp: I just meant that now the list of files is here, _and_ inside the generator, and not a lot of visibility that you need to edit both.

DanielG:
Seems the part of this that wasn't pulled out to #1301 got superseded by d60d7802fd20821a766675545b6ef7a2207207a3.

PR #1138, opened 2019-11-26
tests: gadget0: Add support for nucleo-f070rb

DanielG:
This patch adds gadget0 support for the nucleo-f070rb development board.

Since this board has USART2 connected to the ST-Link debugger I've opted to
route both trace and printf output there for convinience.

### DIFF
--- a/tests/gadget-zero/Makefile.nucleo-f070rb
+++ b/tests/gadget-zero/Makefile.nucleo-f070rb
@@ -1,0 +1,41 @@
+##
+## This file is part of the libopencm3 project.
+##
+## This library is free software: you can redistribute it and/or modify
+## it under the terms of the GNU Lesser General Public License as published by
+## the Free Software Foundation, either version 3 of the License, or
+## (at your option) any later version.
+##
+## This library is distributed in the hope that it will be useful,
+## but WITHOUT ANY WARRANTY; without even the implied warranty of
+## MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+## GNU Lesser General Public License for more details.
+##
+## You should have received a copy of the GNU Lesser General Public License
+## along with this library.  If not, see <http://www.gnu.org/licenses/>.
+##
+
+BOARD = nucleo-f070rb
+PROJECT = usb-gadget0-$(BOARD)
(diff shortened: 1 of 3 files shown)


karlp:
What does this provide that's not covered by the f072?  The aim here is coverage of all the cores, not every single usb capable board ever.  The f070 nucleo (like all nucle64s) doesn't even have a user USB port, so it requires mods or addons to even use this?

DanielG:
The f072 has a seperate reference manual, so I figured it's not absolutely guaranteed that f70 and f72 USB cores are identical.

Yeah indeed it needs some funky usb cable soldering but this is the only off-the-shelf devboard with an f70 I could find.

karlp:
Well, you've verified that it does work the same, so I'd really rather not add something trying to track just this variant. The F072 and L053 test the usbfs_v2 core on both an m0 and an m0+ in two families.  I know the f070 is "value line" and requires a crystal, so... _maybe_ but I can'
t work with on just a nucleo board with custom wiring, that's never going to be useful for anyone else, and I can't ever set that up automatically.

DanielG:
I don't really see the harm of having this upstream even if you're not going to use it but suite yourself. For me the main reason to upstream this is to document the fact that usb peripheral on the f070 is in fact supported. I had to do a fair amount of digging to confirm that fact when we were searching for an as-cheap-as-possible uC for our product.

I suppose instead of the test I could just add something to libopencm3-examples, unless there is a more canonical place to document this other than the test suite?

karlp:
Well, I'm impressed that you checked the gadget0 for status of usb.  That's certainly a good way of doing it.  The problem is that we have to maintain it, and to maintain it has to be feasible to be run.  Perhaps strip it down some more, and make it an f0-valueline-generic or something?  (much like L1, which also has no readily available board with user usb) that makes it clearer that it's not something that can be easily run, but meets your desires of having it "in a test suite" ?  

If the only goal is documenting that it works, perhaps just... expanding on the documentation?  The http://libopencm3.org/docs/latest/stm32f0/html/group__usb__driver__defines.html section could easily be expanded with more information covering notes and targets known to be supported.  Perhaps try adding some doxygen around lib/stm32/st_usbfs_v2.c and trying to make it all show up nicely in the generated output?

PR #1140, opened 2019-11-28
Fully const config descriptors

vampirefrog:
I hope I'm doing this right, it wouldn't let me choose a new branch, so I chose master.

This implements the const USB descriptor support (fully in ROM).

gadget0 is also updated to use those.

Note that two of the gadget0 tests fail, but they failed before my fork so I believe it works fine.

Some careful inspection needs to be taken when it's going through the descriptors, I believe it should be fine, but there is a chance it will overrun.

### DIFF
--- a/lib/usb/usb_standard.c
+++ b/lib/usb/usb_standard.c
@@ -65,80 +65,6 @@ void usbd_register_set_altsetting_callback(usbd_device *usbd_dev,
 	usbd_dev->user_callback_set_altsetting = callback;
 }
 
-static uint16_t build_config_descriptor(usbd_device *usbd_dev,
-				   uint8_t index, uint8_t *buf, uint16_t len)
-{
-	uint8_t *tmpbuf = buf;
-	const struct usb_config_descriptor *cfg = &usbd_dev->config[index];
-	uint16_t count, total = 0, totallen = 0;
-	uint16_t i, j, k;
-
-	memcpy(buf, cfg, count = MIN(len, cfg->bLength));
-	buf += count;
-	len -= count;
-	total += count;
-	totallen += cfg->bLength;
-
-	/* For each interface... */
-	for (i = 0; i < cfg->bNumInterfaces; i++) {
-		/* Interface Association Descriptor, if any */
-		if (cfg->interface[i].iface_assoc) {
-			const struct usb_iface_assoc_descriptor *assoc =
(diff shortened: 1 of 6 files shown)


lib/usb/usb_standard.c
```diff
@@ -277,13 +203,22 @@ usb_standard_set_configuration(usbd_device *usbd_dev,
...
 
 	if (usbd_dev->current_config > 0) {
-		cfg = &usbd_dev->config[usbd_dev->current_config - 1];
+		cfg = usbd_dev->config[usbd_dev->current_config - 1];
 
 		/* reset all alternate settings configuration */
-		for (i = 0; i < cfg->bNumInterfaces; i++) {
-			if (cfg->interface[i].cur_altsetting) {
-				*cfg->interface[i].cur_altsetting = 0;
+		int totalLen = cfg->wTotalLength;
```
yumkam: nit: `uint16_t totalLen = `... and drop `(int)` cast below? (it is a bit more brittle though)

lib/usb/usb_standard.c
```diff
@@ -277,13 +203,22 @@ usb_standard_set_configuration(usbd_device *usbd_dev,
...
+		while(totalLen > 0) {
+			if(totalLen >= (int)sizeof(struct usb_interface_descriptor)) {
+				if(ptr[1] == USB_DT_INTERFACE) {
+					iface = (struct usb_interface_descriptor *)ptr;
+					if(!usbd_dev->current_iface[iface->bInterfaceNumber])
+						usbd_dev->current_iface[iface->bInterfaceNumber] = iface;
+				}
 			}
+			ptr += ptr[0];
+			totalLen -= ptr[0];
```
yumkam: should not two above lines be swapped?

vampirefrog: You may be right, but I haven't looked at this patch in a while, and I haven't seen any signs from the maintainers either, so if you want you can make a test case for the problem at line 220, that would clear that up and make it easy for a maintainer to merge the changes as well. But I'm glad someone's inspecting it.

vampirefrog: you're right, I've fixed the order

lib/usb/usb_standard.c
vampirefrog: nah

PR #1141, opened 2019-12-02
Initial Completed Implementation of PAC55xx GPIO Driver.

vielster:
* Conforms mostly to the STM32 GPIO API where possible.
* Supports pin configuration (direction, pull-up/down, etc.) as well
  as pinmux configuration.
* Supports set/clear/get operations to the GPIO port/pins.
* Attempt at maintaining clear/clean doxy groups.

### DIFF
--- a/include/libopencm3/pac55xx/ccs.h
+++ b/include/libopencm3/pac55xx/ccs.h
@@ -1,0 +1,141 @@
+/** @file
+ *
+ *  Clock Control and System Defines for the Qorvo PAC55xx series of microcontrollers.
+ *
+ *  @defgroup system_defines Clock Config and System Defines
+ *  @ingroup PAC55xx_defines
+ *  @author Brian Viele <vielster@allocor.tech>
+ *  LGPL License Terms @ref lgpl_license
+ *  @date 1 Dec 2019
+ *
+ *  Definitions in this file come from the PAC55XX Family User Guide Rev 1.21
+ *  by Active-Semi dated August 26, 2019.
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef INCLUDE_LIBOPENCM3_PAC55XX_CCS_H_
+#define INCLUDE_LIBOPENCM3_PAC55XX_CCS_H_
+
+#include <libopencm3/cm3/common.h>
+
+/** Clock Control Registers
+ * @defgroup clock_config_regs Clock Config Registers.
+ * @ingroup system_defines
+ * @{*/
+#define CCSCTL MMIO32(SCC_BASE)
+#define CCSPLLCTL MMIO32(SCC_BASE + 0x04)
+#define CCSROSCTRIM MMIO32(SCC_BASE + 0x08)
+/**@}*/
+
+/** Port Pin Config Addresses
+ * @defgroup port_pin_addresses Port Pinmux Register Base.
+ * @ingroup system_defines
+ * @{*/
+#define CCS_PORTA (SCC_BASE + 0x0C)
+#define CCS_PORTB (SCC_BASE + 0x10)
+#define CCS_PORTC (SCC_BASE + 0x14)
+#define CCS_PORTD (SCC_BASE + 0x18)
+#define CCS_PORTE (SCC_BASE + 0x1C)
+#define CCS_PORTF (SCC_BASE + 0x20)
+#define CCS_PORTG (SCC_BASE + 0x24)
+/**@}*/
+
+/** Port Pin Mux Select Registers
+ * @defgroup pmux_sel_regs PMUXSEL register mapping.
+ * @ingroup system_defines
+ * @{*/
+#define CCS_MUXSELR(base) MMIO32(base)
+#define CCS_PAMUXSELR CCS_MUXSELR(CCS_PORTA)
+#define CCS_PBMUXSELR CCS_MUXSELR(CCS_PORTB)
+#define CCS_PCMUXSELR CCS_MUXSELR(CCS_PORTC)
+#define CCS_PDMUXSELR CCS_MUXSELR(CCS_PORTD)
+#define CCS_PEMUXSELR CCS_MUXSELR(CCS_PORTE)
+#define CCS_PFMUXSELR CCS_MUXSELR(CCS_PORTF)
+#define CCS_PGMUXSELR CCS_MUXSELR(CCS_PORTG)
+#define CCS_MUXSELR_MASK 0x7
+/* Enum type for port function setting for type specificity. */
+typedef enum {
+	CCS_MUXSEL_GPIO = 0,
+	CCS_MUXSEL_AF1 = 1,
+	CCS_MUXSEL_AF2 = 2,
+	CCS_MUXSEL_AF3 = 3,
+	CCS_MUXSEL_AF4 = 4,
+	CCS_MUXSEL_AF5 = 5,
+	CCS_MUXSEL_AF6 = 6,
+	CCS_MUXSEL_AF7 = 7,
+} ccs_muxsel_func_t;
+/**@}*/
+
+/** Port Pull-Up/Down Enable Registers.
+ * @defgroup pden_regs PUEN PDEN register mapping.
+ * @ingroup system_defines
+ * @{*/
+#define CCS_PUENR(base) MMIO32(base + 0x1C)
+#define CCS_PAPUENR CCS_PUENR(CCS_PORTA)
+#define CCS_PBPUENR CCS_PUENR(CCS_PORTB)
+#define CCS_PCPUENR CCS_PUENR(CCS_PORTC)
+#define CCS_PDPUENR CCS_PUENR(CCS_PORTD)
+#define CCS_PEPUENR CCS_PUENR(CCS_PORTE)
+#define CCS_PFPUENR CCS_PUENR(CCS_PORTF)
+#define CCS_PGPUENR CCS_PUENR(CCS_PORTG)
+#define CCS_PDENR(base) MMIO32(base + 0x38)
+#define CCS_PAPDENR CCS_PDENR(CCS_PORTA)
+#define CCS_PBPDENR CCS_PDENR(CCS_PORTB)
+#define CCS_PCPDENR CCS_PDENR(CCS_PORTC)
+#define CCS_PDPDENR CCS_PDENR(CCS_PORTD)
+#define CCS_PEPDENR CCS_PDENR(CCS_PORTE)
+#define CCS_PFPDENR CCS_PDENR(CCS_PORTF)
+#define CCS_PGPDENR CCS_PDENR(CCS_PORTG)
+/* Pull Up/Down enum for type specificity. */
+typedef enum {
+	CCS_IO_PULL_NONE = 0,
+	CCS_IO_PULL_UP = 1,
+	CCS_IO_PULL_DOWN = 2
+} ccs_pull_updown_t;
+/**@}*/
+
+/** Port Drive Strength Enable Registers.
+ * @defgroup dsr_regs DSR register mapping.
+ * @ingroup system_defines
+ * @{*/
+#define CCS_DSR(base) MMIO32(base + 0x54)
+#define CCS_PADSR CCS_DSR(CCS_PORTA)
+#define CCS_PBDSR CCS_DSR(CCS_PORTB)
+#define CCS_PCDSR CCS_DSR(CCS_PORTC)
+#define CCS_PDDSR CCS_DSR(CCS_PORTD)
+#define CCS_PEDSR CCS_DSR(CCS_PORTE)
+#define CCS_PFDSR CCS_DSR(CCS_PORTF)
+#define CCS_PGDSR CCS_DSR(CCS_PORTG)
+#define CCS_DSR_MASK 0x7
+#define CCS_DSR_MASK_PIN(pin) (CCS_DSR_MASK << ((pin)) * 4)
+#define CCS_DSR_DS_VAL(pin, ds) (((ds)&CCS_DSR_MASK) << ((pin)*4))
+#define CCS_DSR_SCHMIDT_PIN(pin) (BIT0 << (((pin)*4) + 3))
+
+/* Drive strength enumeration for type specificity. */
+typedef enum {
+	CCS_DSR_DS_6MA = 0x00,
+	CCS_DSR_DS_8MA = 0x01,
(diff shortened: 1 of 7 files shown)


vielster:
These basic functions will be tested on hardware tomorrow (Dec 2nd)

karlp:
given that no-one else is likely to have hardware for these, I'd kinda rather you didn't submit stuff that was only based on the datasheets :)  I'd _like_ it if stuff in the tree had at least plausibly been tested on real hardware by at least the submitter!  I believe you _will_ that's fine, but... what's the rush?

include/libopencm3/pac55xx/ccs.h
```diff
@@ -0,0 +1,148 @@
...
+ *
+ *  @defgroup system_defines Clock Config and System Defines.
+ *  @ingroup PAC55xx
+ *  @author Brian Viele <vielster@allocor.tech>
+ *  LGPL License Terms @ref lgpl_license
+ *  @date 1 Dec 2019
+ *
+ *  Definitions in this file come from the PAC55xx Family User Manual which can
+ *  be found here.
+ *  https://active-semi.com/wp-content/uploads/PAC55XX_Family_User_Guide.pdf
```
karlp: fwiw, a document id and revision are often better than a url, as vendors have wonderful habits of breaking them, especially we _already_ know that active-semi doesn'te xist anymore.

include/libopencm3/pac55xx/gpio.h
```diff
@@ -0,0 +1,367 @@
...
+/* GPIO Mode Register Definitions */
+/** @defgroup gpio_mode_regs GPIO MODE register mapping.
+@ingroup PAC55xx_gpio
+@{*/
+/* Enum definitions for at least minimal type safety. */
+typedef enum {
+  ANALOG_INPUT                      = 0,
+  PUSH_PULL_OUTPUT                  = 1,
+  OPEN_DRAIN_OUTPUT                 = 2,
+  HIGH_IMPEDENCE_INPUT              = 3,
```
karlp: these need a consistent prefix to avoid making a mess of the namespace

include/libopencm3/pac55xx/gpio.h
```diff
@@ -0,0 +1,367 @@
...
+  volatile uint32_t inten;
+  volatile const uint32_t intflag;
+  volatile uint32_t intclear;
+  volatile uint32_t inttype;
+  volatile uint32_t intcfg;
+  volatile uint32_t intintedgeboth;
+  volatile uint32_t clksync;
+  volatile uint32_t doset;
+  volatile uint32_t doclear;
+} pac55xx_gpio_regs_t;
```
karlp: providing both styles?  are you going to be maintaining two apis?  so far we've _only_ done define style.  struct style might often be considered nicer, but changing things is.... not good.

lib/pac55xx/gpio.c
```diff
@@ -0,0 +1,192 @@
...
+  }
+}
+
+/* ========================================================================= */
+void gpio_mode_setup(uint32_t gpioport, gpio_mode_t mode,
+    pull_updown_t pull_up_down, uint16_t gpios) {
+  uint32_t reg =  GPIO_MODER(gpioport);  /* initialize soft value to write. */
+  uint32_t port = get_ccs_port_base(gpioport);
+
+  /* Soft assert. We can't use these bad values. */
```
karlp: I'd be inclined to either make it hard, or drop it personally

lib/pac55xx/gpio.c
```diff
@@ -0,0 +1,192 @@
...
+  }
+  GPIO_OUTMASKR(gpioport) = reg;
+}
+
+/* ========================================================================= */
+void gpio_set(uint32_t gpioport, uint16_t gpios) {
+  GPIO_DOSETR(gpioport) = gpios;
+}
+
+/* ========================================================================= */
```
karlp: drop all these separator stuff please.

include/libopencm3/pac55xx/ccs.h
```diff
@@ -0,0 +1,148 @@
...
+@ingroup system_defines
+@{*/
+#define DSR(base)                 MMIO32(base + 0x54)
+#define PADSR                     DSR(PORTA)
+#define PBDSR                     DSR(PORTB)
+#define PCDSR                     DSR(PORTC)
+#define PDDSR                     DSR(PORTD)
+#define PEDSR                     DSR(PORTE)
+#define PFDSR                     DSR(PORTF)
+#define PGDSR                     DSR(PORTG)
```
karlp: you need better namespace prefixes here.  it pollutes too much without any clear grouping.

karlp:
You've got spaces instead of tabs in ~everything as far as I can tell too.

vielster:
* The reason I put up for review prior to testing is to get non-functional change workflow in process while I also did functional changes (as I assumed there would likely be some of both). It actually all seems functional so far in testing...I have blinking LEDs!!!
* Will drop struct mapping. I thought it nice to provide the overlay for people who have that preference, but it likely will end up with mixed impls in the C code, which would be bad, so I will remove.
* Will drop separators 
* Will drop soft asserts
* Will add namespacing (`CCS_` prefixes...hope that's sufficient)

Spaces/Tabs may be a bit more difficult to maintain on my end since this is a submodule in a tree and all other code in our codebase uses spaces. If this is a hard requirement, please LMK and I'll see if I can do something with clang-format or otherwise.

include/libopencm3/pac55xx/memorymap.h
```diff
@@ -0,0 +1,77 @@
...
+#define PAC55XX_USARTC_BASE         (PAC55XX_PERIPH_BASE + 0x40000)
+#define PAC55XX_USARTD_BASE         (PAC55XX_PERIPH_BASE + 0x50000)
+#define PAC55XX_TIMERA_BASE         (PAC55XX_PERIPH_BASE + 0x60000)
+#define PAC55XX_TIMERB_BASE         (PAC55XX_PERIPH_BASE + 0x70000)
+#define PAC55XX_TIMERC_BASE         (PAC55XX_PERIPH_BASE + 0x80000)
+#define PAC55XX_TIMERD_BASE         (PAC55XX_PERIPH_BASE + 0x90000)
+#define PAC55XX_CAN_BASE            (PAC55XX_PERIPH_BASE + 0xA0000)
+#define PAC55XX_GPTIMERA_BASE       (PAC55XX_PERIPH_BASE + 0xB0000)
+#define PAC55XX_GPTIMERB_BASE       (PAC55XX_PERIPH_BASE + 0xC0000)
+#define PAC55XX_SYS_PERIPH_BASE     (PAC55XX_PERIPH_BASE + 0xD0000)
```
karlp: I couldn't put my finger on it earlier, but I see now :)  None of these need a PAC55XX_ prefix.  these _are_ the peripherals, all the peripheral files will start with the same prefix as well, and  you're never including two memorymaps at the same time, as the dispatch header prevents it.  It doesn't _hurt_ per se here, but you can skip them.

include/libopencm3/pac55xx/ccs.h
```diff
@@ -0,0 +1,146 @@
...
+#define CCS_PEPDENR                   CCS_PDENR(PORTE)
+#define CCS_PFPDENR                   CCS_PDENR(PORTF)
+#define CCS_PGPDENR                   CCS_PDENR(PORTG)
+
+/* Pull Up/Down enum for type specificity. */
+typedef enum {
+  CCS_IO_PULL_NONE                    = 0,
+  CCS_IO_PULL_UP                      = 1,
+  CCS_IO_PULL_DOWN                    = 2
+} pull_updown_t;
```
karlp: the typedef needs a prefix too please

include/libopencm3/pac55xx/ccs.h
```diff
@@ -0,0 +1,146 @@
...
+typedef enum {
+  CCS_DSR_DS_6MA                      = 0x00,
+  CCS_DSR_DS_8MA                      = 0x01,
+  CCS_DSR_DS_11MA                     = 0x02,
+  CCS_DSR_DS_14MA                     = 0x03,
+  CCS_DSR_DS_17MA                     = 0x04,
+  CCS_DSR_DS_20MA                     = 0x05,
+  CCS_DSR_DS_22MA                     = 0x06,
+  CCS_DSR_DS_25MA                     = 0x07,
+} drive_strength_t;
```
karlp: and here

karlp:
The whitespace on tabs is pretty hard I'm afraid.  You should be able to set your editors to only match on certain directories for things though?  Apart from the typedefs, the rest looks good.

include/libopencm3/pac55xx/memorymap.h
vielster: Done.

include/libopencm3/pac55xx/ccs.h
vielster: Done

vielster:
I added a .clang-format file to the top of the tree which allows running  `clang-format -i path/to/file` to re-format the file using a fixed set of rules. I figured I'd use style=Google + updates for tabs, and some other alignment, and you could modify the auto-formatting as you see fit. If you don't mind leaving this in the tree, it can be used by anyone (including me) to do an easy reformat prior to submission.

karlp:
re clang formatter, that's fine, but it needs to go in separately, not part of this stuff.

include/libopencm3/pac55xx/ccs.h
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef INCLUDE_LIBOPENCM3_PAC55XX_CCS_H_
+#define INCLUDE_LIBOPENCM3_PAC55XX_CCS_H_
+
+#include <libopencm3/cm3/common.h>
+
+/* clang-format off */
```
karlp: what? that can't be right

include/libopencm3/pac55xx/ccs.h
```diff
@@ -0,0 +1,154 @@
...
+	CCS_MUXSEL_AF2  = 2,
+	CCS_MUXSEL_AF3  = 3,
+	CCS_MUXSEL_AF4  = 4,
+	CCS_MUXSEL_AF5  = 5,
+	CCS_MUXSEL_AF6  = 6,
+	CCS_MUXSEL_AF7  = 7,
+} ccs_muxsel_func_t;
+/**@}*/
+
+/* clang-format off */
```
karlp: same with everywhere else please, we can't go scattering this all over the place

lib/pac55xx/gpio.c
```diff
@@ -0,0 +1,164 @@
...
+uint16_t gpio_get(uint32_t gpioport, uint16_t gpios) {
+	return GPIO_INR(gpioport) & gpios;
+}
+
+void gpio_set_af(uint32_t gpioport, ccs_muxsel_func_t muxsel, uint16_t gpios) {
+	uint32_t port = get_ccs_port_base(gpioport);
+
+	/* Update each of the pin configs. */
+	uint32_t reg = CCS_MUXSELR(port);
+	int ffs      = __builtin_ffs(gpios);
```
karlp: we've never done this elsewhere, I dont' see any reason to start, I'm guessing this is from a "suboptimal" choice of options when you clang-formatted the whole thing?

include/libopencm3/pac55xx/ccs.h
vielster: So, without this, it will collapse all macro spacing (meaning no macro alignment). I'll delete these now that everything is properly formatted.

lib/pac55xx/gpio.c
vielster: Adjacent assignments are aligned for readability. I can remove this option.

karlp:
You still have lots of mixed spaces/tabs in your #defines, and any clang formatting stuff still needs to be completely separate commit (indeed, separate pr)

include/libopencm3/pac55xx/ccs.h
```diff
@@ -0,0 +1,146 @@
...
+	CCS_DSR_DS_11MA = 0x02,
+	CCS_DSR_DS_14MA = 0x03,
+	CCS_DSR_DS_17MA = 0x04,
+	CCS_DSR_DS_20MA = 0x05,
+	CCS_DSR_DS_22MA = 0x06,
+	CCS_DSR_DS_25MA = 0x07,
+} ccs_drive_strength_t;
+
+#define CCS_DSR_DS(pin_, ds_) (ds_ << (pin << 4))
+#define CCS_DSR_SCHMIDT_EN(pin_) (1 << ((pin << 4) + 3))
```
karlp: you've changed these, and are you sure they work? you're missing _ from pin in both cases?

include/libopencm3/pac55xx/gpio.h
```diff
@@ -0,0 +1,334 @@
...
+#define GPIOF_MODER GPIO_MODER(GPIOF)
+#define GPIOG_MODER GPIO_MODER(GPIOG)
+#define GPIO_MODER_MASK 0x3
+#define GPIO_MODER_MASK_PIN(pin) (GPIO_MODER_MASK << ((pin)*2))
+#define GPIO_MODER_MODE(pin, mode) (((mode)&GPIO_MODER_MASK) << ((pin)*2))
+/**@}*/
+
+/* GPIO Output Mask Register Definitions. This register may be used to lock the output value of
+ * a pin as changes to masked pins will have no effect. */
+/** @defgroup gpio_outmask_regs GPIO OUTMASK register mapping.
```
karlp: You don't need to close and open comments here.  Just put a second * at the start.  That way you actually get the description in the output.  Have you actually been running "make doc" and looking at how this all comes out?  (and checking the warnings in the logs?

karlp:
Is there any particular reason you made a new style and layout for all the doxygen?  You've created peripherals at the top level, mixed with all the definitions.  The other families all have a "peripheral apis" in one place, and "all the defines" in another section.  Look at, for instance, http://libopencm3.org/docs/latest/stm32l1/html/group__gpio__file.html and http://libopencm3.org/docs/latest/stm32l1/html/group__gpio__defines.html

I can't say I'm real hot on the idea of having the docs for this family look different unless there's a good reason for it.

vielster:
I had not intended to have dissimilar docs, I've just been trying to glean from other source files, and have apparently done a bad job. I actually really like the peripheral_apis parent the way that you have it, I just didn't catch that in the sample file I looked at, so that's my bad. 

The only thing I _did_ feel strongly about was putting the doxy in the header rather than the source because when I'm trying to work with the source API, it is helpful for that to all be located in the header file as IMO, rather than having to dig through the source file. My personal opinion is that all someone should need to use an API is the header file.

include/libopencm3/pac55xx/ccs.h
vielster: Nice catch...I ended up with duplicates of these, and the ones above are what are used. I've removed these.

vielster:
I have made a bunch of mods to the groups to attempt to mirror what's in libopencm3. This makes a lot more sense now...

image

include/libopencm3/pac55xx/gpio.h
vielster: Okay...made changes in `ccs.h` and `gpio.h` for this. I like the continuous block better, so will do that in the future.

vielster:
Clang format stuff is out. I believe doxy is all cleaned up now and matches. I've also attempted to ensure tabs-only.

vielster:
The dangers of force push just bit me. I just pushed an old version from a machine that was not sync'd, and will undo this later once I have access to the machine with the latest-and-greatest. My apologies.

karlp:
Merged thanks!  I made some doxygen cleanups afterwards.  It can be worth checking in the doc/pac55xx/doxygen_blah.log file for warnings sometimes.

karlp:
merged in a3406f10

PR #1142, opened 2019-12-03
stm32g0: use spi v2

manor1911:
Seems like the STM32G0 family uses the new SPI version.


grevaillot:
Hmm, indeed. +1.

Something is fishy, g071 SVD file does not match the current RM - I need to dig a bit into the spi "v2" peripheral, i can not remember what were the specificities except the dff/crcl bits.

karlp:
spiv2 is basically whether it has the fifo or not.  I should have caught this when it came in honestly :)

karlp:
merged in 9af9a1db thanks

PR #1143, opened 2019-12-03
device.data: add stm32f105xb stm32f105xc

urx:
Adds support for ld file generation for stm32f105xb and stm32f105xc devices.


karlp:
These are both already covered, see line 97 and 101, what device strings were you trying that weren't working for you? and what errors were you getting?

urx:
> These are both already covered, see line 97 and 101, what device strings were you trying that weren't working for you? and what errors were you getting?

Oh, sorry for that. I just found error in my project which leads to empty generated ld file, so i check device.data and didn't notice that support is already there.

PR #1145, opened 2019-12-06
Add gd32f303 support

nanovna:
The gd32f3 peripherals layout is based on the stm32f1, so this patch simply copies the stm32f1 library definition to gd32f3 but sets different compiler flags (use cortex-m4 and enable fpu). Later patches may add things like expanded prescaler values (for e.g. USB support while running at 96MHz).


karlp:
So, this PR cannot merge in it's current form.  Please look at how gd32f1xx support was merged for some basic ideas.

We _cannot_ simply copy and paste entire trees of code.  That might work for your own prototyping but it is completely unmaintainable long term.

Further, I don't believe there is any interest in code to support early revision silicon.  At least, certainly no in a manner of copy/pasting entire trees of code.  Given the availability, I would advise users of such silicon to simply rebuild their own tree locally overriding FPU_FLAGS.

Given how much would need to change on this PR, I'm inclined to close it, but we can keep it open if you would like to push revised trees to this same PR.

nanovna:
Not really familiar with how the project is supposed to be structured, so I think it's easier if you re-do this patch from scratch instead. It's supposed to just simply declare a new device target that shares the same code as STM32F1, but if I simply reference the .c files in the F1 directory it won't recompile for this target because the output .o files already exist. I think maybe moving the .c files out of the F1 directory and to a common directory might work, but I don't know your naming conventions to do it properly.

PR #1147, opened 2019-12-10
avoid creating blank linker script if gcc fails

dlharmon:
When piping to a file, if arm-none-eabi-gcc is not present in the path,
a blank linker script is created with genlink. After sourcing a bash
script to add GCC to the path, the linker script doesn't get rebuilt
due to a fresh timestamp despite failing to generate.

It may be possible that old GCCs didn't support the -o option here?


karlp:
so..  don't make PRs from your own master branch :)  You end up with unintended commits in the PR.

dlharmon:
Sorry about that. Redone as #1148

PR #1151, opened 2019-12-15
RCC improvements for STM32

joe-skb7:
This patch series introduces next changes:

1. When working on my pet project, I had to copy-paste the rcc_clock_setup_in_hse_8mhz_out_72mhz() function and change ADC divisor to DIV6, in order to achieve max ADC clock frequency. As original rcc_clock_setup_in_hse_8mhz_out_72mhz() function can change in upstream, I'll need to track those changes and fix my internal function correspondingly, so it can be considered a bad practice. Better idea is to have such a function in upstream libopencm3, so everyone can call it directly.

2. Another thing I noticed along the way is a type (ABP -> APB) in mentioned function. So I fixed it tree-wide as well.


karlp:
This actually highlights, again, just how bad teh f1 rcc apis are.  I'm not going to merge more of those copy pasta fixed configurations, We need to add a clock struct style like the rest of the library uses, and mark all the existing ones as deprecated.  I've cherry picked the typo fix though, into f7a952c4 thanks

PR #1152, opened 2019-12-18
stm32h7: PWR and RCC to support PLL configuration.

vielster:
PLL configuration on the H7 is pretty involved, and takes a number of
configurations to make it work. In order to make peripheral drivers a bit
easier to implement, working with a soft clock tree in the rcc module which
stores the clock settings for each clock as they are setup such that users
can request the clock value from the RCC module for configuration.

Clock configuration is still missing values for setting up all of the kernel
clocks for the peripherals, but this is in work.

Have tested to 400MHz without issue. Peripherals that are explicitly supported
are working and the clock tree values appear to follow correctly.

Updated common USART driver to take these clock configs when using the H7
since the APB clocks cannot be directly used.

Added LDO settings to allow setting the scaling to support high frequencies.

### DIFF
--- a/include/libopencm3/stm32/h7/pwr.h
+++ b/include/libopencm3/stm32/h7/pwr.h
@@ -62,24 +62,70 @@ LGPL License Terms @ref lgpl_license
 #define PWR_CR1_SVOS_SCALE_3      (0x3)
 #define PWR_CR1_SVOS_SCALE_4      (0x2)
 #define PWR_CR1_SVOS_SCALE_5      (0x1)
+
 #define PWR_CR1_SVOS_MASK         (0x3)
 
 /** DBP[8]: Disable backup domain write protection. */
 #define PWR_CR1_DBP               (1 << 8)
 
-/** PVDO: PVD output */
-#define PWR_CSR1_PVDO             (1 << 4)
+/** CSR1 Register Bits */
+#define PWR_CSR1_AVDO             BIT16
+#define PWR_CSR1_ACTVOS_SHIFT     14
+#define PWR_CSR1_ACTVOSRDY        BIT13
+#define PWR_CSR1_PVDO             BIT4
+
+/** CR3 Register Bits */
+#define PWR_CR3_USB33RDY          BIT26
+#define PWR_CR3_USBREGEN          BIT25
+#define PWR_CR3_USB33DEN          BIT24
+#define PWR_CR3_VBRS              BIT9
+#define PWR_CR3_VBE               BIT8
+#define PWR_CR3_SCUEN             BIT2
+#define PWR_CR3_LDOEN             BIT1
+#define PWR_CR3_BYPASS            BIT0
+
+/** D3CR Register Bits */
+#define PWR_D3CR_VOS_SHIFT        14
+#define PWR_D3CR_VOSRDY           BIT13
+
+#define PWR_D3CR_VOS_SCALE_3      (0x3)
+#define PWR_D3CR_VOS_SCALE_2      (0x2)
+#define PWR_D3CR_VOS_SCALE_1      (0x1)
+#define PWR_D3CR_VOS_MASK         (0x03)
 
 /* --- Function prototypes ------------------------------------------------- */
 enum pwr_svos_scale {
-  PWR_SCALE3 = PWR_CR1_SVOS_SCALE_3 << PWR_CR1_SVOS_SHIFT,
-  PWR_SCALE4 = PWR_CR1_SVOS_SCALE_4 << PWR_CR1_SVOS_SHIFT,
-  PWR_SCALE5 = PWR_CR1_SVOS_SCALE_5 << PWR_CR1_SVOS_SHIFT,
+  PWR_SVOS_SCALE3 = PWR_CR1_SVOS_SCALE_3 << PWR_CR1_SVOS_SHIFT,
+  PWR_SVOS_SCALE4 = PWR_CR1_SVOS_SCALE_4 << PWR_CR1_SVOS_SHIFT,
+  PWR_SVOS_SCALE5 = PWR_CR1_SVOS_SCALE_5 << PWR_CR1_SVOS_SHIFT,
+};
+
+enum pwr_vos_scale {
+  PWR_VOS_SCALE_0 = 0,    /* Note: This state requires SYSCFG ODEN set. */
+  PWR_VOS_SCALE_1 = (PWR_D3CR_VOS_SCALE_1 << PWR_D3CR_VOS_SHIFT),
+  PWR_VOS_SCALE_2 = (PWR_D3CR_VOS_SCALE_2 << PWR_D3CR_VOS_SHIFT),
+  PWR_VOS_SCALE_3 = (PWR_D3CR_VOS_SCALE_3 << PWR_D3CR_VOS_SHIFT)
 };
 
 BEGIN_DECLS
 
+/** @defgroup pwr_peripheral_api PWR Peripheral API
+ * @ingroup peripheral_apis
+@{*/
+
+/** Set power subsystem to utilize the LDO for CPU. */
+void pwr_set_mode_ldo(void);
+
+/** Set the voltage scaling/strength for the internal LDO when in Stop mode.
+ * @param[in] scale  Voltage scale value to set.
+ */
 void pwr_set_svos_scale(enum pwr_svos_scale scale);
+
+/** Set the voltage scaling/strength for the internal LDO while running.
+ * @param[in] scale  Voltage scale value to set.
+ */
+void pwr_set_vos_scale(enum pwr_vos_scale scale);
+
 
 END_DECLS
 
--- a/include/libopencm3/stm32/h7/rcc.h
+++ b/include/libopencm3/stm32/h7/rcc.h
@@ -27,46 +27,38 @@ LGPL License Terms @ref lgpl_license
 #ifndef LIBOPENCM3_RCC_H
 #define LIBOPENCM3_RCC_H
 
+#include <libopencm3/cm3/common.h>
 #include <libopencm3/stm32/h7/pwr.h>
 
 /**@{*/
 
-/* --- RCC registers ------------------------------------------------------- */
+/** @defgroup rcc_regisers RCC Registers
+ * @ingroup rcc_defines
+@{*/
 #define RCC_CR                    MMIO32(RCC_BASE + 0x000)
+#define RCC_ICSCR                 MMIO32(RCC_BASE + 0x004)  /* Y-devices only */
+#define RCC_HSICFGR               MMIO32(RCC_BASE + 0x004)  /* V-devices only */
+#define RCC_CRRCR                 MMIO32(RCC_BASE + 0x008)
+#define RCC_CSICFGR               MMIO32(RCC_BASE + 0x00C)  /* V-devices only */
 #define RCC_CFGR                  MMIO32(RCC_BASE + 0x010)
-
-/** @addtogroup rcc_cr_values RCC_CR_VALUES
- * @ingroup rcc_registers
-@{*/
-#define RCC_CR_PLL3AIRDY          (1 << 29)
-#define RCC_CR_PLL3AION           (1 << 28)
-#define RCC_CR_PLL2RDY            (1 << 27)
-#define RCC_CR_PLL2ON             (1 << 26)
-#define RCC_CR_PLL1RDY            (1 << 25)
-#define RCC_CR_PLL1ON             (1 << 24)
-#define RCC_CR_HSECSSON           (1 << 19)
-#define RCC_CR_HSEBYP             (1 << 18)
-#define RCC_CR_HSERDY             (1 << 17)
-#define RCC_CR_HSEON              (1 << 16)
-#define RCC_CR_D2CKRDY            (1 << 15)
-#define RCC_CR_D1CKRDY            (1 << 14)
-#define RCC_CR_HSI48RDY           (1 << 13)
-#define RCC_CR_HSI48ON            (1 << 12)
-#define RCC_CR_CSIKERON           (1 << 9)
-#define RCC_CR_CSIRDY             (1 << 8)
-#define RCC_CR_CSION              (1 << 7)
-#define RCC_CR_HSIDIVF            (1 << 5)
-#define RCC_CR_HSIDIV_MASK        (0x03)
-#define RCC_CR_HSIDIV_SHIFT       3
-#define RCC_CR_HSIDIV(n)          (((n) & RCC_CR_HSIDIV_MASK) << RCC_CR_HSIDIV_MASK)
-#define RCC_CR_HSIRDY             (1 << 2)
-#define RCC_CR_HSIKERON           (1 << 1)
-#define RCC_CR_HSION              (1 << 0)
-/**@}*/
-
-/** @addtogroup rcc_rstr_values RCC_RSTR_VALUES
- * @ingroup rcc_registers
-@{*/
+#define RCC_D1CFGR                MMIO32(RCC_BASE + 0x018)
+#define RCC_D2CFGR                MMIO32(RCC_BASE + 0x01C)
+#define RCC_D3CFGR                MMIO32(RCC_BASE + 0x020)
+#define RCC_PLLCKSELR             MMIO32(RCC_BASE + 0x028)
+#define RCC_PLLCFGR               MMIO32(RCC_BASE + 0x02C)
+/* PLLs are 1-based, so reference macros to 1..3, using index 0 will give undefined behavior. */
+#define RCC_PLLDIVR(n)            MMIO32(RCC_BASE + 0x030 + (0x08 * ((n) - 1)))
+#define RCC_PLLFRACR(n)           MMIO32(RCC_BASE + 0x030 + (0x08 * ((n) - 1)))
+#define RCC_PLL1DIVR              RCC_PLLDIVR(1)
+#define RCC_PLL1FRACR             RCC_PLLFRACR(1)
+#define RCC_PLL2DIVR              RCC_PLLDIVR(2)
+#define RCC_PLL2FRACR             RCC_PLLFRACR(2)
+#define RCC_PLL3DIVR              RCC_PLLDIVR(3)
+#define RCC_PLL3FRACR             RCC_PLLFRACR(3)
+#define RCC_D1CCIPR               MMIO32(RCC_BASE + 0x04C)
+#define RCC_D2CCIP1R              MMIO32(RCC_BASE + 0x050)
+#define RCC_D2CCIP2R              MMIO32(RCC_BASE + 0x054)
+#define RCC_D3CCIPR               MMIO32(RCC_BASE + 0x058)
 #define RCC_AHB1RSTR              MMIO32(RCC_BASE + 0x080)
 #define RCC_AHB2RSTR              MMIO32(RCC_BASE + 0x084)
 #define RCC_AHB3RSTR              MMIO32(RCC_BASE + 0x07C)
@@ -102,7 +94,37 @@ LGPL License Terms @ref lgpl_license
 #define RCC_DCKCFGR2              MMIO32(RCC_BASE + 0x90)
 /**@}*/
 
-/** @addtogroup rcc_cfgr_values RCC_CFGR_VALUES
+/** @defgroup rcc_cr_values RCC_CR Values
+ * @ingroup rcc_registers
+@{*/
+#define RCC_CR_PLL3RDY            BIT29
+#define RCC_CR_PLL3ON             BIT28
+#define RCC_CR_PLL2RDY            BIT27
+#define RCC_CR_PLL2ON             BIT26
+#define RCC_CR_PLL1RDY            BIT25
+#define RCC_CR_PLL1ON             BIT24
+#define RCC_CR_HSECSSON           BIT19
+#define RCC_CR_HSEBYP             BIT18
+#define RCC_CR_HSERDY             BIT17
+#define RCC_CR_HSEON              BIT16
+#define RCC_CR_D2CKRDY            BIT15
+#define RCC_CR_D1CKRDY            BIT14
+#define RCC_CR_HSI48RDY           BIT13
+#define RCC_CR_HSI48ON            BIT12
+#define RCC_CR_CSIKERON           BIT9
+#define RCC_CR_CSIRDY             BIT8
+#define RCC_CR_CSION              BIT7
+#define RCC_CR_HSIDIVF            BIT5
+#define RCC_CR_HSIDIV_MASK        (0x03)
+#define RCC_CR_HSIDIV_SHIFT       3
+#define RCC_CR_HSIDIV(n)          (((n) & RCC_CR_HSIDIV_MASK) << RCC_CR_HSIDIV_MASK)
+#define RCC_CR_HSIRDY             BIT2
+#define RCC_CR_HSIKERON           BIT1
+#define RCC_CR_HSION              BIT0
+/**@}*/
+
+
+/** @defgroup rcc_cfgr_values RCC_CFGR Values
  * @ingroup rcc_registers
 @{*/
 /* MCO2: Microcontroller clock output 2 */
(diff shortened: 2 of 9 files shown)


vielster:
@karlp this one's been hanging out a while, and is dense with functionality for the H7...actually allows it to get above 64MHz :-) 

Likely will be more changes to add more support to the tree as we bring up peripherals, since this thing has such a damn complicated clock tree, so would be great to get this merged.

include/libopencm3/stm32/h7/pwr.h
```diff
@@ -55,32 +55,80 @@ LGPL License Terms @ref lgpl_license
 /** Wakeup Domain Power Control register. */
 #define PWR_WKUPCR      MMIO32(POWER_CONTROL_BASE + 0x20)
 
+/** SYSCFG Register for configuring regulator "overdrive" */
+
```
karlp: this looks dangling?

include/libopencm3/stm32/h7/pwr.h
```diff
@@ -55,32 +55,80 @@ LGPL License Terms @ref lgpl_license
...
 /** DBP[8]: Disable backup domain write protection. */
 #define PWR_CR1_DBP               (1 << 8)
 
-/** PVDO: PVD output */
-#define PWR_CSR1_PVDO             (1 << 4)
+/** CSR1 Register Bits */
+#define PWR_CSR1_AVDO             BIT16
+#define PWR_CSR1_ACTVOS_SHIFT     14
+#define PWR_CSR1_ACTVOSRDY        BIT13
+#define PWR_CSR1_PVDO             BIT24
```
karlp: This is still bit4 in rev6 of rm0433?

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -27,46 +27,38 @@ LGPL License Terms @ref lgpl_license
...
+#define RCC_PLLCKSELR             MMIO32(RCC_BASE + 0x028)
+#define RCC_PLLCFGR               MMIO32(RCC_BASE + 0x02C)
+#define RCC_PLLDIVR(n)            MMIO32(RCC_BASE + 0x030 + (0x08 * (n)))
+#define RCC_PLLFRACR(n)           MMIO32(RCC_BASE + 0x030 + (0x08 * (n)))
+#define RCC_PLL1DIVR              MMIO32(RCC_BASE + 0x030)
+#define RCC_PLL1FRACR             MMIO32(RCC_BASE + 0x034)
+#define RCC_PLL2DIVR              MMIO32(RCC_BASE + 0x038)
+#define RCC_PLL2FRACR             MMIO32(RCC_BASE + 0x03C)
+#define RCC_PLL3DIVR              MMIO32(RCC_BASE + 0x040)
+#define RCC_PLL3FRACR             MMIO32(RCC_BASE + 0x044)
```
karlp: these could all just be efined as RCC_PLL3FRACR  RCC_PLLFRACR(3) right?

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -102,7 +94,37 @@ LGPL License Terms @ref lgpl_license
 #define RCC_DCKCFGR2              MMIO32(RCC_BASE + 0x90)
 /**@}*/
 
-/** @addtogroup rcc_cfgr_values RCC_CFGR_VALUES
+/** @defgroup rcc_cr_values RCC_CR Values
+ * @ingroup rcc_registers
+@{*/
+#define RCC_CR_PLL3AIRDY          BIT29
+#define RCC_CR_PLL3AION           BIT29
```
karlp: Where did these names come from?  RM0433 has them as PLL3RDY and PLL3ON?  (Also, you need to use BIT28 for one of them)

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -158,8 +180,133 @@ LGPL License Terms @ref lgpl_license
...
+#define RCC_D1CFGR_D1HPRE_DIV8      0xA
+#define RCC_D1CFGR_D1HPRE_DIV16     0xB
+#define RCC_D1CFGR_D1HPRE_DIV64     0xC
+#define RCC_D1CFGR_D1HPRE_DIV128    0xD
+#define RCC_D1CFGR_D1HPRE_DIV256    0xE
+#define RCC_D1CFGR_D1HPRE_DIV512    0xF
+
+#define RCC_D1CFGR_D1CPRE_SHIFT     8
+#define RCC_D1CFGR_D1PPRE_SHIFT     4
+#define RCC_D1CFGR_RSVD_BITMASK     (0xfffff080)
```
karlp: unless you've got evidence of registers where "keeping them at their reset value" is _required_, we've never done this before, and I don't really want to start doing it now.  In every other case so far, writing them to zero has been perfectly fine, and far less work to maintain when they add/document bits.

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -158,8 +180,133 @@ LGPL License Terms @ref lgpl_license
...
+/** @defgroup rcc_plldivr_values RCC_PLLnDIVR Values
+ * @ingroup rcc_registers
+ * @{*/
+#define RCC_PLLNDIVR_DIVR_SHIFT     24
+#define RCC_PLLNDIVR_DIVQ_SHIFT     16
+#define RCC_PLLNDIVR_DIVP_SHIFT     9
+#define RCC_PLLNDIVR_DIVN_SHIFT     0
+
+/* Need to preserve reserved bits, so give easy mask shortcut. */
+#define RCC_PLLNDIVR_RSVD_BITMASK   (0x80800100)
```
karlp: Here you've even artificually included one of the "illegal values" from a field, that's not actually reserved, making this even harder to track in the future.

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -181,23 +328,228 @@ LGPL License Terms @ref lgpl_license
...
+/** @defgroup rcc_d1ccipr_values RCC_D1CCIP1R Values
+ * @ingroup rcc_registers
+ * @{*/
+#define RCC_D2CCIPR_CKPERSEL_SHIFT     28
+
+#define RCC_D2CCIPR_CKPERSEL_HSI       0
+#define RCC_D2CCIPR_CKPERSEL_CSI       1
+#define RCC_D2CCIPR_CKPERSEL_HSE       2
+#define RCC_D2CCIPR_CKPERSEL_DISABLE   3
+#define RCC_D2CCIPR_CKPERSEL_MASK      3
```
karlp: These should all be D1 I believe?

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -181,23 +328,228 @@ LGPL License Terms @ref lgpl_license
...
+#define RCC_D2CCIP2R_USARTSEL_PLL2Q         1
+#define RCC_D2CCIP2R_USARTSEL_PLL3Q         2
+#define RCC_D2CCIP2R_USARTSEL_HSI           3
+#define RCC_D2CCIP2R_USARTSEL_CSI           4
+#define RCC_D2CCIP2R_USARTSEL_LSE           5
+#define RCC_D2CCIP2R_USARTSEL_MASK          7
+/**@}*/
+
+
+#define HSI_BASE_FREQUENCY              64000000UL
```
karlp: the prefix here is wrong, at least.  I want to see where you this this too :)

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -181,23 +328,228 @@ LGPL License Terms @ref lgpl_license
...
+  RCC_SPI6CLK,          /* SPI6 has its own kernel clock. */
+  RCC_USART1CLK,        /* USART 1/6 share kernel clock. */
+  RCC_USART6CLK,
+  RCC_USART2CLK,        /* USART 2/3/4/5/6/8 share kernel clock. */
+  RCC_USART3CLK,
+  RCC_USART4CLK,
+  RCC_USART5CLK,
+  RCC_USART7CLK,
+  RCC_USART8CLK,
+};
```
karlp: I'm a little concerned about how close this is to the bit macros used for enabling peripheral clocks, and being rather confusing.  We've also never (so far) really provided ways of looking up current clocks _back again_, we've normally considered that the user set the clock, providing backwrds functions wasn't really useful/necessary.  What do you see as the use case for these?

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -181,23 +328,228 @@ LGPL License Terms @ref lgpl_license
...
 
 enum rcc_osc {
   RCC_PLL,
   RCC_HSE,
   RCC_HSI,
-  RCC_LSE,
-  RCC_LSI
+  RCC_LSE
```
karlp: Why are we dropping LSI here again?

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -181,23 +328,228 @@ LGPL License Terms @ref lgpl_license
...
+  RCC_D1PPRE_DIV8 = RCC_D1CFGR_D1PPRE_DIV8 << RCC_D1CFGR_D1PPRE_SHIFT,
+  RCC_D1PPRE_DIV16 = RCC_D1CFGR_D1PPRE_DIV16 << RCC_D1CFGR_D1PPRE_SHIFT,
+};
+
+enum rcc_d1hpre {
+  RCC_D1HPRE_BYP = RCC_D1CFGR_D1HPRE_BYP,
+  RCC_D1HPRE_DIV2 = RCC_D1CFGR_D1HPRE_DIV2,
+  RCC_D1HPRE_DIV4 = RCC_D1CFGR_D1HPRE_DIV4,
+  RCC_D1HPRE_DIV8 = RCC_D1CFGR_D1HPRE_DIV8,
+  RCC_D1HPRE_DIV16 = RCC_D1CFGR_D1HPRE_DIV16,
```
karlp: I gotta say, all these just feels super spammy.  We've never put these wrappers in elsewhere, I don't really see why we need them here.  Everyone else just gets a uint32 for params.  IF, (and it's a big if) you want to move apis all over to be enums, I would say the enums should just _be_ the definitions, and would still be named accordingly, instead of this duplicate-but-not pasting.

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -181,23 +328,228 @@ LGPL License Terms @ref lgpl_license
...
+    enum rcc_d1hpre hclk3_prescale;   /**< HCLK3 prescaler for domain 1. */
+    enum rcc_d1ppre pclk3_prescale;   /**< APB3 Peripheral prescaler for domain 1. */
+  } domain1;
+  struct {
+    enum rcc_d2ppre1 pclk1_prescale;  /**< APB1 Peripheral prescaler for domain 2. */
+    enum rcc_d2ppre2 pclk2_prescale;  /**< APB2 Peripheral prescaler for domain 2. */
+  } domain2;
+  struct {
+    enum rcc_d3ppre pclk4_prescale;   /**< APB4 Peripheral prescaler for domain 3. */
+  } domain3;
```
karlp: unless you're expecting people to pass around and work withi the inner structs individually, just make it flat, there's no reason for the extra layers.

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -429,9 +781,64 @@ enum rcc_periph_rst {
...
+
+/**
+ * Get the clock rate (in hz) of the specified clock source. There are
+ * numerous clock sources and configurations on the H7, so rates for each
+ * configured peripheral clock are aimed to be discoverd/calculated by this
+ * module such that the user does not need to know how the MCU is configured
+ * in order to utilize a peripheral clock.
+ * @param[in] source  Clock source desired to be fetched.
+ * @return Clock rate in Hz for the specified clock. 0 if undefined or error.
+ */
```
karlp: would suggest, if you _really_ see how this is important, to make the name be get_clockrate or clockspeed, or something, to make it clear it's not getting the clock source

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -429,9 +781,64 @@ enum rcc_periph_rst {
...
+ * @param[in] source  Clock source desired to be fetched. Choose from
+ *                    RCC_D2CCIP1R_SPI123_XXX selections above.
+ */
+void rcc_set_spi123_clksel(uint8_t clksel);
+/**
+ * Set the clock select for the SPI 4/5 devices.
+ * @param[in] source  Clock source desired to be fetched. Choose from
+ *                    RCC_D2CCIP1R_SPI45_XXX selections above.
+ */
+void rcc_set_spi45_clksel(uint8_t clksel);
```
karlp: this is going to explode out too isn't it?  Isn't this what you were adding enums for all these clocks was for?

include/libopencm3/stm32/h7/syscfg.h
```diff
@@ -0,0 +1,78 @@
...
+#define SYSCFG_UR8				MMIO32(SYSCFG_BASE + 0x320)
+#define SYSCFG_UR9				MMIO32(SYSCFG_BASE + 0x324)
+#define SYSCFG_UR10				MMIO32(SYSCFG_BASE + 0x328)
+#define SYSCFG_UR11				MMIO32(SYSCFG_BASE + 0x32C)
+#define SYSCFG_UR12				MMIO32(SYSCFG_BASE + 0x330)
+#define SYSCFG_UR13				MMIO32(SYSCFG_BASE + 0x334)
+#define SYSCFG_UR14				MMIO32(SYSCFG_BASE + 0x338)
+#define SYSCFG_UR15				MMIO32(SYSCFG_BASE + 0x33C)
+#define SYSCFG_UR16				MMIO32(SYSCFG_BASE + 0x340)
+#define SYSCFG_UR17				MMIO32(SYSCFG_BASE + 0x344)
```
karlp: Uh, thanks ST.  I guess we keep it like this for consistency, but I'd be totally fine with just having a SYSCFG_UR(n) macro too.  Were you planning on adding in all the URxxx bits as well? or just leave that to the user?

lib/stm32/common/usart_common_all.c
```diff
@@ -52,17 +52,32 @@ usart_reg_base
 
 void usart_set_baudrate(uint32_t usart, uint32_t baud)
 {
+
+#if defined STM32H7
```
karlp: you've opened the box here :)  This is actually something that's an option (depending on the peripheral) on f4, f7, l4, l4+, g4 and I think even g0.   So.... yes, I see what the problem you're trying to solve is, but we might need to go back further and work out something that's going to be a bit more maintainable

lib/stm32/h7/pwr.c
```diff
@@ -19,8 +13,7 @@
 /*
  * This file is part of the libopencm3 project.
  *
- * Copyright (C) 2011 Stephen Caudle <scaudle@doceme.com>
- * Copyright (C) 2017 Matthew Lai <m@matthewlai.ca>
+ * Copyright (C) 2019 Brian Viele <vielster@allocor.tech>
```
karlp: it's generally considered bad form to drop other people's copyright statements...  I'd agree that at least scaudle probably had nothing in this file anyway, but mlai did the f7 code this was based on iirc?

lib/stm32/h7/pwr.c
```diff
@@ -37,14 +30,31 @@
...
-	PWR_CR1 = pwr_cr1_reg;
+	pwr_cr1_reg = (pwr_cr1_reg & ~(PWR_CR1_SVOS_MASK  << PWR_CR1_SVOS_SHIFT));
+	PWR_CR1 = pwr_cr1_reg | scale;
 }
 
-/**@}*/
\ No newline at end of file
+void pwr_set_vos_scale(enum pwr_vos_scale scale) {
+	uint32_t d3cr_masked = PWR_D3CR & ~(PWR_D3CR_VOS_MASK << PWR_D3CR_VOS_SHIFT);
+	if (scale == PWR_VOS_SCALE_0) {
```
karlp: Where does VOS_SCALE_0 even come from? there's no scale 0 in the docs.  Is this you attempting to set scale 1 if people use an unitialized variable?  I'd argue that either ignoring the request, errorring, blocking loop, or setting the default (scale 3) would all make more sense than hard setting to scale 1

karlp: (are you actually checking for vos scal1 here for the overdrive? might be worth a comment)

include/libopencm3/stm32/h7/pwr.h
```diff
@@ -55,32 +55,80 @@ LGPL License Terms @ref lgpl_license
...
-  PWR_SCALE3 = PWR_CR1_SVOS_SCALE_3 << PWR_CR1_SVOS_SHIFT,
-  PWR_SCALE4 = PWR_CR1_SVOS_SCALE_4 << PWR_CR1_SVOS_SHIFT,
-  PWR_SCALE5 = PWR_CR1_SVOS_SCALE_5 << PWR_CR1_SVOS_SHIFT,
+  PWR_SVOS_SCALE3 = PWR_CR1_SVOS_SCALE_3 << PWR_CR1_SVOS_SHIFT,
+  PWR_SVOS_SCALE4 = PWR_CR1_SVOS_SCALE_4 << PWR_CR1_SVOS_SHIFT,
+  PWR_SVOS_SCALE5 = PWR_CR1_SVOS_SCALE_5 << PWR_CR1_SVOS_SHIFT,
+};
+
+enum pwr_vos_scale {
+  PWR_VOS_SCALE_0 = 0,    /* Note: This state requires SYSCFG ODEN set. */
```
karlp: As we get to later, I'm not really sure you should have this case specially

lib/stm32/h7/rcc.c
```diff
@@ -6,16 +6,333 @@
...
+	uint16_t hse_khz;			/* This can't exceed 50MHz */
+} rcc_clock_tree = {
+	.sysclk_mhz = HSI_BASE_FREQUENCY / HZ_PER_MHZ,
+	.cpu_mhz = HSI_BASE_FREQUENCY / HZ_PER_MHZ,
+	.hclk_mhz = HSI_BASE_FREQUENCY / HZ_PER_MHZ,
+	.per.pclk1_mhz = HSI_BASE_FREQUENCY / HZ_PER_MHZ,
+	.per.pclk2_mhz = HSI_BASE_FREQUENCY / HZ_PER_MHZ,
+	.per.pclk3_mhz = HSI_BASE_FREQUENCY / HZ_PER_MHZ,
+	.per.pclk4_mhz = HSI_BASE_FREQUENCY / HZ_PER_MHZ
+};
```
karlp: this is for your "get_clock[rate]" stuff right?  this sort of thing _must_ be considered for everyone else too, if at all.

lib/stm32/h7/rcc.c
```diff
@@ -6,16 +6,333 @@
...
+		uint16_t pclk1_mhz;		/* APB1 clock. */
+		uint16_t pclk2_mhz;		/* APB2 clock. */
+		uint16_t pclk3_mhz;		/* APB3 clock. */
+		uint16_t pclk4_mhz;		/* APB4 clock. */
+	} per;
+	struct {					/* Each PLL output set of data. */
+		uint16_t p_mhz;
+		uint16_t q_mhz;
+		uint16_t r_mhz;
+	} pll[3];
```
karlp: why go to all the effort of structs and then force people to use magic indexes?  Either don't nest them, or name them so they're obvious!

lib/stm32/h7/rcc.c
```diff
@@ -6,16 +6,333 @@
...
+		RCC_PLLDIVR(i) &= RCC_PLLNDIVR_RSVD_BITMASK;
+		RCC_PLLDIVR(i) |= RCC_PLLNDIVR_DIVN(config->pll[i].divn);
+
+		/* Setup the PLL config values for this PLL. */
+		uint8_t vco_addshift = 4 * i; 		/* Values spaced by 4 for PLL 1/2/3 */
+		uint32_t clk = (config->pll_mux == RCC_PLL_HSI) ? HSI_BASE_FREQUENCY
+			: config->hse_frequency;
+		/* Set the PLL input frequency range. */
+		uint32_t pll_clk_mhz = (clk / config->pll[i].divm) / HZ_PER_MHZ;
+		if (pll_clk_mhz <= 2) {}	/* Do nothing, already masked to 0x00. */
```
karlp: just drop the line altogether.

lib/stm32/h7/rcc.c
```diff
@@ -6,16 +6,333 @@
...
+		case RCC_APB1CLK:
+			return rcc_clock_tree.per.pclk1_mhz * HZ_PER_MHZ;
+		case RCC_PCLK2:
+		case RCC_APB2CLK:
+			return rcc_clock_tree.per.pclk2_mhz * HZ_PER_MHZ;
+		case RCC_PCLK3:
+		case RCC_APB3CLK:
+			return rcc_clock_tree.per.pclk3_mhz * HZ_PER_MHZ;
+		case RCC_PCLK4:
+		case RCC_APB4CLK:
```
karlp: can't say I'm a fan of having duplicate names here.  Not sure what it gets us other than confusion and inconsistency if some places only take one or the other.

lib/stm32/h7/rcc.c
```diff
@@ -6,16 +6,333 @@
...
+			return rcc_clock_tree.per.pclk4_mhz * HZ_PER_MHZ;
+		case RCC_PERCLK:
+			clksel = (RCC_D1CCIPR >> RCC_D2CCIPR_CKPERSEL_SHIFT) & RCC_D2CCIPR_CKPERSEL_MASK;
+			if (clksel == RCC_D2CCIPR_CKPERSEL_HSI) {
+				return HSI_BASE_FREQUENCY;
+			} else if (clksel == RCC_D2CCIPR_CKPERSEL_HSE) {
+				return rcc_clock_tree.hse_khz * HZ_PER_KHZ;
```
karlp: why do you need to have things in both khz and mhz anyway?

include/libopencm3/stm32/h7/rcc.h
vielster: That's fine...I'm happy to get rid of these, I was just very pedantically following the user manual :-)

include/libopencm3/stm32/h7/rcc.h
vielster: I'm happy to rename these, but we have already proven the use case that this is pretty valuable in our H7 development. In short, this is the first time there has been a clock tree this complicated, and it makes pushing some of the work in clock discovery into the rcc implementation desirable. 

To do this without an actual tree discovery, the configuration structure has to hold all possible configurations of every clock selection in the system, and/or you need a global version of every clock in the system (which anyone can modify BTW). This makes the configuration kinda gross and unwieldy to get something working, plus the user then needs to know how to calculate each of these clocks by delving through the clock tree themselves. 

By breaking this up into a basic core set of configuration, just for the inputs and main PLLs, and then providing entry points for the clock tree variations, this makes setup a lot easier to manage (building blocks...core stuff works, then add my other clock selections). 

Clock discovery is not absolutely required, but as mentioned above, it provides encapsulation of the values, and it allows some optimization in storage since each value is not individually stored.

lib/stm32/h7/pwr.c
vielster: image

This is a value in the sheet, which is equivalent to VOS1 + boost/overdrive.

I've added a comment.

include/libopencm3/stm32/h7/pwr.h
vielster: Again, this is explicitly referenced in the docs.

lib/stm32/h7/rcc.c
vielster: It's a struct and an array so that it can be easily handled in a loop. The reason #define indexing is for documentation purposes only, because loop from 0-2 when the docs state PLL1-PLL3 is a bit confusing, so the macros are to document this.  I don't know a good way around ST's insistence on one-based numbering on a system that will be coded in a zero-based language. (although, we have found some of their peripherals actually have zero based defs too) 

I could have a dummy pll[0] that is unused?

lib/stm32/h7/rcc.c
vielster: This is mainly because the documentation refers to them with both names. The idea is that if the user is digging through the docs, and it says one or the other, they don't have to dig through the tree documentation to discover that the docs refer to PCLK4 (or per_clk4) which is the same thing as APB4, or vice versa.  It was unclear which of these will be the "correct" name to use if I were to eliminate one of them. Happy to eliminate one, but would like opinions on which to use.

lib/stm32/h7/rcc.c
vielster: I did it this way to reduce the storage size of the clocks.

include/libopencm3/stm32/h7/rcc.h
vielster: I did this specifically for the configuration parameters that go into the pll config struct only to constrain the selections, and allow the compiler/static analyzers/code indexers/etc. to let the user know when they've used the wrong value in the wrong place (not that I've ever done it before or anything :-)). The macros are used elsewhere, so I found having both definitions useful.  

The struct could be changed to be all _uint32_t_ values, but I thought this a pretty narrow area where adding a bit more explicit typing would be beneficial.

As you've noted below, I didn't go "all out" on this with everything, just the config struct. I can see the argument to all/none, so I will go ahead and eliminate the enums.

vielster:
Making changes...there may be a few intermediate pushes here so that I can test with my tree.

lib/stm32/h7/pwr.c
vielster: Re-added...I'm not sure if there is a single line that was retained across the two, but I also don't have any issue with sharing this copyright.

include/libopencm3/stm32/h7/rcc.h
vielster: Mainly for simplicity...I didn't think it useful to run off an internal 32KHz clock anytime in the near future, and it requires managing a bit more state.

include/libopencm3/stm32/h7/rcc.h
vielster: Went ahead and removed these...added some helper macros to make creating a config struct not so painful.

include/libopencm3/stm32/h7/rcc.h
vielster: how about `get_clock_freq()`

lib/stm32/common/usart_common_all.c
vielster: Yes, this happens on every platform, and the existing model was to ifdef out the special cases. I followed this, and my implementation doesn't block out the other platforms, it simply bypasses all of the other if/else logic for the H7 as it gets its clock in a differenter way.

Happy to refactor or have this refactored, but seems like it could/should be done outside of the scope of this review.

include/libopencm3/stm32/h7/rcc.h
vielster: If you look at the comment, it's explicitly stating to use the ```RCC_D2CCIP1R_SPI45_XXX``` selections in the header, which are small values. The function will handle the shifting and what not (which is why the input is clksel and not the reg value). 

I'm open to doing this another way, this was just the way that made sense to me such that the user didn't have to know about the format of the register.

lib/stm32/h7/rcc.c
vielster: Not sure what to do with this.  This was thinking about everyone else in that they don't need to know the clock tree to fetch a clock frequency.

lib/stm32/h7/rcc.c
vielster: Okay...I refactored this to be explicit values, and unrolled the loop into a function call. It's much cleaner now with the register defs and indexing.

include/libopencm3/stm32/h7/rcc.h
vielster: flattened this out and added prefix for the register for documentation.

vielster:
Okay...most of what you listed is addressed. The unresolved items are 

1. UART baudrate clock updates. My changes don't impact the current behavior, as they just if/else'd around the H7 implementation. 

1. Code to actually traverse the clock tree, and questions about why/what/how that's done. I feel pretty strongly that for a clock tree this complex, there should be some smarts to help the user. Additionally, this gives some encapsulation and optimization in storage.

1. Clksel functions presently take the clksel values directly (not the full register value) and the function does the shifting/masking appropriately. Question on whether this is okay, or if the user should pass the full reg value in, or whether we should do enums. 

1. LSI implementation which was left out. I see this as very low probability use case and didn't want to spend time to implement and try and test a 32KHz input clock. 

Thanks for doing this review, I know it was a lot of code.

include/libopencm3/stm32/h7/rcc.h
karlp: I'd really like to see some example code of this, not just the headers if possible.  We have discussed in the past that rather than just keeping three clock uint32s, (ahb/apb1/apb2) to keep the clock struct itself, and I do know that the "choose a clock for this peripheral!" makes things baloon out,

include/libopencm3/stm32/h7/rcc.h
karlp: well, you're slavishly holding all options elsewhere, and then not here?  I agree, running off lsi is pointless, but it still seems odd to just drop it.

include/libopencm3/stm32/h7/rcc.h
karlp: No, I mean, you're going to hve functions like this for every single peripheral.  You've only included a couple so far, but following this, they would all get their own function setter.  You've already made an enum for clocks, couldn't something like that be reused here?  (it's pretty much always the same options for the clocks anyway isn't it?)

lib/stm32/common/usart_common_all.c
karlp: it can be out of scope, but then I'd like to remove this stuff altogether, or limit it to "it only works if you use ahb clocks for usart, as done on "legacy" platforms"  I don want to have this come into the usart code as part of "just supporting some pll options on h7"

lib/stm32/h7/rcc.c
karlp: I'd go with APB, as that's the name used in other parts, both ST and other cortex.  At some point people have to know what they're doing, and I think allowing/encouraing mix and match here makes things worse not better.

include/libopencm3/stm32/h7/rcc.h
vielster: Here is a sample of the PLL configuration for our platform...
```c++
// ================================================================================================
static void InitLdoAndPll() {
  // Setup LDO for high speed operation.
  rcc_periph_clock_enable(RCC_SYSCFG);
  pwr_set_mode_ldo();                  // Ensure we're running on the LDO.
  pwr_set_vos_scale(PWR_VOS_SCALE_0);  // Set highest voltage output for 400+MHz.

  // Clock configuration for this platform.
  // clang-format off
  static const struct rcc_pll_config pll_config = {
    .hse_frequency = 8000000UL,
    .sysclk_mux    = RCC_SYSCLK_PLL,
    .pll_mux       = RCC_PLL_HSE,
    .pll1 = {
        .divm = 4U,     // 2MHz PLL1 base clock pre-multiplier, cancels post div2.
        .divn = 400U,   // 800MHz for PLL1 to drive the sysclk throu post-div of 2.
        .divp = 2U,     // PLL1P post-divider gives 400MHz output.
        .divq = 10U,    // PLL1Q post-divider gives 80MHz output for FDCAN.
                        // PLL1R are disabled for now.
    },
    .pll2 = {},
    .pll3 = {},
    // Set CPU to PLL1P output at 400MHz.
    .d1cfg_core_prescale  = RCC_D1CFGR_D1CPRE(RCC_D1CFGR_D1CPRE_BYP),
    // Constrain HPRE below 240MHz by dividing core clock by 2.
    .d1cfg_hclk3_prescale = RCC_D1CFGR_D1HPRE(RCC_D1CFGR_D1HPRE_DIV2),
    // Constrain APB3 below 120MHz by dividing hclk3 by 2.
    .d1cfg_pclk3_prescale = RCC_D1CFGR_D1PPRE(RCC_D1CFGR_D1PPRE_DIV2),
    // Constrain APB1 below 120MHz by dividing HCLK3 by 2.
    .d2cfg_pclk1_prescale = RCC_D2CFGR_D2PPRE1(RCC_D2CFGR_D2PPRE_DIV2),
    // Constrain APB2 below 120MHz by dividing HCLK3 by 2.
    .d2cfg_pclk2_prescale = RCC_D2CFGR_D2PPRE2(RCC_D2CFGR_D2PPRE_DIV2),
    // Constrain APB4 below 120MHz by dividing HCLK3 by 2.
    .d3cfg_pclk4_prescale = RCC_D3CFGR_D3PPRE(RCC_D3CFGR_D3PPRE_DIV2),
  };
  // clang-format on
  rcc_clock_setup_pll(&pll_config);

  // Setup SPI buses to use the HSI clock @ 64MHz for SPI log2 based dividers.
  rcc_set_spi123_clksel(RCC_D2CCIP1R_SPI123SEL_PERCK);  // PERCLK is defaulted to HSI.
  rcc_set_spi45_clksel(RCC_D2CCIP1R_SPI45SEL_HSI);

  // Set CANFD to use PLL1Q set at 80MHz. 
  rcc_set_fdcan_clksel(RCC_D2CCIP1R_FDCANSEL_PLL1Q);
}
```

```c++

The usart code update is one example, but here is a snippet from another one of our drivers that uses the get_clock_freq() interface:

```c++
ErrorReturn Stm32FdCan::SetBitrate(uint32_t bitrate, uint32_t fdrate) noexcept {
  std::lock_guard<std::mutex> lock(mutex_);
  struct can_bittiming nbt {};
  struct can_bittiming fdbt {};
  uint32_t base_freq = rcc_get_clock_freq(RCC_FDCAN1CLK);

  struct can_bittiming_const nbtc FDCAN_NOMINAL_BITTIMING_CONST;
  RET_USR_IF(can_get_bittiming(bitrate, 0, 1, base_freq, &nbtc, &nbt) == false,
             ErrorReturn::Invalid);
  if (fdrate > 0) {
    struct can_bittiming_const fdbtc FDCAN_FD_BITTIMING_CONST;
    RET_USR_IF(can_get_bittiming(fdrate, 0, 1, base_freq, &fdbtc, &fdbt) == false,
               ErrorReturn::Invalid);
  }
```

include/libopencm3/stm32/h7/rcc.h
vielster: So a few things...all of these can be seen if you reference Table 55 in [RM0433 rev 6.](https://www.st.com/content/ccc/resource/technical/document/reference_manual/group0/c9/a3/76/fa/55/46/45/fa/DM00314099/files/DM00314099.pdf/jcr:content/translations/en.DM00314099.pdf)

1. This is not for every peripheral, as some are grouped, and some are fixed, but yes, there will be a fair few.

1. Unfortunately no, the clksel values are not the same for each device type. In fact, they are even different for the same peripheral type (e.g. SPI1/2/3 and SPI4/5 have completely different selections). See the table, it's a bit of a mess. I can certainly convert these all to enums (code wouldn't change much), but I thought leaving the reg defines would be more appropriate because some folks may want to just make all of the settings themselves without using any of the helpers. I find the helpers useful for readability, modularity, etc.

1. The majority of these peripherals can just be left in their default configuration, and these functions won't need to be used. I thought that leaving this as individual settings would be best from a usage perspective because in most cases, the user probably only sets one or two of them to anything custom, and the rest would get stripped out.

lib/stm32/common/usart_common_all.c
vielster: Okay, I see...I will remove this alltogether. 

As a suggestion, I would maybe just add an entry point that is `usart_set_baudrate_with_clock(uint32_t usart, uint32_t clk, uint32_t baud)` which can then be called by custom functions for each platform rather than having all of the ifdefing?  Just a thought.  I'm simply going to remove these changes.

vielster: Actually...I can't simply leave as-is unfortunately. I will do some refactoring and you can let me know if you are okay with it...

vielster: Alright...after making various attempts at refactoring, the best way I could determine to do this is to move the H7 implementation out of the file and have the default implementation conditionally compiled. This achieves your goal of not mixing this more complicated clocking scheme into the "legacy" implementation.  Unfortunately, due to how everything is structured, there was not an easy way to move this function out of this file without potential breaking users. Possibly we could create a few variant `usart_xxx_setbaud.c` files and update each legacy type to build in the correct one, but this seemed a bit more risky than what I've implemented. Let me know what you think.

karlp:
I've not looked at the newest stuff again just yet, but hav ea look at rcc_set_peripheral_clk_sel for g0 and l0 for inspiration too

vielster:
The `l0` code is pretty similar to what I've done, a simple function for each clksel, and you pass the unshifted clksel value.

The `g0` code does the same, but embeds all of the functions into a single entry-point.

It's essentially...

```c++
rcc_set_<periph>_sel(clksel);
// vs. 
rcc_set_peripheral_clk_sel(<periph>, clksel); 
```

karlp:
l0 and g0 both have the same API right? slk_sel(periph, clk)?  I'd like to have the same api for g4, h7 f7 too.

vielster:
`l0` seems to have both

```
void rcc_set_peripheral_clk_sel(uint32_t periph, uint32_t sel);
void rcc_set_lptim1_sel(uint32_t lptim1_sel);
void rcc_set_lpuart1_sel(uint32_t lpupart1_sel);
void rcc_set_usart1_sel(uint32_t usart1_sel);
void rcc_set_usart2_sel(uint32_t usart2_sel);
```

I can add an additional entryping for `void rcc_set_peripheral_clk_sel(uint32_t periph, uint32_t sel)` my only significant concern is that since peripherals are grouped in the h7, someone might set UART2 peripheral, and will be setting UART2/3/4/5/6 or something like that.

vielster:
Okay...so it seams that you actually do want to have a "tree" like function that discovers the appropriate registers, etc. but you prefer to use the peripheral base address as the input.  I like this, as it makes it even easier for the user, so will update the get_clock_freq() function to take in a uint32_t which is the peripheral base, like the set function here.  Also will add the entry point like the others for setting the sel values.

vielster:
Use of the peripheral base address is nice, it cleans some things up really nicely.  The API now matches others for the _set_ side, and has a similar API to the _get_

```c++
uint32_t rcc_get_peripheral_clk_freq(uint32_t periph);
void rcc_set_peripheral_clk_sel(uint32_t periph, uint32_t sel);
```

vielster:
Thanks for all of the feedback. I think that with the updated API as you suggested, things are a bit cleaner and easier for users. This implementation is not yet complete (there are soooo many clocks on this damn MCU), but I feel pretty good about how it turned out after your inputs.

vielster:
Hey @karlp not sure where you got with this, but this is a pretty key piece of H7 functionality that would be nice to get merged. I'm sure there will be tweaks to be added (flash timings, additional clock controls, etc.), but getting this base in will be helpful such that future changes can be much smaller diff's

karlp:
No, just been really busy.  I still don't like the intrusvie clock getters, but I don't see an alternative to them at the moment, so I'm inclined to merge it and see where it goes :)

include/libopencm3/stm32/h7/rcc.h
karlp: None of these use the enum above?  I'm still concerned about how dense RCC_<something clockish> namespace is, when we used the short form of them for all the clock enabling.  I'm going to try this out, but I'm not excited :|

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -200,6 +439,37 @@ enum rcc_osc {
   RCC_LSI
 };
 
+enum rcc_sysclk_mux {
+  RCC_SYSCLK_PLL,
+  RCC_SYSCLK_HSE,
+  RCC_SYSCLK_HSI,
```
karlp: how are these differrent to the defines that every other platforrm has for sysclock sources?  Why does all of this platform need a sea of enums?  (and enums that are often just equal to some defines?)

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -200,6 +439,37 @@ enum rcc_osc {
...
+
+enum rcc_pll_mux {
+  RCC_PLL_HSI = RCC_PLLCKSELR_PLLSRC_HSI,
+  RCC_PLL_HSE = RCC_PLLCKSELR_PLLSRC_HSE
+};
+
+/** PLL Configuration structure. */
+struct rcc_pll_config {
+  uint32_t hse_frequency;           /**< User configured external crystal frequency. */
+  enum rcc_sysclk_mux sysclk_mux;   /**< SYSCLK source input selection. */
```
karlp: Why not just the enum rcc_osc like others have?  G0 is probably the closest here.  I know h7 is the biggest and most complicated, but there's no reason to make it all up completely new and different.  You can just have the pll source as well, without the renaming enum on line 448.

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -200,6 +439,37 @@ enum rcc_osc {
...
+    uint8_t divq;                   /**< Post divider for PLLQ clock. */
+    uint8_t divr;                   /**< Post divider for PLLR clock. */
+  } pll1, pll2, pll3;               /**< PLL1-PLL3 configurations. */
+  uint32_t d1cfg_core_prescale;     /**< Core prescaler for domain 1. */
+  uint32_t d1cfg_hclk3_prescale;    /**< HCLK3 prescaler for domain 1. */
+  uint32_t d1cfg_pclk3_prescale;    /**< APB3 Peripheral prescaler for domain 1. */
+  uint32_t d2cfg_pclk1_prescale;    /**< APB1 Peripheral prescaler for domain 2. */
+  uint32_t d2cfg_pclk2_prescale;    /**< APB2 Peripheral prescaler for domain 2. */
+  uint32_t d3cfg_pclk4_prescale;    /**< APB4 Peripheral prescaler for domain 3. */
+};
```
karlp: This is _similar_ but not to the existing rcc_clock_scale structs.  I know it does more, but it brings in all of the pclk prescaler selection, makes it massively incompatible.  G0 and L4 are probably closes here too, where it's set by subsequent calls?

include/libopencm3/stm32/h7/rcc.h
karlp: so, can you use _either_ rcc_set_peripheral_clk_sel or the specific ones? or do you need to use both?  if you can use either, please drop all the specific ones, it's marginally easier to optimize out, sure, but it's a pile of useless code otherwise

lib/stm32/h7/usart.c
```diff
@@ -0,0 +1,52 @@
...
+but for simplicity, use the first from grooup.
+
+@param[in] usart unsigned 32 bit. USART block register address base @ref
+usart_reg_base
+@param[in] baud unsigned 32 bit. Baud rate specified in Hz.
+*/
+
+void usart_set_baudrate(uint32_t usart, uint32_t baud)
+{
+	uint32_t clock = rcc_get_peripheral_clk_freq(usart);
```
karlp: This wasn't my intention really, but maybe it works out like this?  *laughs*

karlp:
I've merged this in 2ca56f4c hopefully it will develop well.  I stripped out the usart code, that's gotta go in it's own commit and it didn't seem productive to try and do it here.  I still have serious reservations about much of this code, but it's certainly no worse than plenty of other stuff we have, and you guys seem motivated to look after it, so let's do that, and see where it all ends up.

vielster:
Thanks @karlp I've created a new pull request to address some of your comments, especially got a bit more consistent on naming and removed some of the remaining enums. I need to split out the usart change still, but I think you'll be less unhappy with it after the changes I've made :-)

karlp:
BTW, don't mind me being grumpy too much, this idea of having to look up clocks or save clock structs ins't new, and the problem has been seen before and your method is even one of the ways suggested: see this old RFC for instance https://github.com/libopencm3/libopencm3/pull/410

So, just keep on doing things, it will all be fine in  the end :)

PR #1155, opened 2019-12-31
Fix garbage config returned for wrong index

fornellas:
I used [USB20CV](https://www.usb.org/usb2tools) to test [my procject](https://github.com/fornellas/3dpkbd2) compliance with USB specification, but all tests failed. The root cause is that all tests send a SETUP request, asking for the configuration descriptor, but it **also** does a second request for an **invalid** configuration descriptor, at an invalid index. It is expected for the device to reject such requests, but currently libopencm3 does reply, with garbage data, making all tests fail.

This PR adds a check, based on the device descriptor `bNumConfigurations`, whether or not the requested configuration descriptor index actually exists, and only replies if this is valid.

With this fix, I managed to get all tests from USB20CV to pass on my project.


karlp:
Please add this to the gadget 0 tests too. https://github.com/libopencm3/libopencm3/blob/master/tests/gadget-zero/test_gadget0.py#L40

karlp:
I'm also interested in what you had to do to run the tests, last time I tried the suite I couldn't get it t work, but it's been a while, my memory might be foggy :)

fornellas:
I had to use another computer to control keyboard and mouse (with Synergy), and I suppose I was lucky that my USB host controller works with the driver provided by the test. It is cumbersome, and gives limited error messages, but I managed to capture live USB packets from the wires with a logic analyser so I knew what to fix.

I'll add a test.

fornellas:
@karlp Sorry it took a few months to get back :-D

I added a test. Unfortunately, I don't have hardware at hand to promptly test it with gadged0, but I managed to manually repro the test failing / passing before / after this PR at my project.

fornellas:
@karlp the gadget0 build failure seems unrelated (other unrelated PRs are failing similarly). Once this is fixed, do you think we can have this merged, or do you think there are changes required to the code? Thanks.

karlp:
no, it looks ok, I'm just gettting back onto locm3 stuff again.

fornellas:
Ping @karlp .

PR #1156, opened 2020-01-02
stm32: extract g0 exti stuff to exti_common_v2.

grevaillot:
The same EXTI peripheral is used on stm32l5 (plus some secure/privileged stuff). Extract relevant code to common_exti_v2, with improved documentation.


karlp:
merged thanks!

PR #1157, opened 2020-01-02
cortex: FAULTMASK does not exist on armv6m.

grevaillot:
I'm currently looking at m33/m23 cortex, and comparing architectures, and apparently we provide methods to access faultmask on m0+, while this register does not exist : remove cm_is_masked_faults/cm_mask_faults

Maybe there's a less brutal way to remove these methods, or maybe this already have been discussed, but i could not find anything.

### DIFF
--- a/include/libopencm3/cm3/cortex.h
+++ b/include/libopencm3/cm3/cortex.h
@@ -91,6 +91,7 @@ static inline bool cm_is_masked_interrupts(void)
 	return result;
 }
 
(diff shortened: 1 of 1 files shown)


karlp:
seems reasonable, it's the same sort of macros we're using elsewhere to block things for v6m, so should be fine

karlp:
merged in 9d15ac7a thanks!

PR #1158, opened 2020-01-04
Make use of hex literal case consistent.

via:
The use of 0X confuses zig's translate-c at the moment. '0x' is used
predominantly, so convert the 0Xs.


karlp:
this is pointless churn, and only a _single_ line is even in code, not comments.   "consistency" sure, but if zig can't understand hex literals, sooner or later you _need to fix zig_  If you want to change just the code line, sure, I could accept that.

via:
@karlp this will be fixed in zig with https://github.com/ziglang/zig/pull/4072.  I realize its just the one line of code, it just seemed like it was unintentional. I'm updating my branch to only change the code line if its still something you would consider.

karlp:
I'll merge a change to just the code if you really think it's important.

PR #1161, opened 2020-01-06
SWM050: Adds timer peripheral and updates the README.

Szalacinski:
This commit adds the final peripheral (timers) to the SWM050.  Support for the MCU is now feature-complete.

### DIFF
--- a/include/libopencm3/swm050/timer.h
+++ b/include/libopencm3/swm050/timer.h
@@ -1,0 +1,233 @@
+/** @defgroup timer_defines Timer Defines
+ *
+ * @brief <b>Defined Constants and Types for the SWM050 Timer</b>
+ *
+ * @ingroup SWM050_defines
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2020 Caleb Szalacinski <contact@skiboy.net>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+/**@{*/
+#ifndef LIBOPENCM3_TIMER_H
+#define LIBOPENCM3_TIMER_H
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/swm050/memorymap.h>
+
+/* Timer select */
+/** @defgroup timer_select Timer Select
+@{*/
+#define TIMER_SE0	TIMER_SE0_BASE
+#define TIMER_SE1	TIMER_SE1_BASE
+/*@}*/
+
+/* Timer level definitions */
+/** @defgroup timer_level Timer Level
+@{*/
+enum timer_level {
+	TIMER_LEVEL_LOW,
+	TIMER_LEVEL_HIGH
+};
+/*@}*/
+
+/* Timer edge mode definitions */
+/** @defgroup timer_edge_modes Timer Edge Modes
+@{*/
+enum timer_edge_modes {
+	/* Trigger on rising edge */
+	TIMER_EDGE_RISING,
+	/* Trigger on falling edge */
+	TIMER_EDGE_FALLING
+};
+/*@}*/
+
+/* Timer operation mode definitions */
+/** @defgroup timer_operation_modes Timer Operation Modes
+@{*/
+enum timer_operation_modes {
+	TIMER_MODE_COUNTER,
+	TIMER_MODE_PWM,
+	TIMER_MODE_PULSE_CAPTURE,
+	TIMER_MODE_DUTY_CYCLE_CAPTURE
+};
+/*@}*/
+
+/* Timer clock source definitions */
+/** @defgroup timer_clk_src Timer Clock Source
+@{*/
+enum timer_clk_src {
+	TIMER_CLK_INTERNAL,
+	TIMER_CLK_EXTERNAL
+};
+/*@}*/
+
+/* Timer interrupt mask definitions */
+/** @defgroup timer_int_masked Timer Interrupt Mask
+@{*/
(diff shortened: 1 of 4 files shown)


karlp:
Why have you tried to have an enum, that you then treat as a bool for the timer peripheral?  Why not use the timer register base like all other targets do for all other peripherals?  It woudl get rid of all your "if (timer) { case c, } else { case b }" code throughout.  It seems really obtuse to do it like this.

include/libopencm3/swm050/timer.h
```diff
@@ -0,0 +1,253 @@
...
+/** Interrupt enable */
+#define TIMER_INTCTL_INTEN	1
+/*@}*/
+
+/* Timer registers */
+/** @defgroup timer_registers Timer Registers
+@{*/
+/** Timer control register */
+#define TIMER_SE0_CTRL			MMIO32(TIMER_SE0_BASE + 0x0)
+#define TIMER_SE1_CTRL			MMIO32(TIMER_SE1_BASE + 0x0)
```
karlp: if you look at how other families do this sort of thing, you just get

```
#define TIMER_CTRL(x)  MMIO32(x + 0)
```
And you call it with TIMER_CTRL(TIMER_SE0)  You'll sometimes optionally see explicit per peripheral instance register defines provided as well, but the macro allows the c implentations to take the timer as simply a peripheral base address, instead of an extra indirection layer via your enums.

Szalacinski: Fixed.

lib/swm050/timer.c
```diff
@@ -0,0 +1,457 @@
...
+	}
+}
+
+/**
+ * Enables or disables the timer's internal clock.
+ * @param timer Select timer @ref timer_select
+ * @param en Enable or disable the internal clock
+ */
+void timer_clock_enable(uint32_t timer, bool en)
+{
```
karlp: traditionally, enabling clocks in a different peripheral would be the file for that peripheral, (ie, clk.c) not here, but I can see how much of a mess this target is, and you say this "finishes" the target, so it can just stay like this.

lib/swm050/timer.c
```diff
@@ -0,0 +1,457 @@
...
+ * This has no use in PWM mode.
+ * In loop mode with counter mode, the counter will constantly loop.
+ * In loop mode with the capture modes, the values will be captured
+ * again and again. In single mode, these operations happen only once.
+ * @param timer Select timer @ref timer_select
+ * @param mode The loop mode @ref timer_loop_modes
+ */
+void timer_loop_mode(uint32_t timer, enum timer_loop_modes mode)
+{
+	if (mode) {
```
karlp: I still feel we must have misunderstood each other somewhere that you are so wrapped up in using enums, and then treating them as booleans.  for enumerations of only two choices in other families, we get functions that do "set mode A" (timer_set_loop_mode) and "set mode B" (timer_disable_loop_mode) (which is verbose and wordy, sure)   All of this code will fall apart if anyone ever adds to the enum.  I know this is "complete" so I'm going to merge it as is, but it's.... weird to try it like this, and I don't really understand the motivation.  Using a switch also works, though is marginally bigger when compiled.

include/libopencm3/swm050/timer.h
```diff
@@ -0,0 +1,233 @@
...
+enum timer_pwm_period {
+	TIMER_PERIOD_0,
+	TIMER_PERIOD_1
+};
+/*@}*/
+
+/* Timer clock divider mask */
+/** @defgroup timer_div_mask Timer Clock Divider Mask
+@{*/
+#define TIMER_DIV_MASK			~(0x3F << 16)
```
karlp: ok, this is something I didn't notice earlier.  We definitely don't store inverted masks.  All the masks are the natural mask, so 0x3f.  We generally also keep them unshifted, except for single bits, but I'm ok with you keeping them shifted in your own target here if you like, but the masks are for masking user variables not for internally clearing other fields.  Look at the inconsistency you create even in this file, in for instance:

```
SYSCTL_SYS_CFG_1 &= TIMER_DIV_MASK | ((~0x01 & div) << 16);
vs
TIMER_CTRL(timer) &= ~TIMER_CTRL_EN;
```

have a look at say, "ack '&=' lib/stm32/common" for how consistently this is normally done.

karlp: I'm afraid I don't feel comfortable making this change myself locally though

lib/swm050/timer.c
Szalacinski: What about timer_enable_loop_mode(uint32t timer, bool en)?

include/libopencm3/swm050/timer.h
Szalacinski: Fixed.

lib/swm050/timer.c
```diff
@@ -0,0 +1,453 @@
...
+}
+
+/**
+ * Selects the mode of operation.
+ * @param timer Select timer @ref timer_select
+ * @param mode The mode of operation @ref timer_operation_modes
+ */
+void timer_operation_mode(uint32_t timer, enum timer_operation_modes mode)
+{
+	TIMER_CTRL(timer) &= ~TIMER_OPERATION_MODE_MASK | (mode << 4);
```
karlp: This doesn't appear to be correct (same for the others)  This doesn't really set things the way you think.  Here's a snippet testing it: https://gist.github.com/karlp/eda0f04e46e4846ae3f5bd07c3383c35

Szalacinski: Thanks for catching that.  It should be fixed now.

lib/swm050/timer.c
```diff
@@ -0,0 +1,455 @@
...
+ *	treated by the register as a 1, which would likely be unexpected. A
+ *	value of 0 would also normally be treated as a 2, which would also be
+ *	unexpected behavior.
+ */
+void timer_clock_div(uint8_t div)
+{
+	/* If the value is 0 or 1, make it odd, meaning no divide. */
+	/* Otherwise, drop div to the closest even value. */
+	div = (div <= 1) ? 1 : (div & ~0x1);
+	SYSCTL_SYS_CFG_1 = (~TIMER_DIV_MASK & SYSCTL_SYS_CFG_1) | (div << 16);
```
karlp: The manual says this field is in syscfg0?  Are you sure this works?  Is this a doc error or a code error?  (Don't push any changes though, I'm just in the process of merging this, make a new PR "shortly" if this is a real problem

Szalacinski: Yeah, this is a code error.  I'll submit a PR to take care of it after it's merged.

karlp:
merged in f06a1ca9 thanks.  Please do check that tim_div in syscfg 0 or 1 for me though please.

lib/swm050/timer.c
karlp: groovy. sorry this one took so long

PR #1162, opened 2020-01-10
Fixed a typo in i2c_read7_v1 and i2c_write7_v1

kapacuk:
Replaced & with &&
Wait for all three bits to be set - SB, MSL, and BUSY.


karlp:
As mentioned, there's definitely something fishy with the &, still trepidation that this is the correct code, if this, "also works"

karlp:
merged via rebase in 08c5f2a1fb16b279907eb622ac043046d7a1a35c thanks!

PR #1166, opened 2020-01-20
STM32 FDCAN Register Definitions

mattofak:
Register addresses and definitions taken from the STM32H7 manual.

### DIFF
--- a/include/libopencm3/cm3/common.h
+++ b/include/libopencm3/cm3/common.h
@@ -111,4 +111,26 @@
 #define BIT30 (1<<30)
 #define BIT31 (1<<31)
 
+/* Generic bit mask manipulation */
+/** Set a \p field inside of \p reg_ with bits_. Assumes that the field name can
+ * expand into two definitions _MASK and _SHIFT where the mask is applied before
+ * the shift.
+ * @param reg_   Starting value of register
+ * @param field_ Prefix of field to set in \p reg_
+ * @param bits_	 Value to set in \p field_
+ * @returns The new register value.
+ */
+#define CM3_SET_MASK(reg_, field_, bits_)                    \
+	((((reg_) & (~(field_##_MASK << field_##_SHIFT)))) | \
+	 (((bits_) & field_##_MASK) << field_##_SHIFT))
+
+/** Obtain a \p field inside of \p reg_. Assumes that the field name can expand
+ * into two definitions _MASK and _SHIFT where the shift is applied before the
+ * mask.
+ * @param reg_   Current value of register
+ * @param field_ Prefix of field to extract from \p reg_
+ * @returns Value of the field inside the register.
+ */
+#define CM3_GET_MASK(reg_, field_) (((reg_) >> field_##_SHIFT) & field_##_MASK)
+
 #endif
--- a/include/libopencm3/stm32/common/fdcan_common.h
+++ b/include/libopencm3/stm32/common/fdcan_common.h
@@ -1,0 +1,1614 @@
+/** @defgroup fdcan_defines FDCAN defines
+ * @ingroup STM32_defines
+ * @brief <b>libopencm3 Defined Constants and Types for STM32 FDCAN </b>
+ * @version 1.0.0
+ * LGPL License Terms @ref lgpl_license
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2020 Matt Walker <m@allocor.tech>
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* THIS FILE SHOULD NOT BE INCLUDED DIRECTLY, BUT ONLY VIA FDCAN.H
+ * The order of header inclusion is important. fdcan.h includes the device
+ * specific memorymap.h header before including this header file. */
+
+/** @cond */
+#ifdef LIBOPENCM3_STM32_FDCAN_H
+/** @endcond */
+
+#include <libopencm3/can/can_bit_timing.h>
+
+#ifndef LIBOPENCM3_STM32_COMMON_FDCAN_COMMON_H
+#define LIBOPENCM3_STM32_COMMON_FDCAN_COMMON_H
+
+/** @defgroup fdcan_registers CAN Register Macros
+ * @ingroup fdcan_defines
+ * @{*/
+/** FDCAN core release register (FDCAN_CREL) */
+#define FDCAN_CRELR(fdcan_base)	MMIO32((fdcan_base) + 0x0000)
+/** FDCAN Endian register (FDCAN_ENDN) */
+#define FDCAN_ENDNR(fdcan_base)		MMIO32((fdcan_base) + 0x0004)
+/** FDCAN data bit timing and prescaler register (FDCAN_DBTP) */
+#define FDCAN_DBTPR(fdcan_base) 	MMIO32((fdcan_base) + 0x000C)
+/** FDCAN test register (FDCAN_TEST)*/
+#define FDCAN_TESTR(fdcan_base) 	MMIO32((fdcan_base) + 0x0010)
+/** FDCAN RAM watchdog register (FDCAN_RWD) */
+#define FDCAN_RWDR(fdcan_base) 		MMIO32((fdcan_base) + 0x0014)
+/** FDCAN CC control register (FDCAN_CCCR) */
+#define FDCAN_CCCRR(fdcan_base) 	MMIO32((fdcan_base) + 0x0018)
+/** FDCAN nominal bit timing and prescaler register (FDCAN_NBTP) */
+#define FDCAN_NBTPR(fdcan_base) 	MMIO32((fdcan_base) + 0x001C)
+/** FDCAN timestamp counter configuration register (FDCAN_TSCC) */
+#define FDCAN_TSCCR(fdcan_base) 	MMIO32((fdcan_base) + 0x0020)
+/** FDCAN timestamp counter value register (FDCAN_TSCV) */
+#define FDCAN_TSCVR(fdcan_base) 	MMIO32((fdcan_base) + 0x0024)
+/** FDCAN timeout counter configuration register (FDCAN_TOCC) */
+#define FDCAN_TOCCR(fdcan_base) 	MMIO32((fdcan_base) + 0x0028)
+/** FDCAN timeout counter value register (FDCAN_TOCV) */
+#define FDCAN_TOCVR(fdcan_base) 	MMIO32((fdcan_base) + 0x002C)
+/** FDCAN error counter register (FDCAN_ECR) */
+#define FDCAN_ECRR(fdcan_base) 		MMIO32((fdcan_base) + 0x0040)
+/** FDCAN protocol status register (FDCAN_PSR) */
+#define FDCAN_PSRR(fdcan_base) 		MMIO32((fdcan_base) + 0x0044)
+/** FDCAN transmitter delay compensation register (FDCAN_TDCR) */
+#define FDCAN_TDCRR(fdcan_base) 	MMIO32((fdcan_base) + 0x0048)
+/** FDCAN interrupt register (FDCAN_IR) */
+#define FDCAN_IRR(fdcan_base) 		MMIO32((fdcan_base) + 0x0050)
+/** FDCAN interrupt enable register (FDCAN_IE) */
+#define FDCAN_IER(fdcan_base) 		MMIO32((fdcan_base) + 0x0054)
+/** FDCAN interrupt line select register (FDCAN_ILS) */
+#define FDCAN_ILSR(fdcan_base) 		MMIO32((fdcan_base) + 0x0058)
+/** FDCAN interrupt line enable register (FDCAN_ILE) */
+#define FDCAN_ILER(fdcan_base) 		MMIO32((fdcan_base) + 0x005C)
+/** FDCAN global filter configuration register (FDCAN_GFC) */
+#define FDCAN_GFCR(fdcan_base) 		MMIO32((fdcan_base) + 0x0080)
+/** FDCAN standard ID filter configuration register (FDCAN_SIDFC) */
+#define FDCAN_SIDFCR(fdcan_base) 	MMIO32((fdcan_base) + 0x0084)
+/** FDCAN extended ID and mask register (FDCAN_XIDAM) */
+#define FDCAN_XIDFCR(fdcan_base) 	MMIO32((fdcan_base) + 0x0088)
+/** FDCAN extended ID and mask register (FDCAN_XIDAM) */
+#define FDCAN_XIDAMR(fdcan_base) 	MMIO32((fdcan_base) + 0x0090)
+/** FDCAN high priority message status register (FDCAN_HPMS) */
+#define FDCAN_HPMSR(fdcan_base) 	MMIO32((fdcan_base) + 0x0094)
+/** FDCAN new data 1 register (FDCAN_NDAT1) */
+#define FDCAN_NDAT1R(fdcan_base) 	MMIO32((fdcan_base) + 0x0098)
+/** FDCAN new data 2 register (FDCAN_NDAT2) */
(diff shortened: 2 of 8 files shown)


include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
+/** @defgroup fdcan_defines FDCAN defines
+@ingroup STM32H7_defines
```
karlp: If you're putting this in a common directory, it can't be in the h7 file grouping please.

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/stm32/memorymap.h>
+
+/**@{
+ * @defgroup fdcan_reg_base CAN register base address
+ * @ingroup fdcan_defines
+ * @{*/
+#define FDCAN1			FDCAN1_BASE
+#define FDCAN2			FDCAN2_BASE
```
karlp: I think these sorts of defines are normalyl in the part specific header, like h7/fdcan.h?

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+
+#ifndef LIBOPENCM3_STM32_FDCAN_H
+#define LIBOPENCM3_STM32_FDCAN_H
+
+#include <libopencm3/cm3/common.h>
+#include <libopencm3/stm32/memorymap.h>
+
+/**@{
+ * @defgroup fdcan_reg_base CAN register base address
+ * @ingroup fdcan_defines
```
karlp: you don't need any of these @ingroup fdcan_defines when it's the logical nesting level (like your line 1 group definition) you only need an @ingroup if you want to put it "elsewhere"

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+#define FDCAN_CCCR_NISO BIT15
+/* Edge Filtering during bus Integration */
+#define FDCAN_CCCR_EFBI BIT13
+/* Protocol exception handling disable */
+#define FDCAN_CCCR_PXHD BIT12
+/* FDCAN bitrate Switching */
+#define FDCAN_CCCR_BRSE BIT9
+/* FD operation enable */
+#define FDCAN_CCCR_FDOE BIT8
+/* Test mode enable */
```
karlp: If you add the second * to all of these, they get doxygen too :)  (no big deal, but when you have the descriptions already)

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+/* Accept non-matching frames standard */
+enum fdcan_gfc_anfs {
+	FDCAN_GFC_ANFS_ACCEPT_FIFO0	= 0x0,
+	FDCAN_GFC_ANFS_ACCEPT_FIFO1	= 0x1,
+	FDCAN_GFC_ANFS_REJECT		= 0x2,
+	FDCAN_GFC_ANFS_MASK			= 0x3,
+};
+#define FDCAN_GFC_ANFS_SHIFT 4
+/* Accept non-matching frames extended */
+enum fdcan_gfc_anfe {
```
karlp: is there much value in these two enums being separately named? I'm not a heavy can user, would it be nicer/equally sane from an enduser point of view to just have FDCAN_GFC_ACCEPTFIFO0/1/REJECT/MASK instead of duped for anfs and anfe?

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+
+/* --- FDCAN_NDAT1: FDCAN new data 1 register ------------------------------- */
+/* New data[31:0] */
+#define FDCAN_NDAT1_ND_MASK 0xFFFFFFFF
+#define FDCAN_NDAT1_ND_SHIFT 0
+
+/* --- FDCAN_NDAT2: FDCAN new data 2 register ------------------------------- */
+/* New data[63:32] */
+#define FDCAN_NDAT2_ND_MASK 0xFFFFFFFF
+#define FDCAN_NDAT2_ND_SHIFT 0
```
karlp: I don't really see any point to these NDAT1 and NDAT2 defines?  Is this all autogenerated?

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+	FDCAN_RXESC_F0DS_8		= 0x0,
+	FDCAN_RXESC_F0DS_12		= 0x1,
+	FDCAN_RXESC_F0DS_16		= 0x2,
+	FDCAN_RXESC_F0DS_20		= 0x3,
+	FDCAN_RXESC_F0DS_24		= 0x4,
+	FDCAN_RXESC_F0DS_32		= 0x5,
+	FDCAN_RXESC_F0DS_48		= 0x6,
+	FDCAN_RXESC_F0DS_64		= 0x7,
+	FDCAN_RXESC_F0DS_MASK	= 0x7,
+};
```
karlp: same comment as earlier, these are all the same, it doesn't seem useful/necessary to repeat them

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+	FDCAN_TXESC_TBDS_8		= 0x0,
+	FDCAN_TXESC_TBDS_12		= 0x1,
+	FDCAN_TXESC_TBDS_16		= 0x2,
+	FDCAN_TXESC_TBDS_20		= 0x3,
+	FDCAN_TXESC_TBDS_24		= 0x4,
+	FDCAN_TXESC_TBDS_32		= 0x5,
+	FDCAN_TXESC_TBDS_48		= 0x6,
+	FDCAN_TXESC_TBDS_64		= 0x7,
+	FDCAN_TXESC_TBDS_MASK	= 0x7,
+};
```
karlp: and here

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+/* --- FDCAN_TXBTIE: FDCAN Tx buffer transmission interrupt enable register - */
+/* Transmission interrupt enable */
+#define FDCAN_TXBTIE_TIE_MASK 0xFFFFFFFF
+#define FDCAN_TXBTIE_TIE_SHIFT 0
+
+/* --- FDCAN_TXBCIE: FDCAN Tx buffer cancellation finished interrupt enable
+ * register  */
+/* Cancellation finished interrupt enable. */
+#define FDCAN_TXBCIE_CFIE_MASK 0xFFFFFFFF
+#define FDCAN_TXBCIE_CFIE_SHIFT 0
```
karlp: and all of these up to line 804?

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+#define FDCAN_TXEVENT_ID_FLAG_ESI BIT31
+/** Extended ID indicator; when set the frame uses 29 bit addressing */
+#define FDCAN_TXEVENT_ID_FLAG_XTD BIT30
+/** Remote transmission requested; when set the frame is a remote frame */
+#define FDCAN_TXEVENT_ID_FLAG_RTR BIT29
+/** @{ Mask and shift for the frame message id */
+#define FDCAN_TXEVENT_ID_29_MASK  0x1FFFFFFF
+#define FDCAN_TXEVENT_ID_29_SHIFT 0
+#define FDCAN_TXEVENT_ID_11_MASK  0x7FF
+#define FDCAN_TXEVENT_ID_11_SHIFT 18
```
karlp: These 29 and 11 masks keep coming up too.

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,1420 @@
...
+#define FDCAN_TXEVENT_MSG_TIMESTAMP_MASK  0xFFFF
+#define FDCAN_TXEVENT_MSG_TIMESTAMP_SHIFT 0
+/** @} */
+
+/* --- Filter Definition ---------------------------------------------------- */
+/** 11 bit address filter element. Note: depending on filter mode, the 2nd
+ * address field has different meanings. */
+struct fdcan_stdfilt {
+	uint32_t flags;
+};
```
karlp: This struct doesn't seem to be doing much?

karlp:
General idea looks fine.  Comment style is a little mixed? sometimes doxygen, sometimes not?  Do you have more of this coming, or just this header for now?

include/libopencm3/stm32/fdcan.h
mattofak: That's fair. Cribbing from the `can.h` file, it looks like I should put this in the `STM32F_defines` group. ST claims that this IP is not specific to the H7...

include/libopencm3/stm32/fdcan.h
mattofak: Happy to put them there; but it looks like `can.h` follows this same pattern of defining the device define in the file and deferring the effort to the `libopencm3/stm32/memorymap.h` header to sort things out.

Any thoughts?

include/libopencm3/stm32/fdcan.h
mattofak: I probably should just do a doxygen and see what happens; but since I don't have a `@file` in the first doxyblock in the file, nor do I have a `@{` are we still logically nesting at this point?

include/libopencm3/stm32/fdcan.h
mattofak: Sort of. I wrote a helper script to pull bit definitions from a copy/paste from the reference manual. But some manual massaging had to be done.

Happy to pull these definitions out though. I wasn't sure if convention was to define everything even if it had no obvious use just to be consistent or not.

include/libopencm3/stm32/fdcan.h
mattofak: I was on the fence about this as well. I'll make them consistent.

include/libopencm3/stm32/fdcan.h
mattofak: done

include/libopencm3/stm32/fdcan.h
mattofak: done

include/libopencm3/stm32/fdcan.h
mattofak: Removed.

include/libopencm3/stm32/fdcan.h
mattofak: I thought about this for a while, and although the 11 bit and 29 bit masks are always going to be the same, I wasn't sure about how to make it clear that they belonged to the `SHIFT`.

include/libopencm3/stm32/fdcan.h
mattofak: Agreed. However, the thought was that it would be clearer to use these as `struct fdcan_stdfilt[]` vs `uint32_t[]`.

mattofak:
Alrighty; instead of cribbing from the existing `can.h` I started following what seemed like the convention of a bunch of other modules. So I moved the new `fdcan.h` into `common/fdcan_common.h` and added the macro magic. I also added a blank object file for fdcan so that I could compile libopencm for the H7 and see all the terrible things I did syntax wise.

include/libopencm3/stm32/fdcan.h
mattofak: Resolved this by not using can.h as a template. Started following other modules which means that this did get moved to the H7 dir and out of the common.

mattofak:
Apologies for the commit/force push noise as I've apparently forgotten how to work with multiple upstreams...

include/libopencm3/stm32/fdcan.h
karlp: yes, please do run doxygen yourself and see that it looks how you expect :)  Probably easiest to open a group around the file, we do that a lot elsewhere.

karlp:
No problems with the pushing, this is your code, push as you see fit! :) and yeah, the existing can.h is probably not the best example of how things are normally done either.   I'd definitely recommend trying this out a bit with your own _use_ of this before we merge it as just definitions, but it's good to see it early.

include/libopencm3/stm32/common/fdcan_common.h
```diff
@@ -0,0 +1,1447 @@
+/** @defgroup fdcan_defines FDCAN defines
+ * @ingroup STM32F_defines
```
karlp: this looks suspect to me, but I've not gone and checked against others.

lib/stm32/common/fdcan_common.c
```diff
@@ -0,0 +1,24 @@
+/**
+ * @brief <b>libopencm3 STM32 FDCAN Controller</b>
+ * LGPL License Terms @ref lgpl_license
+ */
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2019 Brian Viele <vielster@allocor.tech>
```
karlp: This is not you, though I know you're both at allocor.tech.  Also, there's nothing _in_ this file (yet?) and you've not claimed copyright on any of the other files?

mattofak:
@karlp I would appreciate another review. I've used this to build a simple FIFO based CAN-FD thing so it works! Controversial things I've added is a new library to ease bit timing calculations which I rewrote from the linux can-utils project; and some macros to make it easier / less error prone to write bit masking operations.

include/libopencm3/stm32/h7/fdcan.h
```diff
@@ -0,0 +1,37 @@
+/** @defgroup fdcan_defines FDCAN defines
+ * @ingroup STM32F_defines
```
karlp: this doesn't look correct?

lib/stm32/common/fdcan_common.c
karlp: This is still wrong?

karlp:
Do you have any example application code you can share?  I realise it's a big peripheral, but damn, it just looks like so much noise.  I want to pull it locally and see how the docs look too, because there's lots more doxygen wrapping than I'm used to seeing :)

karlp:
Also, there's no "bit timing" library code here? Do you need to git push again? or anything?

include/libopencm3/cm3/common.h
```diff
@@ -111,4 +111,26 @@
 #define BIT30 (1<<30)
 #define BIT31 (1<<31)
 
+/* Generic bit mask manipulation */
+/** Set a \p field inside of \p reg_ with bits_. Assumes that the field name can
```
karlp: In general, there's no poitn in having a "non-doxygen" and then also a doxygen comment.  Just one doxygen comment.

include/libopencm3/cm3/common.h
```diff
@@ -111,4 +111,26 @@
...
+	 (((bits_) & field_##_MASK) << field_##_SHIFT))
+
+/** Obtain a \p field inside of \p reg_. Assumes that the field name can expand
+ * into two definitions _MASK and _SHIFT where the shift is applied before the
+ * mask.
+ * @param reg_   Current value of register
+ * @param field_ Prefix of field to extract from \p reg_
+ * @returns Value of the field inside the register.
+ */
+#define CM3_GET_MASK(reg_, field_) (((reg_) >> field_##_SHIFT) & field_##_MASK)
```
karlp: Put this stuff in a separate commit please, I'm not against it, (I mean, I am, this shit is illegible to debug, but I know it's popular) but let's have it in it's own commit so we can find it easier.  (IN this series is fine)  The USB stack uses constructs like this too, so we extract it out of there as well.

include/libopencm3/stm32/h7/fdcan.h
```diff
@@ -0,0 +1,64 @@
...
+
+/** fdcan_bittiming_const structure for FD rate operation
+ * @ingroup fdcan_defines
+ */
+#define FDCAN_FD_BITTIMING_CONST                                             \
+	{                                                                    \
+		.tseg1_min = 1, .tseg1_max = 32, .tseg2_min = 1,             \
+		.tseg2_max = 16, .sjw_max = 16, .brp_min = 1, .brp_max = 32, \
+		.brp_inc = 1,                                                \
+	}
```
karlp: why can't these just be const structs?  this isn't the 80s, we don't have to do everything with macro soup!

lib/can/can_bit_timing.c
```diff
@@ -0,0 +1,197 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Heavily based on code from LinCAN sources, the H8S2638 project, and the
+ * linux-can projects.
+ * Copyright 2004-2006 Pavel Pisa - DCE FELK CVUT cz
+ * Copyright 2005      Stanislav <pisa@cmp.felk.cvut.cz>
+ * Copyright 2019      Matt Walker <m@allocor.tech>
```
karlp: yes, this will be special.  What sort of licenses do those projects have? Are you sure you can add them to an lgpl project?

include/libopencm3/stm32/h7/fdcan.h
karlp: (remember, data sections and garbage collection will remove unused items....)

karlp:
I'm sorry we didn't get this worked out but some of that extra code always looked a bit special.  I believe this is now superseded by 458250dc so I'm closing it, but if there's functions here that we should still add, please feel free to reopen.

PR #1167, opened 2020-01-21
Update Makefile.include

astekachev:
Hi!
When I try to build Bootloader through the PX4 Toolchain on the command line through the 'make' or "make px4fmuv5_bl" commands, the compiler throws an error "cannot find -lopencm3_stm32f4"
I solved this problem by editing the Make files. I wonder how Bootloader generally compiles if there are a lot of errors in Make files.
In the file ... \ Bootloader \ libopencm3 \ lib \ Make.include, you need to fix the location of the library files and their extension. After attempting to compile, library files of the * .a type are compiled

### DIFF
--- a/lib/Makefile.include
+++ b/lib/Makefile.include
@@ -34,7 +34,8 @@ all: $(SRCLIBDIR)/$(LIBNAME).a
 
 $(SRCLIBDIR)/$(LIBNAME).a: $(OBJS)
 	@printf "  AR      $(LIBNAME).a\n"
-	$(Q)$(AR) $(ARFLAGS) "$@" $(OBJS)
+	#$(Q)$(AR) $(ARFLAGS) "$@" $(OBJS)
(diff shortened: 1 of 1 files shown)


karlp:
I suspect your px4 readme says something like, "build the library first" ?  (or it should?)
There have been one or two other pulls that are trying to sort out the dependencies to make this work better.

Further, I'm not convinced this is doing what you think it is.  $@ is simply the filename of the target, ie, $(LIBNAME).a, so you've not really changed anything.  (See https://www.gnu.org/software/make/manual/html_node/Automatic-Variables.html) Are you sure you didn't do anything else while you were editing this?

(Finally, as a future reference, we don't commit commented out code)

If you can elaborate on _why_ this fixes things, and why the existing method is broken for everyone, please feel free to re-open, or push changes, but for now, I'm closing this.

PR #1168, opened 2020-01-29
swm050: register fix for timer_clock_div

Szalacinski:
Just an incorrect register typo bugfix.


karlp:
Thanks for checking this for me!

PR #1170, opened 2020-02-02
Remove SRCLIBDIR definition from the top Makefile

kapacuk:
It's very difficult to write makefiles which can cope with dirnames or filenames containing whitespaces, especially if you want it to work across different versions of make. For this reason, absolute paths should be avoided if possible, relative paths are much safer.

In this case, the definition of SRCLIBDIR in the top level Makefile causes problems with GNU make v4.3 and later. However, this definition is redundant, as all submakefiles have sensible defaults (either .. or ../..). These defaults will work even if the current dir name contains spaces.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,10 +22,6 @@ PREFIX		?= arm-none-eabi-
 STYLECHECK      := scripts/checkpatch.pl
 STYLECHECKFLAGS := --no-tree -f --terse --mailback
 
-space:=
-space+=
-SRCLIBDIR:= $(subst $(space),\$(space),$(realpath lib))
-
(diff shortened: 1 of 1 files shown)


hve:
To add some more details; GNU make v4.3 added a backward-incompatibility:

> Previously appending using '+=' to an empty variable would result in a value starting with a space.  Now the initial space is only added if the variable already contains some value.  Similarly, appending an empty string does not add a trailing space.

So this definition in the top level Makefile:

`
space:=
space+=
SRCLIBDIR:= $(subst $(space),\$(space),$(realpath lib))
`

Generates a string with a trailing backslash causing make to fail.

`SRCLIBDIR="/home/hve/sources/libopencm3/lib\" `

karlp:
Looks good thanks! always good to delete tricks.

PR #1176, opened 2020-02-18
stm32f3: fix missing reg mask for adc_set_multi_mode

FrancoisFinfe:
Missing defines for ADC_CCR DUAL values have also been added.

### DIFF
--- a/include/libopencm3/stm32/f3/adc.h
+++ b/include/libopencm3/stm32/f3/adc.h
@@ -486,6 +486,43 @@
 #define ADC_CCR_DELAY_SHIFT		8
 
 /* DUAL[4:0]: Dual ADC mode selection */
+/****************************************************************************/
+/** @defgroup adc_multi_mode ADC Multi mode selection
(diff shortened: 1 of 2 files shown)


FrancoisFinfe:
My understanding is that stm32f3 only supports dual mode (ADC1 & 2) or (ADC3 & 4) and not triple mode.
Therefore, adc_set_multi_mode function could also be rename to adc_set_dual_mode (in this same commit).

karlp:
I'd rather keep MULTI so that the same code can be used, but in general this patch is on the right track, definite bug in the implementation to not clear bits first.  We also prefer having unshifted forms of the bit definitions

FrancoisFinfe:
Bit definitions hase been changed to unshifted.

karlp:
Merged thanks! (Made some followup amendments)  _lots_ of this is common with l4, so I've filed that for anyone to work on  https://github.com/libopencm3/libopencm3/issues/1181

PR #1177, opened 2020-02-18
stm32: adc-v2: add adc_{en,dis}able_delayed_conversion_mode

FrancoisFinfe:
To control AUTODELAY feature of the ADC.


karlp:
merged in fb0cac49 thanks!

PR #1179, opened 2020-02-26
pac55xx: implemented CAN module interface for qorvo pac55xx.

stefanik08:


lib/pac55xx/can.c
```diff
@@ -0,0 +1,515 @@
...
+@param[in] db2_mask bool. CAN second data byte mask.
+ */
+void can_filter_single_std(uint32_t canport, uint32_t id, uint32_t id_mask,
+                                uint8_t db1, uint8_t db1_mask,
+                                uint8_t db2, uint8_t db2_mask) {
+  /* set value */
+  uint32_t word = ((id << 21) & CAN_ACR_SINGLE_STD_ID) |
+      ((db1<<8) & CAN_ACR_SINGLE_STD_DB1) |
+      ((db2<<0) & CAN_ACR_SINGLE_STD_DB2);
+  CAN_ACR(canport) = __builtin_bswap32(word);
```
karlp: You're making the words from scratch, why do you need to then swap it?

lib/pac55xx/gpio.c
```diff
@@ -110,10 +110,9 @@ void gpio_set_af(uint32_t gpioport, ccs_muxsel_func_t muxsel, uint16_t gpios) {
 	int ffs = __builtin_ffs(gpios);
 	while (ffs) {
 		const int pin = ffs - 1;
-		const int shift = pin * 4;
 
-		reg &= CCS_MUXSELR_MASK << shift;
-		reg |= muxsel << shift;
+		reg &= ~CCS_MUXSELR_MASK_PIN(pin);
+		reg |= CCS_MUXSELR_VAL(pin, muxsel);
```
karlp: this should be in a standalone commit, explaining why you have had to change it.

karlp:
Could you check the style please? it looks like lots of it is done with 2 spaces instead of a tab character?  the pattern of making words, and then swapping them seems terribly suspect, but I'll look closer later.

lib/pac55xx/can.c
stefanik08: Unfortunately, the user guide is messed up.  The macros I made are per the user guide.  However, the register itself is byte-swapped relative to their definition.  We got confirmation from Qorvo in email indicating this swap.  I could go back and change the macros, etc... if you'd prefer, but then it wouldn't "look" like the user guide.

karlp: And.... they're never going to fix the docs?  This is a pretty absurd situation.  Docs get updated, we're wrong again, no matter what we do. I'm _tempted_ at the moment to say have the definitions be reality, not docs, with comments on the definitions,  then, if the docs get updated to match reality, we have no changes whatsowever, and only people who are looking at the definitions explicitly would ever be confused anyway, if they just use the symbolic names everything "just works" as intended.  I'm not sure what to do here yet. :|

stefanik08: Qorvo told us on email "the documentation needs to be corrected".  However, looking at the documentation below, it looks "clean" before doing the byte-swap.  After doing the byte-swap, the documentation could be confusing.  They may just put verbiage in the doc indicating to do a byte-swap for these registers.  Anyway, I wanted you to take a look at this.  My gut is telling me to do the swap in the macros, and go through the headache of mixing up the bits in the function, in order to align it with documentation.  But I think the `__builtin_bswap32()` is the cleanest way to write the code.

If you give me the go-ahead, I'll make the changes to pull out the `__builtin_bswap32()`.

image

stefanik08:
> Could you check the style please? it looks like lots of it is done with 2 spaces instead of a tab character? the pattern of making words, and then swapping them seems terribly suspect, but I'll look closer later.

I've updated the formatting.  let me know what you think about the bswap32 item.  Changing the macros over will get hairy...

lib/pac55xx/can.c
vielster: Unfortunately this is a bit of a conundrum, but because the registers are endian swapped, and we would not expect the user to call this function with endian swapped filter inputs, it is actually cleaner, and more efficient to just do the single swap. It also will be more error prone, difficult to read, etc. to do individual swaps because 11-bit ID's will not necessarily cross "natural byte boundaries" (i.e. you may have to shift-and-swap-and-shift to get right value).

I will contact the MFG again and see how they intend to fix the documentation. If their fix is just a note that the registers should be accessed as a BE register, then I think the code as is actually meets that.

vielster: FYI, just received response from MFG, their thought is the likely update will just be that these registers should be accessed as BE.

karlp: ok, given that, I guess bswap internally to the library is as rational as things get here then I guess :)

vielster:
Not sure what's up with the checks above...it looks like Travis is showing this passed, but github isn't reporting it properly.

image

lib/pac55xx/can.c
stefanik08: cool. I'll resolve this then

include/libopencm3/pac55xx/can.h
```diff
@@ -0,0 +1,289 @@
...
+#define CAN_SR_TBS      BIT21
+/** DSO: Data Overrun Status */
+#define CAN_SR_DSO      BIT22
+/** RBS: Receive Buffer Status */
+#define CAN_SR_RBS      BIT23
+/**@}*/
+
+/** CAN Interrupt Status Register bit definitions.
+ * 1: interrupt cleared
+ * 0: interrupt triggered
```
karlp, line 137: This didn't look write, and doesn't match the docs either. it should be the other way around

include/libopencm3/pac55xx/can.h
```diff
@@ -0,0 +1,289 @@
...
+ * @ingroup PAC55xx_defines
+ * @author Kevin Stefanik <kevin@allocor.tech>
+ * LGPL License Terms @ref lgpl_license
+ * @date February 13, 2020
+ *
+ * Definitions in this file come from the PAC55XX Family User Guide Rev 1.23
+ * by Active-Semi dated November 19, 2019.
+ *
+ * Note: all memory-mapped writes must be performed using 32-bit registers.
+ * Any 8-bit memory-mapped registers below may only be used to read.
```
karlp, line 14: Where is this documented?  I can't find it in the user guide.

stefanik08: It's not.  But when I tried to do MMIO8(), things blew up.  Since it's a 32-bit bus, it must be 32-bit transactions.  It's probably a limitation of the peripherals themselves rather than the bus.

include/libopencm3/pac55xx/can.h
stefanik08: Looks like I need to fix this documentation.  The bit goes high when the peripheral triggers the interrupt.  However, in order to clear the interrupt (by the user), you have to write a '1' to the bit.

include/libopencm3/pac55xx/can.h
stefanik08: Anyway, it's not a fun rabbit-hole to go down.  So, I got rid of all MMIO8() references.  Should we get rid of this?  Or is there a better place to put documentation like this, @karlp ?

vielster: FYI, this was the case in the vendor's SDK, so I did confirm with their support that all write accesses must be 32-bits. 

It might be interesting to make an `MMIOXXRO` macro that declares the register as `const` to enforce read only access?

karlp: ok then :) yuckkkk.  I was just trying to review things and all the bit offsets are different because it's all documented as just 8 bit registers :)  Let's not block this any longer on this at least.  I'll take it as fact now and keep going

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+	}
+	if ((CAN_ISR_SR_CMR_MR(canport) & CAN_ISR_RI) == 0) {
+		return -1;
+	}
+	if (CAN_RMC(canport) == 0) {
+		return -1;
+	}
+	uint32_t can_buffer = CAN_RXBUF(canport); /* read 32-bit word */
+	uint8_t rx_length = can_buffer & 0x0F;
+	bool is_extended = (can_buffer & BIT7) ? true : false;
```
karlp, line 400: this is redundant, you can just do is_extended = can_buffer & BIT7;

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+					((id & 0x1FE00000u) >> 13) | /* ID 28:21 ==> 15:8 */
+					((id & 0x001FE000u) << 3)  | /* ID 20:13 ==> 23:16 */
+					((id & 0x00001FE0u) << 19);  /* ID 12:5 ==> 31:24 */
+	CAN_TXBUF(canport) = word; /* write first 32-bit word to FIFO */
+
+	word = ((id & 0x1F) << 3); /* ID 4:0 ==> 7:3 */
+	if (length > 0) {
+		word |=			(data[0] << 8) |
+						(data[1] << 16) |
+						(data[2] << 24);
```
karlp, line 342: this is all weirdly aligned?

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief CAN Transmit Extended Frame
+@param[in] canport Unsigned int32. CAN block register base address.
+@param[in] id Unsigned int32. Message ID bits 28:0 used.
+@param[in] rtr bool. Remote Request bit value.
+@param[in] length Unsigned int8. Message payload length.
+@param[in] data Unsigned int8[]. Message payload data.
+@returns int 0 on success, or -1 on failure.
```
karlp, line 320: if you're going to use return codes, you should make them meaningful, not just "-1 for everything"  and that's _if_ you're goign to use error codes

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+*/
+int can_transmit_ext(uint32_t canport, uint32_t id, bool rtr, uint8_t length,
+				const uint8_t *data) {
+	if (length > 8) {
```
karlp, line 324: _generally_ we've not put this sort of error checking into the library code, it's been documented, and the user is meant to do the right thing.  at the very least, this restriction should be in the documentation though!

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+		return -1;
+	}
+
+	/* if TBS is 0, then not ready to transmit */
+	if ((CAN_ISR_SR_CMR_MR(canport) & CAN_SR_TBS) == 0) {
+		return -1;
+	}
+	uint32_t word = (length) | (rtr ? BIT6 : 0) | /* DLC/RTR/FF ==> 7:0 */
```
karlp, line 287: the alignment in this section looks really weird again.

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+					((id & 0x07F8) << 5) | /* ID 10:3 ==> 15:8 */
+					((id & 0x07) << 21)  | /* ID 2:0 ==> 23:21 */
```
karlp, line 289: given the absolute _mountain_ of defines you added, how come these magic numbers are all still here?

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+*/
+void can_enable_irq(uint32_t canport, uint8_t imr) {
+	/* set to 1 (not masked) to enable */
+	CAN_BTR1_BTR0_RMC_IMR(canport) |= (uint32_t)imr;
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief CAN Disable IRQ
+@param[in] canport Unsigned int32. CAN block register base address.
+@param[in] irq Unsigned int8. IRQ bit(s).
```
karlp, line 261: doc parameter doesn't match parameter?

karlp: _please_ run "make doc" and then read the doc/pac55xx/doxygen_pac55xx.log to look for warnings in the files you're adding

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+/** @brief CAN Filter Single Extended Frame w/RTR set
+Note: Acceptance Code Mask Register values of 1 indicate the filter is to
+ignore the bit. However, standard CAN driver APIs use a positive logic for the
+mask. So this function inverts the mask to make this more portable/intuitive.
+
+@param[in] canport Unsigned int32. CAN block register base address.
+@param[in] id Unsigned int32. CAN ID. Only bits 28:0 are used.
+@param[in] id_mask Unsigned int32. CAN ID mask. Only bits 28:0 are used.
+@param[in] rtr bool. Remote Request bit value.
+@param[in] rtr_mask bool. Remote Request mask value.
```
karlp, line 235: same doxygen mismatches here

lib/pac55xx/can.c
```diff
@@ -0,0 +1,470 @@
...
+ * @brief <b>PAC55xxxx CAN Driver</b>
+ * @author @htmlonly &copy; @endhtmlonly 2020 Kevin Stefanik <kevin@allocor.tech>
+ * @date February 13, 2020
+ *
+ * This library supports the CAN module in the PAC55xx SoC from Qorvo.
+ *
+ * Note: Acceptance Code Mask Register values of 1 indicate the filter is to
+ * ignore the bit. However, standard CAN driver APIs use a positive logic for the
+ * mask. The implementations in this file inverts masks as appropriate to
+ * the mask to make this more portable/intuitive.
```
karlp, line 13: I think somewhere in this header intro is where you should document the reasoning behind the bswap32 stuff please, otherwise, it's only tracked in this github pr as far as I can tell.  It's a fairly visible WTF at the moment, even though we've got good reasons for it now :)

lib/pac55xx/Makefile
```diff
@@ -36,6 +36,7 @@ TGT_CFLAGS  += $(STANDARD_FLAGS)
 ARFLAGS		= rcs
 
 OBJS += gpio.o
+OBJS += can.o
```
karlp, line 39: (this is minor, but this should be alphabetically sorted)

lib/pac55xx/can.c
karlp: (sorry about the _mountain_ comment, I'm mixing some of this up with the fdcan PR for stm32, still, these sort of special magic should be defined somewhere.)

lib/pac55xx/can.c
stefanik08: whitespace?  yeah, I'll clean that up.

lib/pac55xx/can.c
stefanik08: no problem.  yeah, I wish there was a way to do a macro like BITS(2,5), which would result in 0x3C (all bits from 2 to 5 set).  There are so many of these magic numbers throughout this file...  There's got to be a better way!

lib/pac55xx/can.c
stefanik08: only place I left return codes was for transmit - if unable to transmit it returns false.

lib/pac55xx/can.c
stefanik08: Added another hill of defines in the header to have a place for the magic numbers.

lib/pac55xx/can.c
stefanik08: Looks like I've resolved the Doxy issues.

include/libopencm3/pac55xx/can.h
karlp: Ok, so.... so I'm 100% clear on this, the docs are 8bit, the register is defined as w1c bits, (ie, you can write zeros all day with no impact, you only write 1s to clear the bit, very standard isr register flags, particularly talking abotu CAN_ISR now) and it actually _doesn't work_ unless you do a 32bit write to the neighbouring registers as well? And there's no errata on their website.  That's _really_ how this part behaves?  The 32bit description on RMW applies to SRAM, and it claims it's handled internally as a performance cost, not something you have to handle yourself.

lib/pac55xx/can.c
```diff
@@ -0,0 +1,460 @@
...
+	}
+	if (length) {
+		*length = rx_length;
+	}
+	uint32_t _id;
+	if (is_extended) {
+		/* Parse extended message ID from RXBUF */
+		_id = ((can_buffer & CAN_BITS_15_8) << 13)   /* ID 28:21 <== 15:8 */
+			| ((can_buffer & CAN_BITS_23_16) >> 3)   /* ID 20:13 <== 23:16 */
+			| ((can_buffer & CAN_BITS_31_24) >> 19); /* ID 12:5 <== 31:24 */
```
karlp, line 408: This is not what I mean by getting rid of the magic numbers :)  you've just replaced one with a different one.

What's wrong with just _id = can_buffer & 0xffffff << 5; can_buffer =can_rxbuf(); _id |= canbuffer & 0xff >> 3

Just save the first byte before h and if that's important.  I don't really understand why you're making 32bit poitners and then trying to do it all byte by byte?

lib/pac55xx/can.c
```diff
@@ -0,0 +1,460 @@
...
+		can_buffer = CAN_RXBUF(canport);
+		_id |= ((can_buffer & CAN_BITS_7_3) >> 3); /* ID 4:0 <== 7:3 */
+
+		/* Parse extended message data from RXBUF */
+		data[0] = can_buffer >> 8;
+		data[1] = can_buffer >> 16;
+		data[2] = can_buffer >> 24;
```
karlp, line 415: shouldn't you be respecting rx_length here? are you just trusting that you're allowed to write whatever you like to the dest buffer?

lib/pac55xx/can.c
```diff
@@ -0,0 +1,460 @@
...
+@param[in] data Unsigned int8[]. Message payload data.
+@returns true if able to transmit, false otherwise.
+*/
+bool can_transmit_ext(uint32_t canport, uint32_t id, bool rtr, uint8_t length,
+			const uint8_t *data) {
+	/* if TBS is 0, then not ready to transmit */
+	if ((CAN_ISR_SR_CMR_MR(canport) & CAN_SR_TBS) == 0) {
+		return false;
+	}
+	uint32_t word = (length & CAN_BITS_3_0)
```
karlp, line 336: this too, BITS_3_0 isn't a helpful definition.  you've just got different magic.  this is "CAN_DLC_MASK"

lib/pac55xx/can.c
```diff
@@ -0,0 +1,460 @@
...
+		| (rtr ? BIT6 : 0) | BIT7 /* DLC/RTR/FF ==> 7:0 */
+		| ((id & CAN_BITS_28_21) >> 13) /* ID 28:21 ==> 15:8 */
+		| ((id & CAN_BITS_20_13) << 3)  /* ID 20:13 ==> 23:16 */
+		| ((id & CAN_BITS_12_5) << 19); /* ID 12:5 ==> 31:24 */
```
karlp, line 340: why not just ```| (id >> 5)```?

lib/pac55xx/can.c
stefanik08: I added to the parameter doc that it must be 'min length 8'

lib/pac55xx/can.c
stefanik08: Let's see.  So that paradigm works for any value that gets shifted up to bit 31.  So, it is what it is for consistency at the moment.  I'll look through the file and see if there's other instances where we could apply this paradigm as well before making the change.

include/libopencm3/pac55xx/can.h
stefanik08: Yes, you have to handle it yourself.  This is an actual excerpt from the sample application that they provided (and they confirmed in email you have to do 32-bit word-aligned writes):
`PAC55XX_CAN->ISR_SR_CMR_MR = (PAC55XX_CAN->ISR_SR_CMR_MR & 0x00FFFFFF) | ISR_RI; // Clear RX interrupt`

lib/pac55xx/can.c
stefanik08: Are you talking about the CAN_RXBUF() reads?  Those HAVE to be 32-bit reads, because they also pop from the 32-bit hardware FIFO.  So, if you do a single 8-bit read, then you end up losing (and can't get back) the other 24-bits from that place in the RX FIFO, because the next read will be from the next 32-bit word.  Is that what you are talking about?  A lot of these reads are "clearing", so I just got rid of all 8-bit accesses to prevent messing up.

lib/pac55xx/can.c
stefanik08: yeah, I guess this would be the only line in the file where we do it like that.  My vote would be to leave it for consistency's sake.

stefanik08:
@karlp Biggest question is what do we want to do about the magic numbers?  Do we want #defines for them all? Or should the "miscellaneous" ones be put back to literals in can.c?

lib/pac55xx/can.c
stefanik08: @karlp Can I mark this as resolved?  Or did you have another thought?

karlp:
this has now merged to master thanks!

PR #1180, opened 2020-02-28
pac55xx: fixing gpio alternate function settings.

stefanik08:
pac55xx: fixing gpio_set_af for pin alternate function settings. register bits were not cleared before setting. refactored to be similar to how the drive strength register is being set.


karlp:
merged in 89e90e0e thanks

PR #1182, opened 2020-03-05
stm32h7: per comments, improved consistency with other rcc impls.

vielster:
Reduced the sea of enums, and renamed config parameters to match other
implementations, cribbing off of the G0 config, as it is closer to the level
of complexity. Updated initialization code to utilize the new values.

Added flash and LDO configuration from RCC init to be more consistent with
STM32 platform initialization.

### DIFF
--- a/include/libopencm3/stm32/h7/rcc.h
+++ b/include/libopencm3/stm32/h7/rcc.h
@@ -415,9 +415,8 @@ LGPL License Terms @ref lgpl_license
 
 #define RCC_HSI_BASE_FREQUENCY              64000000UL
 
-/** Enumerations for clocks in the clock tree to allow user to get the current configuration of the
- *  clocks from the RCC module. These clock sources will each be tracked through the settings.
- */
+/** Enumerations for core system/bus clocks for user/driver/system access to base bus clocks
+  * not directly associated with a peripheral. */
 enum rcc_clock_source {
   RCC_CPUCLK,
   RCC_SYSCLK,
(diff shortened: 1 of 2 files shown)


lib/stm32/common/usart_common_all.c
```diff
@@ -45,11 +45,15 @@ rcc_clock_setup-* functions in RCC).
...
 
+Note: This is not compatible with the complex clock tree implemented by the H7
+series of parts, and is defined separately for those modules.
+
 @param[in] usart unsigned 32 bit. USART block register address base @ref
 usart_reg_base
 @param[in] baud unsigned 32 bit. Baud rate specified in Hz.
 */
 
+#ifndef STM32H7
```
karlp: yeah, we're not goign to do this.  If there's a lot of other common code we do want, fine, then this goes somewhere else, but we're not goign to start putting ifndefs into this file.

karlp:
The RCC stuff looks fine, but can you please separate out the uart stuff again?  we've got to find something else for that.

vielster:
Split the usart stuff into a separate branch. I'll likely put up later...noodling with ideas...order of ideas in my head right now, at least in my preference...please lmk if you prefer any of these.

1. implement `rcc_get_peripheral_clk_freq(uint32_t periph_base)` for f/g/l devices, which presently only supports the USARTs, but can be extended to other peripherals. Note, this would be very useful for things like the SPI buses which have different clock bases per-each. I like this one the best as I think it gives the most flexibility, encourages adding to this implementation, gets away from globals, and is super simple to use in peripheral drivers. The global approach just can't really work for platforms with the clksel features, so this seems to be the most sensible (for now).

1. implement a unique `usart_setbaudrate()` for platform groups, or individual platforms (e.g. `g0/usart.c` OR `usart_common_g.c` and update makefiles appropriately. This works around the clock differences for now.

1. implement `rcc_get_peripheral_clk_freq(uint32_t periph_base)` as a weak symbol, provide global apb variables for h7, use the function if implemented, else use the old way. I don't like perpetuating the global feature in the h7, but this might be the least intrusive.

karlp:
I'd say option 1 is probably best, we know we already have this on l4, f7, g0, and will on g4 as well, so it's kinda inevitable, I just want to make sure we get something ~sane early, and not chop too much into other code, or design it all based on h7 and then try and see how the rest of them need it.  I'd say 2 is your "easy" option to get a uart on h7 right now, and 3 is kinda yuk really, I agree with you.

karlp:
merged in 4953d67a thanks!

PR #1183, opened 2020-03-05
stm32: added rcc_get_peripheral_clk_freq for all stm32 targets.

vielster:


vielster:
FWIW, I fully expect this build to break as is, but wanted to give you a peek @karlp 

Let me know if this will be acceptable. I still need to finish adding f/g platforms for things to build I'll also run some tests on the f3/f4 platforms I have, but I don't have any f0/f1/l0/l1/l4 to test on, do you?

include/libopencm3/stm32/common/rcc_common_all.h
```diff
@@ -67,6 +67,13 @@ bool rcc_is_osc_ready(enum rcc_osc osc);
  */
 void rcc_wait_for_osc_ready(enum rcc_osc osc);
 
+/**
+ * This will return the divisor 1/2/4/8/16/64/128/256/512 which is set as a
+ * 4-bit value, typically used for hpre and other prescalers.
+ * @param div_val  Masked and shifted divider value from register (e.g. RCC_CFGR)
+ */
+uint16_t rcc_prediv_log_skip32_div(uint8_t div_val);
```
karlp: fwiw, personally I hate this name, and I don't think it needs to be public api.  (obviously I'm not math enough to just go "oh yeah, I need this log function" but I don't think any of your log functions for working out bit values are really api level, they'r einternals to make the library developers day easier)

lib/stm32/common/usart_common_all.c
```diff
@@ -37,9 +37,9 @@ Devices can have up to 3 USARTs and 2 UARTs.
 /*---------------------------------------------------------------------------*/
 /** @brief USART Set Baudrate.
 
-The baud rate is computed from the APB high-speed prescaler clock (for
-USART1/6) or the APB low-speed prescaler clock (for other USARTs). These values
-must be correctly set before calling this function (refer to the
+The baud rate is computed from the APB slock associated with the USART/UART
+device. The RCC must be configured for desired frequency before this function
+is called or this may use the incorrect clock for calculations (refer to the
```
karlp: I know you want uart working, but can you stop adding things with uarts when the commit titles are "rcc: add feature xxx" ? :)  add the functionality, then add a single standalone commit for the uart stuff itself.

lib/stm32/h7/Makefile
```diff
@@ -45,7 +45,7 @@ OBJS += pwr.o rcc.o
 OBJS += rcc_common_all.o
 OBJS += spi_common_all.o spi_common_v2.o
 OBJS += timer_common_all.o
-OBJS += usart_common_v2.o usart_common_fifos.o
+OBJS += usart_common_all.o usart_common_v2.o usart_common_fifos.o usart.o
```
karlp: this is out of scope of the rcc add feature commit too

lib/stm32/h7/usart.c
```diff
@@ -0,0 +1,52 @@
...
+
+@param[in] usart unsigned 32 bit. USART block register address base @ref
+usart_reg_base
+@param[in] baud unsigned 32 bit. Baud rate specified in Hz.
+*/
+
+void usart_set_baudrate(uint32_t usart, uint32_t baud)
+{
+	uint32_t clock = rcc_get_peripheral_clk_freq(usart);
+	USART_BRR(usart) = (clock + baud / 2) / baud;
```
karlp: doesn't this go away when you add the section above in common _all?

karlp:
my first impression is yuk, but I'll come around, I know if you want to read all this at run time, there's going to be a bunch of swithces on periph somewhere.  Is there a way to make this stuff not always required?  for many people, they don'ðt need to pull in these big functions.  This was one of the comments  in the old rfc on rcc from years ago, how much code was justififed in recalculating this sort of thing.

include/libopencm3/stm32/common/rcc_common_all.h
vielster: I had it as a static helper in the l0/l4 implementations, and also needed it in the h7, so I figured I'd make it common.  If you don't mind having a few copies of this function running around in separate rcc.c files, I'm fine having the copies.

lib/stm32/common/usart_common_all.c
vielster: I'm sorry...I wanted to have a sample usage to make sure it links and what not...I will pull this out :-)

include/libopencm3/stm32/common/rcc_common_all.h
karlp: eh, ok then, I guess we can let the optimizers deal with it.

lib/stm32/common/usart_common_all.c
karlp: I understand how it happens :)

karlp:
as far as physical testing goes, I have ~everything _except_ h7/g0/g4, and I'll probably get them at some stage. it's more an issue of finding appropriate test code.

vielster:
I will be testing on H7, so just G0 and G4....For now, the only code using this will be the uarts, so that's all that would need to be tested. 

BTW, if there aren't any fancy clock selects, I think it can simplify into something like this...lmk wha tyou think of this...

```c++
uint32_t rcc_get_peripheral_clk_freq(uint32_t periph)
{
	/* Get clock settings based on base address.
	 * 	Note: Clock mappings can be inferred from RCC_APB1ENR and RCC_APB2ENR. */
	if (periph >= TIM2_BASE && periph <= TIM14_BASE) {
			uint8_t ppre1 = (RCC_CFGR >> RCC_CFGR_PPRE1_SHIFT) & RCC_CFGR_PPRE1_MASK;
			return (ppre1 == RCC_CFGR_PPRE_DIV_NONE) ? rcc_apb1_frequency
				: 2 * rcc_apb1_frequency;
	} else if ((periph >= TIM1_BASE && periph <= TIM8_BASE) ||
		         (preiph >= TIM9_BASE && periph <= TIM11_BASE)) {
		uint8_t ppre2 = (RCC_CFGR >> RCC_CFGR_PPRE2_SHIFT) & RCC_CFGR_PPRE2_MASK;
			return (ppre2 == RCC_CFGR_PPRE_DIV_NONE) ? rcc_apb2_frequency
				: 2 * rcc_apb2_frequency;
	} else if (periph >= PERIPH_BASE_APB1 && periph < PERIPH_BASE_APB2) {
		return rcc_apb1_frequency;
	} else if (periph >= PERIPH_BASE_APB2 && periph < PERIPH_BASE_AHB1) {
		return rcc_apb2_frequency;
	} else {
		return 0;
	}
}
```

vielster:
I went with the address range approach as it drastically reduced the codesize...alright...that was fun. Goodnight.

karlp:
I'v ebeen thinking abotu this some more, and you've already got a static struct of clock settings in the h7 rcc code, do we really need functions to go and calculate the periph clocks on demand?  can't the _setters_ update the interal struct?  then the users (like setting the baudrate) stay simple, and if you don't want to use the big clock helpers, you can still just set the struct itself with the values you need, and the periph functions still all work.  I'm suggesting moving the present three global bus speeds to a struct/(overloaded arrray might be even better, to allow aliasing) of _peripheral_ speeds.  It's a bigger set of constants, but it's not as big as _deriving_ clocks on fetch.

You could do something like 

uint32_t periph_speeds[n] and index it via PERIPH_CLOCK_TIMER6, or something, and we can use aliases to have multiple peripherals all use the same clock speed when it's shared, so we don't have 100 clocks.  I'm not sure this would be ~much bigger (if at all) than trying to keep some clock setup structs, and the code of reading things back out and calculating clocks based on dividers can all be dropped, you just directly use the value that you know at setting time.
timer3 =

vielster:
I don't think that approach will solve this. It still relies on someway to get from a `periph` to a `clock_index` which means an if/else or switch/case. It is more flexible, because you have a bigger table, but it is still static, ie. you have to be able to make assumptions about things that are true on all platforms. Using global "constants" in this way is just generally bad because it provides no abstraction, so if assumptions change, things break.

There are lots of different ways to slice this up _functionally_ which could break it into smaller chunks, and make codesize smaller. Some of these are

1. for any function dependent on a clock, you have to have a per-platform implementation of it
1. any function dependent on a clock takes the clock as an input
1. have a clock getter for each peripheral type that needs one (e.g. `uint32_t rcc_get_usart_clk_freq(int index)` and `uint32_t rcc_get_spi_clk_freq(int index)`

It's clear, you're not a fan of this _big_ logic approach, and that's fine, but I think this needs to be done functionally in order to allow for platforms that don't just use `apb1` for all usarts except USART1 and USART6. I went with this larger scoped approach, partially because you suggested to do it like the `rcc_set_peipheral_clk_sel()` function, but think it completely fine to break this up into smaller pieces. For example, I could just add this to each platform

```c++
uint32_t rcc_get_usart_clk_freq(uint32_t preiph) {
  if (periph == USART1 || periph == USART6) {
    return rcc_apb2_frequency;
  } else {
    return rcc_apb1_frequency;
  }
}
```

This is essentially a no-op from the current implementation. but would allow platform abstraction.

I've already gone through the trouble of writing a function to handle the clksel aspects on these platforms, so it'd be nice to include that, but honestly, I'm fine just doing the above to keep things simple, and is in the right direction.

include/libopencm3/stm32/common/rcc_common_all.h
vielster: renamed `rcc_get_div_from_hpre()` since that's what it's used for :-)

vielster:
Hey @karlp looking for some guidance to move this forward. Summary above is 

* approach in this MR is fairly complete, but is complicated and has codesize implications.
* making a table of clocks work has similar issues to the current approach, may not give enough flexibility, may create other compatibility issues, and is just a bit fragile  
* clock dependent function entrypoints which accept a clock as an input could work, but may cause some use case confusion (e.g. `usart_set_baudrate()` may not work and user would need to know to call `usart_set_baudrate_with_clk()` for their platform vs. `usart_set_baudrate()` just working). 
* Simplifying the RCC function calls per peripheral (e.g. `rcc_get_usart_clk_freq(uint32_t usart)`) has smaller scope of changes, smaller codesize, and allows abstraction of only peripheral clocks that need it.

karlp:
Sorry, I don't have a lot of time for this project at the best of times, and it's unfortunately even more limited right now.  That's about the best I can give you right now I'm afraid.  I know a lot of people are isolated and have time on their hands, but I've got a lot _less_ right now :)

JerryZheng89:
so, how about let others do something for this project?

karlp:
Jerry, reviewing and commenting would be a really great start.  This is a major change in direction for a core component of clocking, and **no-one** else has commented.  What sort of things did you have in mind for "letting others do things" ?

vielster:
FYI, closing this for preference to #1191 as it lessens the scope of changes, codesize, etc. Can re-open if this is the preferred approach.

PR #1184, opened 2020-03-10
stm32h7: fix inverted VOS settings for Vcore.

vielster:
The scaling values for Vcore were flipped, which caused the MCU to not run reliably at higher frequencies. With this fix, the system is now running happily at the full 480MHz rated frequency.


karlp:
merged in 89074d6a thanks

PR #1185, opened 2020-03-16
added i2c transfer function with a MAX retry counter

W-M-D:
I'm not sure if I should add the max retry variable as an arguemnt. This seems to work for the chips I have tried.


karlp:
Is this just the existing transfer function with retry wrapping around the write portion? (but not the read?)  This seems pretty out of scope IMO as a first instinct at least.  What sort of hardware are you working with that you need _200_ retries on write, but none on the read?

PR #1191, opened 2020-03-23
stm32: added peripheral clock get helpers for all stm32 platforms.

vielster:
Allows for abstraction for code that's dependent on knowing the source clock
for a peripheral. Implemented a few core peripherals that tend to have clock
tree differences between platforms (USART, timers, I2C, SPI).

### DIFF
--- a/include/libopencm3/stm32/common/rcc_common_all.h
+++ b/include/libopencm3/stm32/common/rcc_common_all.h
@@ -67,6 +67,13 @@ bool rcc_is_osc_ready(enum rcc_osc osc);
  */
 void rcc_wait_for_osc_ready(enum rcc_osc osc);
 
+/**
+ * This will return the divisor 1/2/4/8/16/64/128/256/512 which is set as a
+ * 4-bit value, typically used for hpre and other prescalers.
+ * @param div_val  Masked and shifted divider value from register (e.g. RCC_CFGR)
+ */
+uint16_t rcc_get_div_from_hpre(uint8_t div_val);
+
 END_DECLS
 /**@}*/
 
--- a/include/libopencm3/stm32/f0/rcc.h
+++ b/include/libopencm3/stm32/f0/rcc.h
@@ -149,6 +149,7 @@ Control</b>
 
 #define RCC_CFGR_PPRE_SHIFT			8
 #define RCC_CFGR_PPRE				(7 << RCC_CFGR_PPRE_SHIFT)
+#define RCC_CFGR_PPRE_MASK			0x7
 /** @defgroup rcc_cfgr_apb1pre RCC_CFGR APB prescale Factors
 @{*/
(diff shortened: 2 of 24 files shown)


vielster:
@karlp this is a new version of the rcc merge request but has each of the peripherals broken out individually. This allows for much smaller contained blocks for these helpers. I actually prefer this to the other more massive `rcc_get_peripheral_clk_freq()` which tended to be massive.

vielster:
@karlp bump (in case this email got lost in the shuffle too since this may have been around the same time)

nicovince:
I could really use this PR to be merged into master. 

@karlp @vielster  How can I help get this merged ? Tested ? 

I'm working on libopencm3 support for stm32mp1 device with an stm32mp1-DK2 board and the clock frequency getter could really be useful :)

Thanks.

vielster:
@nicovince that port should be fun! I'm guessing it has a pretty substantial clock tree, so something similar to what's used for the H7 may be a good starting point (although I'm guessing you'll need to add a detailed implementation for that platform). 

@karlp This PR should not have any negative impact on other platforms, as the calls are not yet being _used_ in this MR. In that regard, significant testing shouldn't be needed. The place where it will get more critical is changes to the peripherals (e.g. usart) to use these calls instead of the global clock values, at which point we'll need to test on as many platforms as possible before we call it good.

nicovince:
@vielster Yes it's quite a big clock tree, and I took inspiration from the H7 implementation. I think I may have to make a rcc_common_h7mp1.c or something similar. Not sure of the details yet.
If you want to have a look at what I have done yet, have a look at my [fork](https://github.com/nicovince/libopencm3/commits/stm32mp1) and feel free to comment.

karlp:
I'm still concerned abotu what the right user api is, I (today! hah!) kinda lean towards if an end user application needs this, the single function ```rcc_get_peripheral_clk_freq(uint32_t periph)``` is arguably better than the multiple ```rcc_get_[usart|timer|i2c|spi]_clk_freq(uint32_t periph)``` functions, but we're pulling in heaps code no matter what.  And if we're mostly expecting _library_ developers to need these functions, then the more granular we can make them, the better as far as code size goes.  So, I'm pulling in _this_ version, as it's the one we felt best about at the time, and we certainly don't need to have waste more time on another conversation about it!  (If anyone's had second thoughts abotu this themselves, by all means let me know!)

karlp:
for followup for the bored, can't the 
```
uint32_t rcc_get_usart_clk_freq(uint32_t usart);
uint32_t rcc_get_timer_clk_freq(uint32_t timer);
uint32_t rcc_get_i2c_clk_freq(uint32_t i2c);
uint32_t rcc_get_spi_clk_freq(uint32_t spi);
```
go into ```rcc_common_all.h``` instead of into every single .h file?

vielster:
If there are peripherals guaranteed to be present on all platforms, I would agree they can go in `rcc_common_all.h`

karlp:
this version has now landed in master thanks!

PR #1196, opened 2020-03-30
pac55xx: adding memctl and clock/pll config

stefanik08:
pac55xx: adding memctl for flash/sram access, and clock/pll configuration functions.

### DIFF
--- a/include/libopencm3/pac55xx/ccs.h
+++ b/include/libopencm3/pac55xx/ccs.h
@@ -4,6 +4,7 @@
  *  @defgroup system_defines Clock Config and System Defines
  *  @ingroup PAC55xx_defines
  *  @author Brian Viele <vielster@allocor.tech>
+ *  @author Kevin Stefanik <kevin@allocor.tech>
  *  LGPL License Terms @ref lgpl_license
  *  @date 1 Dec 2019
  *
@@ -26,18 +27,95 @@
  * You should have received a copy of the GNU Lesser General Public License
  * along with this library.  If not, see <http://www.gnu.org/licenses/>.
  */
-#ifndef INCLUDE_LIBOPENCM3_PAC55XX_CCS_H_
-#define INCLUDE_LIBOPENCM3_PAC55XX_CCS_H_
+#ifndef LIBOPENCM3_PAC55XX_CCS_H_
+#define LIBOPENCM3_PAC55XX_CCS_H_
 
 #include <libopencm3/cm3/common.h>
+#include <libopencm3/cm3/memorymap.h>
+#include <libopencm3/pac55xx/memorymap.h>
+#include <libopencm3/pac55xx/memctl.h>
 
 /**@{*/
 
-/** Clock Control Registers
- * @defgroup clock_config_regs Clock Config Registers.
- * @{*/
+/** @defgroup ccs_frequencies CCS Frequencies
+@{*/
+/** Ring Oscillator Frequency */
+#define CCS_ROSC_FREQ       (16000000U)
+/** Internally generated and trimmed 4MHz clock */
+#define CCS_CLKREF_FREQ     ( 4000000U)
+/** Maximum external clock frequency */
+#define CCS_EXTCLK_MAX_FREQ (20000000U)
+/**@}*/
+
+/** @defgroup ccs_ctl_reg Clock Control Register
+@{*/
 #define CCSCTL MMIO32(SCC_BASE)
+#define CCS_CTL_FRCLKMUXSEL_MASK (0x03)
+#define CCS_CTL_FRCLKMUXSEL(sel) ((sel) & CCS_CTL_FRCLKMUXSEL_MASK)
+#define CCS_CTL_FRCLKMUXSEL_ROSC    (0)
+#define CCS_CTL_FRCLKMUXSEL_CLKREF  (1)
+#define CCS_CTL_FRCLKMUXSEL_EXTCLK  (3)
+#define CCS_CTL_ROSCEN BIT2
+#define CCS_CTL_SCLKMUXSEL BIT4
+#define CCS_CTL_SCLKMUXSEL_FRCLK    (0)
+#define CCS_CTL_SCLKMUXSEL_PLLCLK   (1)
+#define CCS_CTL_CLKFAILEN BIT5
+#define CCS_CTL_CLKFAILMUXSEL BIT6
+#define CCS_CTL_CLKFAILIF BIT7
+#define CCS_CTL_LDOEN BIT8
+#define CCS_CTL_SWRESET BIT11
+#define CCS_CTL_PCLKEN BIT12
+#define CCS_CTL_ACLKEN BIT13
+#define CCS_CTL_ADCCLKEN BIT14
+#define CCS_CTL_STCLKSLPEN BIT15
+#define CCS_CTL_PCLKDIV_MASK (0x07)
+#define CCS_CTL_PCLKDIV_SHIFT (16)
+/* Supported PCLK divisors: 1-8 */
+#define CCS_CTL_PCLKDIV(div) (((div-1) & CCS_CTL_PCLKDIV_MASK) << CCS_CTL_PCLKDIV_SHIFT)
+#define CCS_CTL_ACLKDIV_MASK (0x07)
+#define CCS_CTL_ACLKDIV_SHIFT (20)
+/* Supported ACLK divisors: 1-8 */
+#define CCS_CTL_ACLKDIV(div) (((div-1) & CCS_CTL_ACLKDIV_MASK) << CCS_CTL_ACLKDIV_SHIFT)
+#define CCS_CTL_HCLKDIV_MASK (0x07)
+#define CCS_CTL_HCLKDIV_SHIFT (24)
+/* Supported HCLK divisors: 1-8 */
+#define CCS_CTL_HCLKDIV(div) (((div-1) & CCS_CTL_HCLKDIV_MASK) << CCS_CTL_HCLKDIV_SHIFT)
+#define CCS_CTL_USAMODE BIT28
+#define CCS_CTL_USBMODE BIT29
+#define CCS_CTL_USCMODE BIT30
+#define CCS_CTL_USDMODE BIT31
+/**@}*/
+
+/** @defgroup ccs_pllctl_reg CCS PLL Control Register
+@{*/
 #define CCSPLLCTL MMIO32(SCC_BASE + 0x04)
+/** PLL Enable */
+#define CCS_PLLCTL_PLLEN BIT0
+/** PLL Bypass */
+#define CCS_PLLCTL_PLLBP BIT1
+#define CCS_PLLCTL_PLLOUTDIV_MASK (0x03)
+#define CCS_PLLCTL_PLLOUTDIV_SHIFT (2)
+/** PLL Output Divisor */
+#define CCS_PLLCTL_PLLOUTDIV(div) (((div) & CCS_PLLCTL_PLLOUTDIV_MASK) << CCS_PLLCTL_PLLOUTDIV_SHIFT)
+#define CCS_PLLCTL_PLLOUTDIV1     (0)
+#define CCS_PLLCTL_PLLOUTDIV2     (1)
+#define CCS_PLLCTL_PLLOUTDIV4     (2)
+#define CCS_PLLCTL_PLLOUTDIV8     (3)
+#define CCS_PLLCTL_PLLINDIV_MASK (0x0F)
+#define CCS_PLLCTL_PLLINDIV_SHIFT (4)
+/** PLL Input Divisor */
+#define CCS_PLLCTL_PLLINDIV(div) (((div) & CCS_PLLCTL_PLLINDIV_MASK) << CCS_PLLCTL_PLLINDIV_SHIFT)
+#define CCS_PLLCTL_PLLFBDIV_MASK (0x3FFF)
+#define CCS_PLLCTL_PLLFBDIV_SHIFT (8)
+/** PLL Feedback Divisor */
+#define CCS_PLLCTL_PLLFBDIV(div) (((div) & CCS_PLLCTL_PLLFBDIV_MASK) << CCS_PLLCTL_PLLFBDIV_SHIFT)
+/** PLL Lock */
+#define CCS_PLLCTL_PLLLOCK BIT24
+/**@}*/
+
+/** @defgroup ccs_rosctrim Ring Oscillator Trim Control Register
+@{*/
+#define CCSROSCTRIM_MASK (0x7F)
 #define CCSROSCTRIM MMIO32(SCC_BASE + 0x08)
 /**@}*/
 
@@ -65,6 +143,9 @@
 #define CCS_PFMUXSELR CCS_MUXSELR(CCS_PORTF)
 #define CCS_PGMUXSELR CCS_MUXSELR(CCS_PORTG)
 #define CCS_MUXSELR_MASK 0x7
+#define CCS_MUXSELR_MASK_PIN(pin) (CCS_MUXSELR_MASK << ((pin) * 4))
+#define CCS_MUXSELR_VAL(pin, muxsel) (((muxsel) & CCS_MUXSELR_MASK) << ((pin) * 4))
+
 /* Enum type for port function setting for type specificity. */
 typedef enum {
 	CCS_MUXSEL_GPIO = 0,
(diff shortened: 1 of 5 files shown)


lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,271 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <libopencm3/pac55xx/ccs.h>
+#include <libopencm3/pac55xx/memorymap.h>
+#include <libopencm3/pac55xx/memctl.h>
+#include <libopencm3/cm3/assert.h>
+
+/** Global clock tree structure that is configured by ccs_configure_clocks */
```
vielster: Looks like structure has been removed, but comment doesn't match.

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,271 @@
...
+uint32_t ccs_extclk_frequency = CCS_ROSC_FREQ;
```
vielster: Consider making these static (global write access to these seems bad)

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,271 @@
...
+		case WWDT_BASE:
+			return (select == 0) ? ccs_frclk_frequency : CCS_ROSC_FREQ;
+		case RTC_BASE:
+			return ccs_frclk_frequency;
+		case CRC_BASE:  /* fall through */
+		case SYS_TICK_BASE:
+			return ccs_hclk_frequency;
+		default:
+			cm3_assert_not_reached();
+			break;
```
vielster: nit: break and return 0 below are not necessary

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,271 @@
...
+			ccs_frclk_frequency = CCS_ROSC_FREQ;
+			break;
+		case CCS_CTL_FRCLKMUXSEL_CLKREF:
+			ccs_frclkmux_select(CCS_CTL_FRCLKMUXSEL_CLKREF);
+			ccs_frclk_frequency = CCS_CLKREF_FREQ;
+			break;
+		case CCS_CTL_FRCLKMUXSEL_EXTCLK:
+			if (config->extclk_frequency > CCS_EXTCLK_MAX_FREQ
+					|| config->extclk_frequency == 0) {
+				FLASHLOCK = 0;
```
vielster: Suggest `#define FLASHLOCK_LOCKED` or similar instead of magic number 0...here and below.

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,271 @@
...
+}
+
+void ccs_reset_clocks(void) {
+	CCSCTL = CCS_CTL_LDOEN | CCS_CTL_ROSCEN |
+			CCS_CTL_PCLKEN | CCS_CTL_ACLKEN |
+			CCS_CTL_ADCCLKEN | CCS_CTL_STCLKSLPEN;
+	CCSPLLCTL = 0;
+}
+
+int ccs_configure_clocks(const struct ccs_clk_config *config) {
```
vielster: returning errors is atypical in libopencm clock setup code. Consider adding `cm3_assert` or other mechanism rather than error returns for consistency.

include/libopencm3/pac55xx/ccs.h
```diff
@@ -26,18 +27,100 @@
...
 
 /**@{*/
 
-/** Clock Control Registers
- * @defgroup clock_config_regs Clock Config Registers.
- * @{*/
+/** @defgroup ccs_frequencies CCS Frequencies
+@{*/
+/** Ring Oscillator Frequency */
+#define CCS_ROSC_FREQ       (16000000ul)
```
vielster: opencm "standard" would seem to be caps literal type definitions, so should be `16000000U`  (`long` is redundant as this will _always_ be a 32-bit machine, so would drop to match with more "standard" formatting)

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,271 @@
...
+}
+void css_pll_config_enable(uint32_t indiv, uint32_t fbdiv, uint32_t outdiv) {
+	ccs_pll_disable();
+	ccs_pll_set_fbdiv(fbdiv);
+	ccs_pll_set_outdiv(outdiv);
+	ccs_pll_set_indiv(indiv);
+	ccs_pll_enable();
+	while (!ccs_pll_locked()) ; /* Wait for PLL lock ~500us */
+}
+uint32_t ccs_get_peripheral_clk_freq(uint32_t periph, uint32_t select) {
```
vielster: A couple of comments...

1. Depending on outcome of #1191 we may want to split this up to be consistent. 
1. I would suggest that the code here determine the `select` value for the peripheral and not rely on it being passed in so that the user doesn't need to know the details of how the clocks are/have been configured.

lib/pac55xx/ccs.c
stefanik08: comment removed

lib/pac55xx/ccs.c
stefanik08: they are now static

lib/pac55xx/ccs.c
stefanik08: removed break/return

lib/pac55xx/ccs.c
stefanik08: instead of 0, using FLASHLOCK_CLEAR

include/libopencm3/pac55xx/ccs.h
stefanik08: changed suffix to cap U

lib/pac55xx/ccs.c
stefanik08: removed return value.  errors in config values will now go into `cm3_assert_not_reached()` loop.

lib/pac55xx/ccs.c
vielster: For now, resolving as okay as is, and both can be fixed pretty easily later, especially since limited use of this at this point.

vielster:
Not sure if my approval matters much :-P  @karlp is most common person to actually merge changes.

include/libopencm3/pac55xx/ccs.h
```diff
@@ -133,6 +218,151 @@ typedef enum {
...
+ * This library supports the CCS module in the PAC55xx SoC from Qorvo.
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+
+/**@{*/
+
+/**
+ * Select the source for FRCLK.
+ * @param[in] sel  One of ccs_ctl_frclkmuxsel_t.
```
karlp: if that's a type, why isn't it the argument type? if it's not a type, why does it end in _t?  (does the docs autolink that or do you need a @ref?)

stefanik08: looks like a remnant that needs removal.  I had a few things as a `typedef enum` that were changed to standard macros instead.

include/libopencm3/pac55xx/memctl.h
```diff
@@ -0,0 +1,114 @@
...
+#define MEMCTL_MCLKDIV7     (6)
+#define MEMCTL_MCLKDIV8     (7)
+#define MEMCTL_MCLKDIV9     (8)
+#define MEMCTL_MCLKDIV10    (9)
+#define MEMCTL_MCLKDIV11    (10)
+#define MEMCTL_MCLKDIV12    (11)
+#define MEMCTL_MCLKDIV13    (12)
+#define MEMCTL_MCLKDIV14    (13)
+#define MEMCTL_MCLKDIV15    (14)
+#define MEMCTL_MCLKDIV16    (15)
```
karlp: this looks like it would be easier and simpler and clearere for the users to be just
```
#define MEMCTL_MCLKDIV(d) (((d-1) & MASK << SHIFT)
```
This lets users jsut do ```MEMCTL |= MEMCTL_MCLKDIV(4)``` instead of ```MEMCTL |= MEMCTL_MCLKDIV(MEMCTL_MCLKDIV4)``` and that seems simpler and easier to read and work with.

include/libopencm3/pac55xx/memctl.h
```diff
@@ -0,0 +1,114 @@
...
+#define FLASHLOCK_CLEAR                 (0)
+#define FLASHLOCK_ALLOW_FLASH_WRITE     (0x43DF140A)
+#define FLASHLOCK_ALLOW_MEMCTL_WRITE    (0xD513B490)
+#define FLASHLOCK_ALLOW_INFO2_SWDFUSE   (0x79B4F762)
+/**@}*/
+
+/** Flash Page Address Register */
+#define FLASHPAGE       MMIO32(MEMCTL_BASE + 0x000C)
+/** SWD Unlock Register */
+#define SWDUNLOCK       MMIO32(MEMCTL_BASE + 0x0010)
```
karlp: We _normally_ try and prefix all registers with the peripheral to avoid namespace pollution.  This (might, for instance) lead to
```
MEM_MEMCTL
MEM_FLASHPAGE
MEM_SWDUNLOCK
```

Or similar

lib/pac55xx/Makefile
```diff
@@ -36,6 +36,7 @@ TGT_CFLAGS  += $(STANDARD_FLAGS)
 ARFLAGS		= rcs
 
 OBJS += gpio.o
+OBJS += ccs.o
```
karlp: _try_ and keep this alphabetised :)

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,264 @@
...
+	CCSCTL = CCS_CTL_LDOEN | CCS_CTL_ROSCEN |
+			CCS_CTL_PCLKEN | CCS_CTL_ACLKEN |
+			CCS_CTL_ADCCLKEN | CCS_CTL_STCLKSLPEN;
+	CCSPLLCTL = 0;
+}
+
+void ccs_configure_clocks(const struct ccs_clk_config *config) {
+	FLASHLOCK = FLASHLOCK_ALLOW_MEMCTL_WRITE;
+
+	// set safe defaults:
```
karlp: only C style comments please /*  (applies throughout)

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,264 @@
...
+	ccs_reset_clocks();
+	ccs_frclkmux_select(CCS_CTL_FRCLKMUXSEL_ROSC);
+	ccs_sclkmux_select_frclk();
+	MEMCTL &= ~MEMCTL_MCLKSEL; // select ROSCCLK
```
karlp: are you not adding functions for these memctrl stuff?  You'd get to memctl_flash_enable(config->mem_cachedis)

include/libopencm3/pac55xx/ccs.h
```diff
@@ -133,6 +218,151 @@ typedef enum {
...
+	uint32_t pll_indiv;                 /**< PLL Input Divider 1-15 */
+	uint32_t pll_fbdiv;                 /**< PLL Feedback Divider 4-16383 */
+	uint32_t pll_outdiv;                /**< PLL Output Divider */
+	uint32_t hclkdiv;                   /**< Divisor from SCLK to HCLK */
+	uint32_t aclkdiv;                   /**< Divisor from SCLK to ACLK */
+	uint32_t pclkdiv;                   /**< Divisor from HCLK to PCLK */
+	uint32_t mem_wstate;                /**< Number of Flash Read wait states */
+	uint32_t mem_mclkdiv;               /**< Divisor from HCLK to MCLK */
+	bool mem_mclksel;                   /**< false: ROSCLK, true: HCLK/MCLK */
+	bool mem_cachedis;                  /**< false: enable cache, true: disable cache */
```
karlp: I would generally prefer to see positive logic without strong reasons.  (and "default init value is false" isn't really good enough in my mind)

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,264 @@
+/**
+ * ccs.c
```
karlp: you don't really need this?

lib/pac55xx/ccs.c
```diff
@@ -0,0 +1,264 @@
...
+}
+void ccs_pll_set_indiv(uint32_t div) {
+	if (div <= 15 && div >= 1) {
+		CCSPLLCTL = (CCSPLLCTL & ~CCS_PLLCTL_PLLINDIV(CCS_PLLCTL_PLLINDIV_MASK)) | CCS_PLLCTL_PLLINDIV(div);
+	}
+}
+void ccs_pll_set_fbdiv(uint32_t div) {
+	if (div <= 16383 && div >= 4) {
+		CCSPLLCTL = (CCSPLLCTL & ~CCS_PLLCTL_PLLFBDIV(CCS_PLLCTL_PLLFBDIV_MASK)) | CCS_PLLCTL_PLLFBDIV(div);
+	}
```
karlp: if you're going to bother error checking, it would seem reasonable to provide a return code, but that's very rarely done anywhere else either so *shrugs*

lib/pac55xx/ccs.c
stefanik08: I didn't feel the 2-4 needed functions necessary were worthy of another source file.  could these function defs go into ccs.c? eh, another source file doesn't hurt either.

include/libopencm3/pac55xx/ccs.h
stefanik08: yeah, this follows the documentation of the register (below)
image

lib/pac55xx/ccs.c
stefanik08: The `select` is used since several of the peripherals choose their own clock within their register set.  If we want this to operate as described in (2) above, then ccs.c will have dependencies on each peripheral's register definition, or at least a dumbed-down version of it.  Not saying it's impossible, just wouldn't look very pretty to have each of those register definitions included here.

vielster: Okay, that makes sense if it's controlled in the local regs for the peripheral, and not ccs registers. Seems like a bit of splitting the baby, as maybe at the point that they're telling the clock module what their setting is, they should just do the logic internally and simply need a getter for aclk/pclk

include/libopencm3/pac55xx/ccs.h
karlp: and for bitfield definitions, that's more relevant, for user facing helper structures, very much less so.  I see where you're going, but the aims of helpers like this is helping, not just mirroring datasheets.

lib/pac55xx/ccs.c
stefanik08: yeah, somewhat unrelated: found that there isn't a debug mechanism to unbrick a chip if you get the clock config wrong.  so we put in some software hooks outside of libopencm3 to avoid bricking.  regardless, I just wanted to prevent writing an invalid value to a register and let the user debug why it isn't working. they'd figure it out quickly with GDB

include/libopencm3/pac55xx/memctl.h
stefanik08: Everything is now prefixed with `MEMCTL_`

lib/pac55xx/Makefile
stefanik08: done

lib/pac55xx/ccs.c
stefanik08: memctl.c added

include/libopencm3/pac55xx/ccs.h
stefanik08: changed as recommended

lib/pac55xx/ccs.c
stefanik08: I put in `cm3_assert_not_reached();` if the parameters are bad - to be similar to the rest of this source file

include/libopencm3/pac55xx/memctl.h
stefanik08: done

lib/pac55xx/ccs.c
stefanik08: done

include/libopencm3/pac55xx/ccs.h
```diff
@@ -26,18 +27,95 @@
...
+@{*/
 #define CCSPLLCTL MMIO32(SCC_BASE + 0x04)
+/** PLL Enable */
+#define CCS_PLLCTL_PLLEN BIT0
+/** PLL Bypass */
+#define CCS_PLLCTL_PLLBP BIT1
+#define CCS_PLLCTL_PLLOUTDIV_MASK (0x03)
+#define CCS_PLLCTL_PLLOUTDIV_SHIFT (2)
+/** PLL Output Divisor */
+#define CCS_PLLCTL_PLLOUTDIV(div) (((div) & CCS_PLLCTL_PLLOUTDIV_MASK) << CCS_PLLCTL_PLLOUTDIV_SHIFT)
```
stefanik08, line 99: I was thinking about it, and one way to make this more consistent like the rest of the divisors below (since this is power of 2 instead of incremental), we could do something like this, so the user would only be able to input 1,2,4, or 8, but the register value would still turn out alright, and we'd delete lines 100-104 below.
```
#define CCS_PLLCTL_PLLOUTDIV(div) (((((div + 1) * 2) / 5) & CCS_PLLCTL_PLLOUTDIV_MASK) << CCS_PLLCTL_PLLOUTDIV_SHIFT)
```

stefanik08: @karlp Would my suggestion above be consistent with libopencm3 standards?

include/libopencm3/pac55xx/ccs.h
```diff
@@ -65,6 +143,8 @@
 #define CCS_PFMUXSELR CCS_MUXSELR(CCS_PORTF)
 #define CCS_PGMUXSELR CCS_MUXSELR(CCS_PORTG)
 #define CCS_MUXSELR_MASK 0x7
+#define CCS_MUXSELR_MASK_PIN(pin) (CCS_MUXSELR_MASK << ((pin)) * 4)
```
karlp, line 146: I'm just about to merge this, and it complains on this as a conflict with 89e90e0e5da5258432f4e16fdcf5911d3f8a22b2
I _believe_ from inspection that _that_ commit is correct, and this has the bracketing about pin wrong, but please let me know if I've got that wrong! :)

stefanik08: Yes, you are correct. the other one is right, while this one is wrong.

lib/pac55xx/ccs.c
stefanik08: @vielster Do we want to keep this as-is, or look at restructuring?

vielster: I think this probably _should_ be restructured to match the STM32 implementation when it gets merged per #1191 but this platform is separated enough from normal use in this tree that I don't think it's an issue to leave as-is.  It will increase codesize a small amount.

karlp:
this landed in 253a09193 thanks!

PR #1197, opened 2020-04-02
pac55xx: add usart definitions and basic support code.

stefanik08:


include/libopencm3/pac55xx/usart.h
```diff
@@ -0,0 +1,225 @@
...
+/** Enable the RX Buffer Register Interrupt */
+#define USART_IER_RBRIE         BIT0
+/**@}*/
+
+/** @defgroup usart_iir_bits Interrupt ID Register bits
+@{*/
+/** This bit is active low to indicate an interrupt is pending */
+#define USART_IIR_INTSTATUS     BIT0
+/** TX Holding Register Empty */
+#define USART_IIR_TXEMPTY       (0b0010)
```
karlp, line 79: please use hex here.  binary literals don't arrive until c14, and are a gcc extension only until then.

include/libopencm3/pac55xx/usart.h
```diff
@@ -0,0 +1,225 @@
...
+BEGIN_DECLS
+
+uint32_t usart_set_baudrate(uint32_t usart, uint32_t baud);
+void usart_configure_lcr(uint32_t usart, uint8_t data_bits, uint8_t stop_bits,
+				uint8_t parity);
+void usart_break_enable(uint32_t usart);
+void usart_break_disable(uint32_t usart);
+void usart_enhanced_enable(uint32_t usart);
+void usart_enhanced_disable(uint32_t usart);
+bool usart_configure_fcr(uint32_t usart, uint8_t tx_depth, uint8_t rx_depth);
```
karlp, line 209: would usart_set_fifo_depth() not be easier to understand?

lib/pac55xx/usart.c
```diff
@@ -0,0 +1,207 @@
...
+@param[in] rx_depth unsigned 8 bit. One of USART_FIFO_TRIG_1/2/4/14CHAR.
+@return true if the FIFOs are enabled and the set was successful. false if FIFOs are disabled.
+*/
+bool usart_configure_fcr(uint32_t usart, uint8_t tx_depth, uint8_t rx_depth) {
+	if (USART_FCR(usart) & USART_FCR_FIFOEN) {
+		USART_FCR(usart) = USART_FCR_TXTL(tx_depth) | USART_FCR_RXTL(rx_depth)
+						| USART_FCR_FIFOEN;
+		return true;
+	}
+	/* do nothing if FIFOs aren't enabled */
```
karlp, line 123: why not just have the function _do it_
```
USART_FCR |= FIFOEN | TXDEPTH | RXDEPTH;
```
What's the advantage of this? who would ever want to look at the return value here?

karlp:
In general, there's no point putting the data type into the doxygen comments, that information is available from the signature.  we can clean it up later though,

lib/pac55xx/usart.c
stefanik08: The FIFOEN bit must be set prior to calling this function.  It has to be done in 2 separate writes to the register.  I wanted some way to indicate "you're doing it wrong" if `usart_fifo_enable()` wasn't called first.

include/libopencm3/pac55xx/usart.h
stefanik08: fixed

include/libopencm3/pac55xx/usart.h
stefanik08: I think it probably would be easier to understand.  So, we could have usart_set_fifo_depth() first call usart_fifo_enable(), then this wouldn't need a return value.  Is that the cleanest way you all see @karlp @vielster ?

vielster: That sounds okay to me, I would just suggest it is documented that the call will enable FIFOs if they're not already.

stefanik08:
@karlp I think this is ready to merge.

karlp:

PR #1199, opened 2020-04-05
To merge/g0

grevaillot:
hi,

here are a couple of patches regarding g0 support : 
 - add some basic flash programming stuff, update doc and fix flash keys.
 - add desig stuff
 - update sysconfig for low pin count g0 chips

some flash flags clear helpers could be mutualized with l4. I'll have a look later, i need to check g4 RM to get a clearer picture of the job and avoid more duplication.


karlp:
Looks good

lib/stm32/g0/flash.c
```diff
@@ -30,6 +30,145 @@
...
+
+/** @brief Erase a page of FLASH
+ * @param[in] page
+ */
+void flash_erase_page(uint32_t page)
+{
+	flash_wait_for_last_operation();
+
+	uint32_t reg = FLASH_CR;
+	reg &= (FLASH_CR_PNB_SHIFT << FLASH_CR_PNB_SHIFT);
```
karlp, line 87: that doesn't look correct.  you want an ~ and a mask right?

lib/stm32/g0/flash.c
```diff
@@ -30,6 +30,145 @@
...
+/** @brief Erase All FLASH
+ * This performs all operations necessary to erase all sectors in the FLASH
+ * memory.
+ */
+void flash_erase_all_pages(void)
+{
+	flash_wait_for_last_operation();
+
+	FLASH_CR |= FLASH_CR_MER;
+	FLASH_CR |= FLASH_CR_STRT;
```
karlp, line 108: can these not be done together?

lib/stm32/g0/flash.c
```diff
@@ -30,6 +30,145 @@
...
+
+	FLASH_CR &= ~FLASH_CR_PG;
+}
+
+/** @brief Program a Data Block to FLASH
+ *
+ * This programs an arbitrary length data block to FLASH memory.
+ * The program error flag should be checked separately for the event that memory
+ * was not properly erased.
+ *
```
karlp, line 67: might be worth mentioning that it is min 8 bytes here?  (i guess that's common fialing of the docs really.)

lib/stm32/g0/flash.c
grevaillot: ow ..

lib/stm32/g0/flash.c
grevaillot: RM does that in two steps : 

> 3.  Set the MER1 bit of the FLASH control register (FLASH_CR).
> 4.  Set the STRT bit of the FLASH control register (FLASH_CR).

I've not tried, might work, or not, or might work sometime :) - I'd prefer to keep it that way.

lib/stm32/g0/flash.c
grevaillot: Indeed, same issue for other chips, that code ends up writting end of data garbage if not properly sized.
I wondered about putting an assert on data size ? might be a bit too much brutal

grevaillot:
push forced with proper mask

lib/stm32/g0/flash.c
karlp: fair enough, just looked at some existing code and it all does it in two steps too.  good enough for me :)

lib/stm32/g0/flash.c
karlp: yeah, we can just leave it.  not in scope for this at least.

karlp:
merged thanks!

PR #1201, opened 2020-04-07
stm32g0: fix RCC enums for UCPD peripherals

florolf:
The order of the UCPD1 and UCPD2 bits as described by the
rcc_periph_clken and rcc_periph_rst enums is swapped with regard to the
datasheet. Swap them.


grevaillot:
+1, indeed.

karlp:
merged thanks!

PR #1202, opened 2020-04-08
STM32: I2C Add missing filter register

ChuckM:
This register is missing in the current includes and is essential in
setting up slave operation. With out it the interface often issues
a "BUS ERROR" (I2C bus, not ARM memory bus).

### DIFF
--- a/include/libopencm3/stm32/common/i2c_common_v1.h
+++ b/include/libopencm3/stm32/common/i2c_common_v1.h
@@ -99,6 +99,11 @@ specific memorymap.h header before including this header file.*/
 #define I2C1_TRISE			I2C_TRISE(I2C1)
 #define I2C2_TRISE			I2C_TRISE(I2C2)
 
+/* FLTR register (I2Cx_FLTR) (noise filter) */
+#define I2C_FLTR(i2c_base)		MMIO32((i2c_base) + 0x24)
+#define I2C1_FLTR			I2C_FLTR(I2C1)
+#define I2C2_FLTR			I2C_FLTR(I2C2)
+
 /* Not all parts have i2c3 */
 #ifdef I2C3_BASE
 #define I2C3_CR1			I2C_CR1(I2C3)
@@ -110,6 +115,7 @@ specific memorymap.h header before including this header file.*/
 #define I2C3_SR2			I2C_SR2(I2C3)
 #define I2C3_CCR			I2C_CCR(I2C3)
 #define I2C3_TRISE			I2C_TRISE(I2C3)
+#define I2C3_FLTR			I2C_FLTR(I2C3)
 #endif
 
 /* --- I2Cx_CR1 values ----------------------------------------------------- */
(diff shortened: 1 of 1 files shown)


include/libopencm3/stm32/common/i2c_common_v1.h
```diff
@@ -365,6 +371,23 @@ specific memorymap.h header before including this header file.*/
  * TRISE[5:0]: Maximum rise time in Fast/Standard mode (master mode)
  */
 
+/* --- I2Cx_FLTR values ----------------------------------------------------- */
+
+/*
```
karlp, line 376: if you're adding docs, please add them as doxygen, and check they look how you expect from "make doc"

include/libopencm3/stm32/common/i2c_common_v1.h
```diff
@@ -365,6 +371,23 @@ specific memorymap.h header before including this header file.*/
...
+ * Bits [3:0] DNF - Digital noise filter
+ * These bits configure a digital noise filter on SDA and SCL pins.
+ * Value sets minimum pulse width needed to trigger i2c operations.
+ * 0 disables, 1 - 15 set minimum width to 'n' * TPCLK1
+ */
+#define	I2C_FLTR_DNF_MASK	0xf
+#define I2C_FLTR_DNF_SHIFT	0
```
karlp, line 383: do you have a whitespace issue here?

karlp:
Also, these digital filters belong in the f4 files, they're not present on f1, l1, or f2, which are the other i2c-v1 targets.

include/libopencm3/stm32/common/i2c_common_v1.h
ChuckM: I was matching style for style in the existing file.  Could use your guidance on if a commit that both doxygenizes and changes a file is "mixed changes in a single commit" or not. Happy to get the register issue first, and then go back and doxygenize and submit as a second PR, or change it into two commits in this PR. Let me know what you prefer.

include/libopencm3/stm32/common/i2c_common_v1.h
ChuckM: Yeah, I tend to use 4 space tabs (because screen width) I'll fix these up.

ChuckM:
Also need guidance on the question of F4 version. Are you asking that I remove these changes, and then go back to the f4 include and after it includes the common include add the changes there? And if so do I doxy them? I may be missing something about the doxy state here, so looking into that too.

include/libopencm3/stm32/common/i2c_common_v1.h
karlp: if the old style is non-doxygen, there's not a lot of reason to follow it.  You don't have to fix anything pre-existing, though, and yes, this might make a file mixed.

karlp:
Re f4, I mean, there is no i2c_fltr register on anyone that uses i2c-common-v1 _other than_ f4. So these register definitoins should go in the f4 file only (as they're _not_ common to the v1 peripheral)

karlp:
this merged in 3ee9710 but wasn't closed then

PR #1205, opened 2020-04-09
Update comments and fix discrepancies in generic Cortex-M linker script

roddehugo:
Hi,

First, thanks for the great work! It is very much appreciated.

While using the `lib/cortex-m-generic.ld` for bare-metal ARM programming I noticed a few discrepancies in the comments layout. I appreciate to have relevant code blocks commented as a separation between them. I took the liberty to remove the extra empty line at the end of the linker script.

Hope this helps,
Cheers

### DIFF
--- a/lib/cortex-m-generic.ld
+++ b/lib/cortex-m-generic.ld
@@ -112,4 +112,3 @@ SECTIONS
 
 /* Point to the very end of the RAM. */
 PROVIDE(_stack = ORIGIN(ram) + LENGTH(ram));
-
--- a/lib/cortex-m-generic.ld
+++ b/lib/cortex-m-generic.ld
@@ -43,15 +43,15 @@ ENTRY(reset_handler)
 SECTIONS
 {
 	.text : {
-		*(.vectors)	/* Vector table */
-		*(.text*)	/* Program code */
+		*(.vectors)	/* Vector table. */
+		*(.text*)	/* Program code. */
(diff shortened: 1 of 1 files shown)


karlp:
What's the discrepancy you found? Otherwise, sure I guess.  Squash your fixup commit though please. I'm not at all convinced that some of these count as sentences that need periods either, but ok, whatever, it's "just another commit"  I think that emission was what emmition was meant to be, and "visibility" is probably actually more correct.  Do you have other changes forthcoming?

roddehugo:
> What's the discrepancy you found?

It was actually in the comments. Some were following a kind of multi-line rule whereas other the single line one. Does not really matter I agree. Just my eye was twitching.

> Squash your fixup commit though please.

I'll do for sure.

> I think that emission was what emmition was meant to be, and "visibility" is probably actually more correct.

Sounds good to me. I'll edit the commit while rebasing.

> Do you have other changes forthcoming?

Nothing I am very convinced of yet. But I do question the absence of `KEEP` around `*(.vectors)` and `NOLOAD` for the `bss` section. What's your thoughts on that? It's another topic, might need a proper issue.

Thanks.

karlp:
has something changed on this that you no longer want it merged?

roddehugo:
I guess this is not really useful as these are minor comment updates. But If you like I can still squash and merge :)

PR #1206, opened 2020-04-13
I2c f4 fix

ChuckM:
Not the real PR here, I'm trying to figure out how to make Doxygen do something sane if I put the definition of the FLTR register in the F4's i2c file after including the i2c_common_v1.h file. @karlp any thoughts on that? Right now a `make doc` puts the FLTR macro definitions after the function calls on the first page which looks pretty ridiculous.


karlp:
merged thanks!

PR #1208, opened 2020-04-23
stm32f7: add DCMI registers

smolin35:
This patch adds definition of the DCMI registers for sm32f7 MCU series. 
I tested part of this code with nucleo-f746zg and nucleo-f767zi boards to get data from an external ADC.

### DIFF
--- a/include/libopencm3/stm32/f7/dcmi.h
+++ b/include/libopencm3/stm32/f7/dcmi.h
@@ -1,0 +1,246 @@
+/** @defgroup dcmi_defines DCMI Defines
+ *
+ * @ingroup STM32F7xx_defines
+ *
+ * @brief Defined Constants and Macros for the STM32F7xx DCMI Peripheral
+ *
+ * @version 1.0.0
+ *
+ * @date 2019-09-04
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+
+
+//TODO check
+/*
+ * STM32F7 DCMI Defines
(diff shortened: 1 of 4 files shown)


include/libopencm3/stm32/f7/dcmi.h
```diff
@@ -0,0 +1,246 @@
...
+ *
+ * @version 1.0.0
+ *
+ * @date 2019-09-04
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+
+
+//TODO check
```
karlp, line 15: check what? if this is finished, remove this please

karlp:
As far as I can see, this is 5 extra bits in the CR register?  This needs to be pulled out to a common dcmi peripheral header, with just the special extras for f7, not just copied over please.

smolin35:
Sorry for the long answer.

"As far as I can see, this is 5 extra bits in the CR register? "
Yes it is.
If I understand you right, you would like I put the common part for f4 and f7  in include/libopencm3/stm32/common/dcmi_common_f47.h, and leave the specific extras for f7 and f4 in include/libopencm3/stm32/fx/dcmi. h?

karlp:
This looks good, but the doxygen is now broken, f4 has no docs, and f7 only has the specific docs.

karlp:
NVM, I figured out how to fix the doxygen! :)

karlp:

Issue #1014, opened 2019-01-27
stm32f4:rcc: flash config set twice


alexkazik:
`clock->flash_config` does contain the wait state and cache config.
The caches are set.
And `flash_set_ws` clears the wait state bits and then or's the flash config and thus setting the cache again.

https://github.com/libopencm3/libopencm3/blob/1adc418f9afaa5b99c34f3eea797dc4c1155c056/lib/stm32/f4/rcc.c#L798

karlp:
yep. should be separated. feel free to send a pr,  but try not to break existing users too much.

wingunder:
@alexkazik, I hope you don't mind, but I took the liberty to submit a fix for this problem: #1209

alexkazik:
I don't mind. I forgot about this because I write in assembler now (at least for most parts).

PR #1209, opened 2020-04-23
Fix for 'flash config set twice' #1014.

wingunder:
The flash_set_ws() function didn't mask its input and it was possible
to set more bits in the FLASH_ACR register, than just the waitstate
bits. This caused the stm32f4 code to set the FLASH_ACR_DCEN and
FLASH_ACR_ICEN registers twice: Once before calling flash_set_ws() and
once inside flash_set_ws().

Once flash_set_ws() was changed to only set the waitstate bits of the
FLASH_ACR register, the rcc_clock_setup_hse_3v3() function for stm32f2
processors had to be modified to set/clear the FLASH_ACR_DCEN and
FLASH_ACR_ICEN registers, as it previously relied on the
flash_set_ws() function to do this.

The only processors that were affected by this feature and fix, was
the stm32f2 and stm32f4 processors, as they passed more than just the
waitstate bits to flash_set_ws().

flash_set_ws() now _only_ sets the waitstate bits of the FLASH_ACR
register.

### DIFF
--- a/lib/stm32/common/flash_common_all.c
+++ b/lib/stm32/common/flash_common_all.c
@@ -39,7 +39,7 @@ void flash_set_ws(uint32_t ws)
 
 	reg32 = FLASH_ACR;
 	reg32 &= ~(FLASH_ACR_LATENCY_MASK << FLASH_ACR_LATENCY_SHIFT);
-	reg32 |= (ws << FLASH_ACR_LATENCY_SHIFT);
+	reg32 |= (FLASH_ACR_LATENCY_MASK & ws) << FLASH_ACR_LATENCY_SHIFT;
 	FLASH_ACR = reg32;
 }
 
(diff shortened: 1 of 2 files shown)


karlp:
I think the correct path here is to not attempt to re-use the same field of the struct, I thought we'd already done that on other platforms?

wingunder:
Hi @karlp,
Thanks for your feedback. I hope that I understood you correctly and that 
c7b1629 puts this pr more in line with what you asked for. I took the liberty to deprecate rcc_clock_scale.flash_config. I could undo this if it's not required.
Regards

karlp:
I'm sorry to have led you astray here, but we're not going to start using bitfield definitions.  I intended just to rename things as necessary and drop that clearly wrong clearing of only certain wait state bits.  I'll review the other families and see what we can do, the bug is very valid, and needs to be fixed though  :)

wingunder:
No problem :)
Thanks for looking at it in any case.

PR #1212, opened 2020-04-26
sam:d: port: complete port (define, configuration and access)

trabucayre:
Tested with Feather M0 based on SAMD21 with success

### DIFF
--- a/include/libopencm3/sam/d/port.h
+++ b/include/libopencm3/sam/d/port.h
@@ -241,24 +241,35 @@ enum port_mux {
 /* --- Convenience enums --------------------------------------------------- */
 
 /* GPIO mode definitions (for convenience) */
-/** @defgroup gpio_mode GPIO mode configuration
-@ingroup gpio_mode
+/** @defgroup gpio_direction GPIO Pin direction
+@ingroup gpio_defines
+@li Input
+@li Output
+@li InOut
 @{*/
-enum gpio_mode {
-	GPIO_MODE_OUTPUT,
-	GPIO_MODE_INOUT,
-	GPIO_MODE_INPUT_FLOAT,
-	GPIO_MODE_INPUT_PULLDOWN,
-	GPIO_MODE_INPUT_PULLUP,
-	GPIO_MODE_AF
-};
+#define GPIO_MODE_INPUT		0x00
+#define GPIO_MODE_OUTPUT	0x01
+#define GPIO_MODE_INOUT		0x02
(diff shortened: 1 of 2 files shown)


grevaillot:
nice !

I had a project cancelled at work on samr34, so i started some support for d/l/r chips last summer. I'll have a review, but i think you can pick some bits - not much time currently to continue work in it :)

https://github.com/libopencm3/libopencm3/compare/master...grevaillot:wip/samlr

include/libopencm3/sam/d/port.h
```diff
@@ -66,3 +134,216 @@
...
+	GPIO_MODE_INPUT_PULLUP,
+	GPIO_MODE_AF
+};
+/**@}*/
+
+/* --- Function prototypes ------------------------------------------------- */
+
+BEGIN_DECLS
+
+void gpio_mode_setup(uint32_t gpioport, enum gpio_mode mode, uint32_t gpios);
```
grevaillot, line 261: most most chips here (except efm32 it seems) are are using mode setup with pull/down parameter :

> void gpio_mode_setup(uint32_t gpioport, uint8_t mode, uint8_t pull_up_down, uint16_t gpios)

maybe we could stick to that, and keep the same gpio_mode enum ?

include/libopencm3/sam/d/port.h
```diff
@@ -66,3 +134,216 @@
...
+
+/** @brief Set a group of Pins (atomic)
+ *
+ * Set a group of Pins for the given port. this is an atomic operation.
+ *
+ * @param[in] gpioport port register address base @ref port_reg_base
+ * @param[in] gpios @ref gpio_defines. Any combinaison of pins may be
+ *          specified by OR'ing then together.
+ */
+static inline void gpio_set(uint32_t gpioport, uint32_t gpios)
```
grevaillot, line 276: i'm not sure it is really usefull to inline these and keep them into headers -  compiler will do the job if needed, no ?

karlp: I'll go further and just say, "don't add static inline functions to the headers"

trabucayre: Thanks all for review. I've updated in consequence

karlp:
KARL: note to self, compare with _old_ samd stuff, see if anything overlaps

karlp:
this has merged in d11680638b with a set of followup commits on style and doxygen.  thanks for this!

PR #1215, opened 2020-05-03
Adds ability to enable .eh_frame while generating linker script

denizzzka:
Usage: just pass ARCH_FLAGS=-D_EH_FRAME while run mk/*.mk files

### DIFF
--- a/ld/linker.ld.S
+++ b/ld/linker.ld.S
@@ -182,11 +182,9 @@ SECTIONS
 	} >nfcram
 #endif
 
-	/*
-	 * The .eh_frame section appears to be used for C++ exception handling.
-	 * You may need to fix this if you're using C++.
-	 */
+#ifndef _EH_FRAME
 	/DISCARD/ : { *(.eh_frame) }
+#endif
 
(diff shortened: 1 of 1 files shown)


denizzzka:
I removed comment because it looks like copy and pasted, and eh_frame used not only with C++ and not only for exception handling

karlp:
Care to elaborate on what it _is_ used for then?

denizzzka:
For stack tracing and changing stackframes, as I know.
In any case there is no such comments on other variables in this file.

denizzzka:
An idea came up: it is possible to add all comments from `cortex-m-generic.ld` into `linker.ld.S` and enable `-C` option "Do not discard comments" for gcc preprocessor. Then a more pleasant file will be generated.

denizzzka:
This PR isn't usable, section leaved as not defined. External patching by build scripts will be better.

karlp:
unifying the cortex-m-generic and the template for the generator is definitely a good idea, totally agree.  As for this PR,  ok? great :) that was easy :)

PR #1221, opened 2020-05-25
add usb audio, cdc, and midi stub files for doxy

### LINKED ISSUE #1219: USB documentation issues
I noticed several documentation issues for USB:

1. Documentation for several subclasses is missing from the online documentation, even though the files themselves are doxygen-commented: audio (usb/audio.h), CDC (usb/cdc.h), DFU (usb/dfu.h), and MIDI (usb/midi.h). 
HID and MSC documentation is present, however. 
I don't really know anything about doxygen but I think the solution might be to add a stub source file for the missing docs in lib/usb/, which was already done for HID (see lib/usb/usb_hid.c, which is noted as a stub for doxygen)

2. The comment on ```struct usb_interface``` (usb/usbstd.h) says "you should never need to instance this struct in user code!" As far as I can tell this comment is false. All of the examples instance it for each interface (see [here](https://github.com/libopencm3/libopencm3-examples/blob/9b7c2c3dbf31a92a1d970910f58ff612190971da/examples/stm32/f1/stm32-h107/usb_simple/usb_simple.c#L53) for example), and my code doesn't work if I don't instance the struct and reference it in the usb_config_descriptor. Perhaps there's a way to make it work, but none of the examples use it? Or maybe I'm dense?

gkeeth:
- add stub files to relevant makefiles
- include <stdint.h> in respective headers to fix compilation

I didn't add a stub for DFU because #1220 does it, although it also makes some other changes. Perhaps it would be better to separate the Doxygen stub addition from the other documentation and implementation changes?

Fixes #1219

### DIFF
--- a/include/libopencm3/usb/audio.h
+++ b/include/libopencm3/usb/audio.h
@@ -40,6 +40,8 @@ LGPL License Terms @ref lgpl_license
 #ifndef LIBOPENCM3_USB_AUDIO_H
 #define LIBOPENCM3_USB_AUDIO_H
 
+#include <stdint.h>
+
 /*
  * Definitions from the USB_AUDIO_ or usb_audio_ namespace come from:
  * "Universal Serial Bus Class Definitions for Audio Devices, Revision 1.0"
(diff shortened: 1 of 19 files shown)


karlp:

PR #1223, opened 2020-05-29
stm32f3: Add support for OPAMP

ben-brewer-codethink:
Added (and tested) OPAMP support for STM32F3.

### DIFF
--- a/lib/stm32/f3/opamp.c
+++ b/lib/stm32/f3/opamp.c
@@ -1,0 +1,156 @@
+/** @defgroup opamp_file OPAMP
+ *
+ * @ingroup STM32F3xx
+ *
+ * @brief <b>libopencm3 STM32F3xx OPAMP</b>
+ *
+ * @version 1.0.0
+ *
+ * @date 22 May 2020
+ *
+ * LGPL License Terms @ref lgpl_license
+ */
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * This library is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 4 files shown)


ben-brewer-codethink:
Fixed some accessor functions.

lib/stm32/f3/opamp.c
```diff
@@ -0,0 +1,156 @@
...
+
+void opamp_vps_select(uint32_t base, uint32_t vps)
+{
+	OPAMP_CSR(base) &= ~(OPAMP_CSR_VPS_SEL_MASK
+		<< OPAMP_CSR_VPS_SEL_SHIFT);
+	OPAMP_CSR(base) |= vps << OPAMP_CSR_VPS_SEL_SHIFT;
+}
+
+void opamp_vms_select(uint32_t base, uint32_t vms)
+{
```
karlp, line 149: why _select on all of these, with _set for others?

ben-brewer-codethink: I used select wherever the input was expected to be an enum, and set where it's a value. Saw that kind of thing elsewhere and assumed, but am happy to change them all to set if that makes it more consistent?

karlp:
I've pulled this locally, but things that would make it "way more rad" are:
* doxygen of any sort for the functions and their arguments
* doxygen grouping of the registers and values

(I'm also 99% sure we're going to have to move this code into common/opamp for l1, l4, and l5, but... let'ss make that tomorrow's problem, not todays :)

karlp:
hah! I see now you were the one who added this to f3!  you knew you had a copy in g4!  naughty :)

Anyway, as this was "first" it has landed in master now, thanks!

PR #1226, opened 2020-06-01
Update checkpatch.pl with changes to the Linux Kernel Coding Style

Szalacinski:
The Linux kernel just deprecated the 80 character line limit.
This was always a mild issue of mine because of operations on lengthily-named registers that go just over the 80 character limit.
https://www.phoronix.com/scan.php?page=news_item&px=Linux-Kernel-Deprecates-80-Col


karlp:
sure, seems reasonable, though we've never _required_ things to be less than 80, just something to consider checking

karlp:

Issue #1230, opened 2020-07-01
adc_set_regular_sequence() is broken


rma-x:
The `adc_set_regular_sequence()` functions in adc_common_v1.c and adc_common_v2_multi.c ccontain several expresions that use `&` instead of `&&` and thus they don't work for sequences of more than four (v2) or six (v1) channels.
Additionally I think the whole functions should be refactored to be more efficient.
I plan to provide a patch.

karlp:
cute :)  WHile you're totally correct that it should be &&, because all the tests are bools, the & actually does work anyway, (but, by accident)  If you'd actually tried this you would have seen that the sequences do actually get set correctly.

rma-x:
Yes, it occurred to me afterwards that it might actually work, but only by relying on unguaranteed behavior of comparison operators.

PR #1231, opened 2020-07-02
Fix and improve adc_set_regular_sequence()

rma-x:
AFAICS all STM32 have a hardware divider or at least a fast multiplier, so using division here should be faster and more compact than an if-cascade.

This also fixes issue #1230 .

### DIFF
--- a/include/libopencm3/stm32/common/gpio_common_f234.h
+++ b/include/libopencm3/stm32/common/gpio_common_f234.h
@@ -51,6 +51,8 @@ specific memorymap.h header before including this header file.*/
 #define GPIOD				GPIO_PORT_D_BASE
 #define GPIOE				GPIO_PORT_E_BASE
 #define GPIOF				GPIO_PORT_F_BASE
+#define GPIOG				GPIO_PORT_G_BASE
+#define GPIOH				GPIO_PORT_H_BASE
 
 /**@}*/
 
@@ -64,6 +66,8 @@ specific memorymap.h header before including this header file.*/
 #define GPIOD_MODER			GPIO_MODER(GPIOD)
 #define GPIOE_MODER			GPIO_MODER(GPIOE)
 #define GPIOF_MODER			GPIO_MODER(GPIOF)
+#define GPIOG_MODER			GPIO_MODER(GPIOG)
+#define GPIOH_MODER			GPIO_MODER(GPIOH)
 
 /* Port output type register (GPIOx_OTYPER) */
 #define GPIO_OTYPER(port)		MMIO32((port) + 0x04)
@@ -73,6 +77,8 @@ specific memorymap.h header before including this header file.*/
 #define GPIOD_OTYPER			GPIO_OTYPER(GPIOD)
 #define GPIOE_OTYPER			GPIO_OTYPER(GPIOE)
 #define GPIOF_OTYPER			GPIO_OTYPER(GPIOF)
+#define GPIOG_OTYPER			GPIO_OTYPER(GPIOG)
+#define GPIOH_OTYPER			GPIO_OTYPER(GPIOH)
 
 /* Port output speed register (GPIOx_OSPEEDR) */
 #define GPIO_OSPEEDR(port)		MMIO32((port) + 0x08)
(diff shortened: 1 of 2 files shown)


lib/stm32/common/adc_common_v1.c
```diff
@@ -573,44 +573,28 @@ the newly defined group.
...
-		}
-		if ((i > 18) & (i <= 24)) {
-			fourth6 |= (channel[i - 1] << ((i - 18 - 1) * 5));
-		}
-		if ((i > 24) & (i <= 28)) {
-			fifth6 |= (channel[i - 1] << ((i - 24 - 1) * 5));
-		}
-	}
+	for (i = 0; i <= length; i++) {
+		int i1 = i+1;
```
karlp: just in general, you could  just do ``` i = 1; i < length;i++``` here, isntead of using this i1 variable

karlp:
Also, in general, we don't merge fixup commits.  You squash them yourself in your own branch.

More specifically, I'll totally merge a switch from & to && today.  It more clearly conveys the intent.  I'm not really interested in the other "cleanup" at this stage.

lib/stm32/common/adc_common_v1.c
rma-x: But then I would have to write `channel[i-1]` one line below. Pick your poison. ;)

rma-x:
> Also, in general, we don't merge fixup commits. You squash them yourself in your own branch.

Sorry, I don't quite get what you mean by this.

>  I'm not really interested in the other "cleanup" at this stage.

At what stage or time would you be?

karlp:
Your PR has two commits in it, one that introduces errors, and another one that fixes them.  If I merge your PR as is, we get both those commits in the permanent history, which is undesirable.  You fix your commits locally, so that the PR contains only good code.  (How to do that is kinda out of scope of this issue, but i'm happy to help if you come by the IRC channel)

As for when I want to see a switch to a division? probably never, as mentioned, M0 and M0+ don't have hardware divide.  The existing code is, sure, not "sexy" but it's been working in production for years for people.  Switching from the separate register vars to an array is a no-op change on it's own, so wouldn't ever come in by itself either.

The & vs && is a definite improvement, and well spotted.  I'm kinda curious what you were doing that made you look at it even, as it's perfectly functional, and most of us only end up looking the guts when we encounter breakages :)

karlp:
(it's generally not a very good idea to make a pull request based on your master branch, as it always updates.  You probably wanted to make a branch for this fix, and make a PR from that branch)

rma-x:
> I'm kinda curious what you were doing that made you look at it even, as it's perfectly functional, and most of us only end up looking the guts when we encounter breakages :)

I was using sequences for the first time and couldn't get them to work right away, due to a misunderstanding on my side, as it turned out later. As there was no instance of a sequence >1 in libopencm3-examples I wanted to make sure that it is not the setup code that causes my problem and had a look at it...

> As for when I want to see a switch to a division? probably never, as mentioned, M0 and M0+ don't have hardware divide.

Based on input from a colleague I was assuming that the compiler would play some clever tricks with the multiplier here, but I just verified it and it turned out that this is only the case for divide by 5 in the v2 code, and only when using -O3. All other optimization levels as well as the v1 code regardless of optimization end up calling library functions for division, so division indeed doesn't make sense here.

> You probably wanted to make a branch for this fix, and make a PR from that branch

OK, will try...

As you've probably guessed, I am new to contributing via git and especially github. So far I've always just sent patches or contributed to projects that use other VCSes.

karlp:
No problem.  I'm happy to help you if you want to just make the PR for switching the & to &&.  If it sounds like a boring work, I'm also happy to just do it myself.  Unfortunately, because of the way github PRs work, _this_ PR is basically broken now, as it's permanently tied to your repositoriy's master branch, and any changes you make there will always be reflected here.

You probably need to do something like
$ # from your current master
$ git checkout -b new-improved-version
$ # edit files further to get to just what you want to submit
$ git commit
$ git rebase -i master
# here, you should see, ~4 commits, your first two, your merge commit, and the final one, roughyl
# choose to "squish" all the preceding ones
$ git log -p # just to see that you now have one commit in your branch, with just the right code in it
$ git push # will push your new-improved-version branch to your github
# github should give you alink to create a new PR here, click that and do so

~roughly like that anyway.  If you come by IRC I can help more rapdily.

Alternatively, instead of squishing and editing away your old code, you can also do something like adding "upstream" libopencm3 as a remote, and checking out a new branch based on upstream/master, and then making just the simple edits to do the & to && and commit that and push it.  This is all just cleaning up around your modified local master.

PR #1234, opened 2020-07-08
[RFC] GD32VF103 support (RISC-V)

lkundrak:
Hi,

I've recently got my hands on an Alef Polos board with GD32VF103 on it: 
https://www.analoglamb.com/product/polos-gd32v-alef-board-risc-v-mcu-board/ 

It's a lot like a Blue Pill board, but with a RISC-V based MCU. It turns out the peripherals are similar enough to STM32 and this it might be reasonably easy to have it supported by a libopencm3; which indeed turned out to be true and that is what this patch set does.

While there is probably more work needed, I figured I'd share this now to see whether it makes sense to pursue this further and collect feedback early.

The current state and the overview of the MCU has been placed here for now:
https://github.com/hackerspace/libopencm3-gd32v/blob/lr/gd32v/lib/gd32v/f103/README.md

To build the branch, the a riscv-elf toolchain with newlib is needed. Fedora doesn't ship such toolchain, not sure about other distros. I'm using this one:
https://github.com/riscv-mcu/riscv-gnu-toolchain/releases

The branch is then built by setting `ARCH` to `riscv`:

```
$ make ARCH=riscv
```

I'll share the example project shortly that will demonstrate how it differs from STM32F103.

I'll be super thankful for any feedback.

Take care
Lubo


lkundrak:
Here's a project that runs on both STM32F103 and GD32VF103:
https://github.com/hackerspace/libopencm3-gf32v-examples/commit/d1e19733e7201

It utilizes GPIO, USART, SPI, sysclk, interrupts, Device ID and USB device peripheral. Watch out for `#ifdef GD32VF103` and `#ifndef GD32VF103` to find spots where code is different for a different MCU.

karlp:
Just upfront, most of your copyright headers are wrong.  When the file is simply an include of another file, you can eithe rhave no copyright (it's likely unenforceable anyway) or just your own.  You can't go adding multiple people from years in the past to all these stub headers.

lib/usb/usb_dwc_common.c
```diff
@@ -214,11 +214,11 @@ uint16_t dwc_ep_write_packet(usbd_device *usbd_dev, uint8_t addr,
 
 	/* Copy buffer to endpoint FIFO, note - memcpy does not work.
 	 * ARMv7M supports non-word-aligned accesses, ARMv6M does not. */
```
karlp, line 216: _if_ you're going to to do this sort of thing, you need to update the comments to.

lib/gd32v/f103/gd32vf103x4.ld
```diff
@@ -0,0 +1,33 @@
...
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/* Linker script for GD32F103x4, 16k flash, 6k RAM. */
```
karlp, line 21: none of these fixed files.  linker generator only

lkundrak:
> Just upfront, most of your copyright headers are wrong. When the file is simply an include of another file, you can eithe rhave no copyright (it's likely unenforceable anyway) or just your own. You can't go adding multiple people from years in the past to all these stub headers.

Not a lawyer here. I've kept the existing copyright headers in case I started by copying an existing file because I believe that in case the original file was copyrighted then mine is a derived work.

karlp:
Further, you've made a _very_ specific part id to be equivalent to the "family" level trees and defines used in other parts.  Are you _really_ sure about that?  if you _realllllly_ want to add "riscv" then you surely also need to be prepared for alternative riscv implementations too I presume.  I'm also concerned by how much duplication the rcc files appear to be.

lkundrak:
> Further, you've made a _very_ specific part id to be equivalent to the "family" level trees and defines used in other parts. Are you _really_ sure about that?

I'm not sure fully I understand this. By "part id" you mean "f103"? If so, I've done that because there are (for the time being anyway) no other similar chips. What could or should have I done instead?

> if you _realllllly_ want to add "riscv" then you surely also need to be prepared for alternative riscv implementations too I presume.

I don't get this either and need some clarification. How does the chip being a "riscv" matter? It's a detail that doesn't affect how the files are laid out or how the defines are named. I don't think anything I've done could affect a possible future chip with a riscv core. Unless it would be another gd32v chip and you meant to say "gd32v" or "f103" and imply that I could've chosen a better name? Or am I getting that completely wrong?

> I'm also concerned by how much duplication the rcc files appear to be.

The rcc driver has some similarities to gd32/f1x0 and stm32/f1, but is most similar to the latter. I kept it separate because I saw some duplication already going on between the two, and it was not immediately clear to me how to approach sharing code with otherwise largely unrelated MCU family. I'll try to do something about it, but ideas are more than welcome.

Thanks for taking time to look at this.

lkundrak:
v2
---

I've pushed second, hopefully better, version now. I've made an attempt to address the issues mentioned above; detailed change log is kept in descriptions of individual patches. This time I've made sure doxygen output looks reasonable. I still haven't touched the top-level README.md to mention GD32VF103 support, because it's not clear to me how to avoid turning it into a mess.

I also didn't remove the copyright notices from the stub headers, because I am not too sure they're not copyrightable. They files I copied them from claimed copyright and they seem to contain copyrightable material such as paragraphs of documentation text. Note that I am not a lawyer and I don't care too much
 about copyright; all I care about is a copyleft license applied wherever appropriate.

stappersg:
> Note that I am not a lawyer and I don't care too much about copyright; all I care about is a copyleft license applied wherever appropriate.

I do share that sentiment. And do know that copyright law is about protecting the rights of the copywriter.  It is the creator that decides to whom the creation belongs and lawyers tend to defend that choice in case it is not respected.

lkundrak:
Hi,

I've been contacted about this patch set privately, making me suspect there probably are users interested in this.

I've now rebased it, resolved the conflicts and added changes to README.md making it a (really) tiny bit less Cortex-M specific. I'm removing the [RFC] tag from the subject, since I believe it's basically ready. Please take a look.

Thanks
Lubo

lkundrak:
v4
---

I've now integrated patches sent to the list by Filip Moc (can't see them in the archives though...).

lkundrak:
v5
---

Rebased & tested.

UweBonnes:
I will sure have a look. Maybe you would also like to have a look at https://github.com/UweBonnes/blackmagic/tree/ruabmbua. This is the blackmagic probe extended to debug and flash GD32VF103. The state is "works for me", but I have no report yet about other peoples. having success with the PR.

The GD32VF103 as an additional BMP platform would also be fine.

reinauer:
Is there a path to get this integrated?

PR #1235, opened 2020-07-12
stm32/f1/rcc.c: fix/improve comments

diegoherranz:
- Fix some wrong references to HSE/HSI in the comments.
- Fix some frequencies mentioned in the comments.
- Clarify the frequency limits for flash wait states (it was ambiguous).
- Unify some sentences, capitalisation and space formatting.

Thanks!

### DIFF
--- a/lib/stm32/f1/rcc.c
+++ b/lib/stm32/f1/rcc.c
@@ -501,7 +501,7 @@ void rcc_set_rtc_clock_source(enum rcc_osc clock_source)
 /*---------------------------------------------------------------------------*/
 /** @brief ADC Setup the A/D Clock
 
-The ADC's have a common clock prescale setting.
+The ADCs have a common clock prescale setting.
 
 @param[in] adcpre uint32_t. Prescale divider taken from @ref rcc_cfgr_adcpre
 */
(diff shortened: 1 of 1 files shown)


lib/stm32/f1/rcc.c
```diff
@@ -637,10 +637,10 @@ void rcc_clock_setup_in_hsi_out_64mhz(void)
...
 	/*
-	 * Sysclk is running with 64MHz -> 2 waitstates.
-	 * 0WS from 0-24MHz
-	 * 1WS from 24-48MHz
-	 * 2WS from 48-72MHz
+	 * SYSCLK will run at 64MHz -> 2 waitstates.
+	 * 0WS if 0MHz < SYSCLK <= 24MHz
+	 * 1WS if 24MHz < SYSCLK <= 48MHz
+	 * 2WS if 48MHz < SYSCLK <= 72MHz
 	 */
```
karlp, line 644: This is a fixed function, if you want to touch these comments, just delete them IMO

karlp:
Most of these comments I would simply drop if they bother you in their present state.  They don't add any value over the code, and are, as you discovered prone to copy/paste errors.

diegoherranz:
Thanks for having a look at this.

I was trying to see how this clock setup worked and the comments were useful to me. However, some wrong ones were a bit misleading. They looked to be caused by copy/paste as you said.

I've now simplified many of them, removing redundant information which is not needed in many cases. I would prefer not to remove more comments since they were useful to me and I think they can be useful to other people. Let me know what you think. 

I guess the most important thing is avoiding wrong/false comments by either fixing them or getting rid of the wrong information.

Thanks!

karlp:
While these comment fixes are correct and useful, I'm going to close this.  I've brought struct based clock setup to f1, (as is done on every other target) so these fixed single purpose functions are now all deprecated, and only kept so that people's code doesn't break.

PR #1238, opened 2020-07-26
Enable running USB protocols in non-ISR context

matswebjorn:
Add functions for having a primary-ISR running in ISR context to register interrupts, and a secondary-ISR running at task context to parse interrupts and to call protocol handlers, thus enabling USB protocol handlers to run in task context.


karlp:
I'm afraid you need to make a better case for why you need this.  you'ðve got quite a lot of changes, and your comments on the functions doesn't really match your description here.

many/most of us run these in non-isr context already, just polling usbd_poll() from task context already, or setting events in the usb irqs and letting an rtos call usbd_poll when it gets the event.  Neither of those options require changes like you're proposing here.

PR #1239, opened 2020-07-31
469 device data

ChuckM:
Updates to the devices file and the i2c filter definition


ChuckM:
Something weird happened here, going to close this and try again.

PR #1243, opened 2020-08-04
Fix documentation

Themroc:
`gpio_set_output_options()` seems to be from some older (now gone) API.
`rcc_periph_clock_enable(RCC_TIM1);` appeared twice, but RCC_AFIO was missing.
GPIO A9 seems plain wrong there. AFAICT from the manual, A9 is only for timer 1 channel 2.


karlp:
what you're seeing is damage from the f1 gpio block having a different api to the later gpio blocks.

karlp:
I'm not sure what to do here honestly, either provide two examples in the doxygen comment or what.  F1 is the _only_ target with that form, but it's still an important target.

karlp:
I've pulled this locally, merging both forms into the example, hopefully that's an improvement

Themroc:
Didn't know that the F1-API is that different. In that case i'd say both examples should be there - just the way you did it. Thanx!

karlp:
This has landed in master now thanks!

PR #1244, opened 2020-08-06
STM32H7: Support the RNG Peripheral

mattofak:
The random number generator on the STM32H7 is identical to the v1 RNG
already present in the library. So use that, and make sure that the RCC
knows about the peripheral.


lib/stm32/h7/Makefile
```diff
@@ -42,7 +42,7 @@ OBJS += flash_common_all.o flash_common_f.o flash_common_f24.o
 OBJS += fmc_common_f47.o
 OBJS += gpio_common_all.o gpio_common_f0234.o
 OBJS += pwr.o rcc.o
-OBJS += rcc_common_all.o
+OBJS += rcc_common_all.o rng_common_v1.o
```
karlp, line 45: separate lines please, exactly for this reason, it reduces merge conflicts.  only things on same lines are same peripheral

lib/stm32/h7/rcc.c
```diff
@@ -415,3 +420,8 @@ void rcc_set_spi45_clksel(uint8_t clksel) {
 	RCC_D2CCIP1R &= ~(RCC_D2CCIP1R_SPI45SEL_MASK << RCC_D2CCIP1R_SPI45SEL_SHIFT);
 	RCC_D2CCIP1R |= clksel << RCC_D2CCIP1R_SPI45SEL_SHIFT;
 }
+
+void rcc_set_rng_clksel(uint8_t clksel) {
+  RCC_D2CCIP2R &= ~(RCC_D2CCIP2R_RNGSEL_MASK << RCC_D2CCIP2R_RNGSEL_SHIFT);
+  RCC_D2CCIP2R |= clksel << RCC_D2CCIP2R_RNGSEL_SHIFT;
```
karlp, line 426: whitespace looks wrong here at least

lib/stm32/h7/Makefile
mattofak: Would you like me to split the other lines while we're at it? (Or is it that things related to the same peripheral can be kept together?)

karlp:

PR #1245, opened 2020-08-06
STM32H7 DAC Updates

mattofak:
The DAC on the STM32H7 has an additional trigger bit for each channel
and some additional choices when it comes to low power operation. ST
therefore opted to give us a new DAC mode register and shift the trigger
channel in the CR register.

Account for this by checking to see if we should define the MCR register
and if so shift some things around and define a new set_mode function.

The old enable/disable buffering is emulated by setting the appropriate
new modes.

### DIFF
--- a/include/libopencm3/stm32/common/dac_common_all.h
+++ b/include/libopencm3/stm32/common/dac_common_all.h
@@ -77,6 +77,13 @@ specific memorymap.h header before including this header file.*/
 /* DAC channel2 data output register (DAC_DOR2) */
 #define DAC_DOR2			MMIO32(DAC_BASE + 0x30)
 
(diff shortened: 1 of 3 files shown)


karlp:
I can't say I like this approach, I'll look over these sheets and see what we can do.

mattofak:
I thought about making a dac_v2; but it's soooooo close to being compatible. Story of the H7 it seems.

karlp:
Someone else needed a dac-v2 for g4 and some f3 parts, so we have a dac-v2 now, and we also took that as the required time to break the dac apis, and add dac parameters to them all.  Can this all j ust switch to using that?  I do vastly prefer the enum mode types from this PR though, and would happily replace the mode1/mode2 defines that just came in.

PR #1246, opened 2020-08-19
STM32H7: Add QSPI Defines

mattofak:
Move the STM32F4 QuadSPI peripheral defines to the common folder as the
F4 and H7 variants of the IP share almost all the same bits. For those
bits that are separate put them into their own headers.

### DIFF
--- a/include/libopencm3/stm32/common/quadspi_common_v1.h
+++ b/include/libopencm3/stm32/common/quadspi_common_v1.h
@@ -1,0 +1,169 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ *
+ * This library is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


karlp:
This looks reasonable, and I'm merging it, but I'm pretty sure this isn't what you actually had locally for testing this, as you didn't include the changes to stm32/quadspi.h to actually include the h7 part.  It's a lot easier for me ifyou can make sure your PRs actyually have the working code :)  I know it's tricky when you have multiple patches outstanding, I do commiserate, but please try :)

karlp:
(I've also kept chuck's copyrighton the v1 file as you just moved it all.)

karlp:
ok, all merged now!

PR #1247, opened 2020-08-19
stm32h7: fixed flash impl from pretending to be stm32 f2/f4.

vielster:
Added the minimum functionality required for general use (setting waitstates).

### DIFF
--- a/include/libopencm3/stm32/h7/flash.h
+++ b/include/libopencm3/stm32/h7/flash.h
@@ -27,27 +27,44 @@
 #define LIBOPENCM3_FLASH_H
 
 #include <libopencm3/stm32/common/flash_common_all.h>
-#include <libopencm3/stm32/common/flash_common_f.h>
-#include <libopencm3/stm32/common/flash_common_f24.h>
 
 /**@{*/
-
-/** @addtogroup flash_acr_values FLASH_ACR values
- * @ingroup flash_registers
+/** @defgroup flash_registers Flash Registers
+ * @ingroup flash_defines
 @{*/
-#define FLASH_ACR_WRHF_VOS1_70MHZ     (0 << FLASH_ACR_WRHIGHFREQ_SHIFT)
-#define FLASH_ACR_WRHF_VOS1_185MHZ    (1 << FLASH_ACR_WRHIGHFREQ_SHIFT)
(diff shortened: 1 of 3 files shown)


include/libopencm3/stm32/h7/flash.h
```diff
@@ -27,27 +27,53 @@
...
-#define FLASH_ACR_WRHF_VOS3_225MHZ    (2 << FLASH_ACR_WRHIGHFREQ_SHIFT)
+#define FLASH_ACR_LATENCY(w)          ((w) & FLASH_ACR_LATENCY_MASK)
+#define FLASH_ACR_LATENCY_0WS         0x00
+#define FLASH_ACR_LATENCY_1WS         0x01
+#define FLASH_ACR_LATENCY_2WS         0x02
+#define FLASH_ACR_LATENCY_3WS         0x03
+#define FLASH_ACR_LATENCY_4WS         0x04
+#define FLASH_ACR_LATENCY_5WS         0x05
+#define FLASH_ACR_LATENCY_6WS         0x06
+#define FLASH_ACR_LATENCY_7WS         0x07
```
karlp, line 54: I don't really see a lot of value in having defines that are are N=N, which APIs do you think are improved by this?

include/libopencm3/stm32/h7/flash.h
```diff
@@ -27,27 +27,53 @@
...
 /*@}*/
-#define FLASH_ACR_WRHIGHFREQ_MASK     (0x3)
-#define FLASH_ACR_WRHIGHFREQ_SHIFT    (0x4)
 
+/** @defgroup flash_acr_values FLASH_ACR values
+ * @ingroup flash_registers
+ * @brief Access Control register values
+ * @{*/
+/* Unsupported, make this operation a no-op. */
+#define FLASH_ACR_PRFTEN              (0)
```
karlp, line 66: which common api is trying to use this that breaks without this define?  It seems cute, but I'm a little surprised it maybe surprising.

include/libopencm3/stm32/h7/flash.h
vielster: I'm just modeling after other implementations. I'm happy to remove them as we use the integer values in our code.

include/libopencm3/stm32/h7/flash.h
vielster: flash_prefetch_enable()/disable()

PR #1248, opened 2020-09-02
Double buffering for stm32f0

tblah:
Adds double buffered bulk endpoint support for stm32f0.

The fix in  d3bfb68  is kept separate because it isn't obvious that it is needed from the datasheet (which is ambiguous about which bits of double buffered flow control are done on the first transaction). I arrived at that fix experimentally.

### DIFF
--- a/include/libopencm3/stm32/common/st_usbfs_common.h
+++ b/include/libopencm3/stm32/common/st_usbfs_common.h
@@ -160,6 +160,7 @@ LGPL License Terms @ref lgpl_license
 
 #define USB_EP_SETUP		0x0800 /* Setup transaction completed */
 #define USB_EP_TYPE		0x0600 /* Endpoint type */
+#define USB_EP_TYPE_OFFSET	9      /* Endpoint type offset */
 #define USB_EP_KIND		0x0100 /* Endpoint kind.
 				* When set and type=bulk    -> double buffer
 				* When set and type=control -> status out
@@ -242,6 +243,9 @@ LGPL License Terms @ref lgpl_license
 		(USB_EP_NTOGGLE_MSK & \
 		(~USB_EP_TYPE))) | TYPE)
 
+#define USB_GET_EP_TYPE(EP) \
+	((GET_REG(USB_EP_REG(EP)) & USB_EP_TYPE) >> USB_EP_TYPE_OFFSET)
+
 #define USB_SET_EP_KIND(EP) \
 	SET_REG(USB_EP_REG(EP), \
 		(GET_REG(USB_EP_REG(EP)) & \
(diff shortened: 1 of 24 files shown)


manuelbl:
Double buffered USB would be a great addition to libopencm3. Can you describe how your addition is supposed to be used?

- How is double buffering activated? Can it be activated for certain endpoints individually?
- How does flow control for bulk endpoints work? How can an OUT endpoint be stalled?
- If data is received in quick succession: is it possible to retrieve data from two buffers at once or does the data need to be retrieved from two separate callbacks?
- How does `usbd_ep_write_packet` behave? Will the first call fill the first buffer, the second call the second buffer and third one will return an error (if not data is transmitted in between)? Is it possible to check if the second buffer is available for writing?

tblah:
Double buffering would be activated globally by passing `&st_usbfs_double_usb_driver` as the first argument to `usbd_init`. Non-bulk endpoints are not double buffered. I'm aiming to open source a project using double buffering soon, hopefully this will make it clearer.

Flow control for OUT endpoints happens automatically. The hardware keeps track of which buffer it is currently sending and which buffer is in use by software and so when it receives a packet it can write the packet into the buffer indicated by HW_BUF if and only if SW_BUF indicates that buffer is free. Otherwise it will hold the endpoint in NAK status.

A callback is required for each packet.

For `usbd_ep_write_packet` I had to make this block until the *previous* buffer is sent. So if the function returns then the packet was sent successfully and there is now a new buffer available to write again. I haven't implemented a way to check if the function would block in advance of calling it because I wasn't sure how this would fit into the usb driver api. Basically you would have to check if SW_BUF equals HW_BUF for that endpoint, if so the function will not block.

manuelbl:
Thanks a lot for the detailed answer. It sounds very reasonable except for the blocking. The current implementation of `st_usbfs_double_ep_write_packet` immediately returns 0 if the buffer is still occupied by previous data. You might want to consider doing the same.

tblah:
I'm glad I could help.

Yeah I thought about following the normal libopencm3 convention but doing this would remove the advantage of double buffering. For example, if I checked if the hardware is finished sending the previous buffer at the start of `st_usbfs_double_ep_write_packet` then that would force copying into the current buffer not to start until the hardware was done sending the previous buffer, in which case you might as well always use only one buffer.

The expected use case is that the hardware is not quite done sending the previous buffer when we start copying in the new buffer and that it will finish some time during that copy. But if it does not then we have to block otherwise the caller could call `st_usbfs_double_ep_write_packet` again too soon.

Please let me know if you have any better ideas :)

manuelbl:
Double buffering is relevant for cases requiring high bandwidth. In those cases, data buffering will be used and the next endpoint packet will be submitted whenever a previous one has been successfully transmitted to keep the USB peripheral busy. So `usbd_ep_write_packet` will be called from the endpoint callback, which in turn is called from the USB interrupt handler. So we would end up blocking the interrupt handler for up to 1ms. I think that's a definite no go.

I'm not sure I understand your explanation. The behavior I would expect is:

- `usbd_ep_write_packet` successfully submits a packet if either no PMA buffer is in use or just a single one.  No blocking is used. 
- `usbd_ep_write_packet` immediately returns 0 if both PMA buffers are in use.

"in use" means the buffer is under the control of the USB peripheral, i.e. it is waiting to be transmitted or being transmitted at the moment.

This approach should allow to leverage the full potential of double buffering and should be safe (no data is overwritten).

The easiest way to implement it is to use a counter: increase it every time a buffer is filled, decrease it every time a transmission has completed. If the counter is 2, both buffers are in used and `usbd_ep_write_packet` should return 0.

I think it's even possible to derive the state from the USB registers. So the counter might not even be needed. I'll think about it some more.

manuelbl:
By the way: A full implementation will likely require two or three driver to also cover the STM32 devices with older USB peripheral and/or the USB OTG peripheral. So instead of `st_usbfs_double_usb_driver`, it should probably be called `st_usbfs_double_usb_driver_v2` to match the name of the driver it replaces.

tblah:
Thanks for your feedback.

My use case did not involve calling it from an interrupt context, but I can see why you would want to and I will see what I can do.

I understand the behavior you would expect. I was unable to recreate this behavior using only the information in the hardware registers, but yeah perhaps storing some per-endpoint state such as the counter you mention would work, if I can get the right information from interrupts. I will think about it, thanks.

+1 to the name change

manuelbl:
Here's an attempt to derive the status form the endpoint register:

| DTOG_TX  | SW_BUF_TX  | STAT_TX  | Status  | Action  |
| :-------------: | :-------------: | :-------------: | ------------- | ------------- |
| 0 | 0 | VALID | Both buffers are in use | Return 0 |
| 0 | 0 | * | Idle | Fill buffer 0 |
| 0 | 1 | * | Buffer 0 in use | Fill buffer 1 |
| 1 | 0 | * | Buffer 1 in use | Fill buffer 0 |
| 1 | 1 | VALID | Both buffers are in use | Return 0 |
| 1 | 1 | * | Idle | Fill buffer 1 |

I'm not entirely sure if my understanding of the STAT_TX bits is correct. The decision table above works if the USB peripheral has completed sending a packet, toggles DTOG_TX, recognizes that no new packet is available (DTOG_TX == SW_BUF_TX) and immediately sets the STAT_TX to NAK.

The description of the reference manual could also be read that NAK is not set immediately but only when the host initiates the next transaction. But that's an unlikely implementation.

tblah:
Thanks for taking a look at this.

I don't think that will work with the STAT bit pair. From the top of page 870 of the reference manual:
> STAT bit pair is not affected by the transaction termination and its value remains '11 (Valid). However, as the token packet of a new transaction is received, the actual endpoint status will be masked as '10 (NAK) when a buffer conflict between the USB peripheral and the application software is detected

My interpretation of this is that the STAT pair will always read as valid from software but the USB peripheral will transmit a NAK anyway (masking the status in the software-addressable register) when there is a buffer conflict.

manuelbl:
It can certainly be interpreted that way. Might still be worth a try to figure out how it effectively works. The description isn't easy to understand.

tblah:
Thanks @manuelbl you were right about the datasheet. See 997ceef

karlp:
> So `usbd_ep_write_packet` will be called from the endpoint callback, which in turn is called from the USB interrupt handler. So we would end up blocking the interrupt handler for up to 1ms. I think that's a definite no go.

Do note that there's no requirement that you hook usbd_poll to an interrupt handler.  Some people do, and we shouldn't automatically break that, but you shouldn't build assumptions on that.

manuelbl:
@tblah Thanks, I'll probably look into on the weekend.

I'm also looking into another issue affecting all the single buffered drivers. Flow control is unnecessarily difficult to implement as the endpoint is automatically reset to VALID if the data is read from the PMA buffer. To fix it, the current USB interface probably needs to be changed. Very ugly...

@karlp I agree. Both approaches for calling `usbd_poll` should work, and from what I have seen in the PR, they do. I've used the interrupt based approach to emphasize that blocking isn't a good idea.

karlp:
> Flow control is unnecessarily difficult to implement as the endpoint is automatically reset to VALID if the data is read from the PMA buffer. To fix it, the current USB interface probably needs to be changed. Very ugly...

yes :) this is hands down the longest standing problem with the stack, and arguably has been the cause of many of the harder to diagnose bugs.  complete redesign would definitely be one of the options.

manuelbl:
I don't think that a complete redesign is needed. The adjustments are
likely minor but they will break the current driver API. The client API
could probably provide  backward compatibility.

I will think about it some more and propose a suitable design soon.

tblah:
Above rebase is to pull in the CI changes on master, as suggested on IRC

tblah:
I've been working on integrating double buffering into gadget0 tests. See the WIP commits just pushed. The tests don't pass and I don't have any good ideas how to fix it so I will close this pull request. The commits are all here if anyone wants to take a look.

I was testing also using the commits in https://github.com/libopencm3/libopencm3/pull/1251 because otherwise the tests don't even pass for the single buffered v2 driver on my setup.

Several tests fail, but I've been focusing on `TestConfigLoopBack.test_simple_loop` as it seems simplest. In essence it does the following
1. Configure the device in the sourcesink configuration
2. Send one packet
3. Reconfigure the device in loopback configuration
4. Receive a packet
5. Send that packet back

The issue I am seeing is that the hardware is sending an extra packet in addition to the one deliberately sent in step 5. These are sent resulting from the call to the packet_write function (not earlier). The extra packet is delivered to to the host correctly, with libusb not indicating any errors, however the callback for CTR_TX is not invoked for the extra packet. According to the reference manual this should happen only when the transaction is ended with a NAK or a STALL, but in that case I don't think the packet would show up on the host without error.

The issue only happens when step 2 is present: without that write before the configuration is changed, double buffering works correctly as far as I can tell.

I suspect the issue is related to my workaround in https://github.com/libopencm3/libopencm3/pull/1248/commits/fe117a875a4709e63c1c36c2ce74422dc1bf4425, but I never got properly to the bottom of what was going on there.

PR #1251, opened 2020-09-21
Gadget0 fix

tblah:
I was having trouble getting the gadget0 tests to pass reliably on stm32f0-discovery. These patches address that.

See commit messages for details.


karlp:
The delay ~must be there, as it's how we enforce that user code is actually allowed to run.  There's likely been a regression in other places, and it's why my current project is getting more test boards into automatic regression testing.

tblah:
Control packets are sent by the host more often than the length of the delay. This means that the second packet will (sometimes) be dropped because we haven't yet read in the first packet. I haven't moved the delay entirely, just reduced it.

I don't think this can be a regression in libopencm3 because I don't think it is libopencm3 can influence how quickly the host is sending packets.

Either we have to poll often enough not to drop the packets, or we have to handle the control packets with an interrupt somewhere, or the tests need to allow some of the control packets to be dropped.

karlp:
try before https://github.com/libopencm3/libopencm3/commit/26ab78a710b4cb7abcace769768c4781332b17ba#diff-347d4202e8c8ffe47b02c486ecca187b or before https://github.com/libopencm3/libopencm3/commit/58f2ee34fa7b6fc04c41acff221d7e916467f7b9#diff-347d4202e8c8ffe47b02c486ecca187b instead.  f0 _did_ work, and it is currently somewhat expected that something's not quite right.

tblah:
Thanks. I will take a look tomorrow

tblah:
@karlp do you know which revision definitely did work for f0? Do you remember if you tested it once or several times? The issue fixed by 0fc7b21 didn't occur every time for me (but is very likely to occur within 5 test runs).

Going to the commit immediately before each of those didn't work but I can see how they look related so I will dig further. Thanks for suggesting them.

karlp:
I've tested it multiple times in the past, but it was exhausting doing it for every commit, so it slipped, and as you've found, doesn't pass all tests at the moment.  The 100us delay is pretty hard, I would tolerate lengthening it, but even enumeration time you are allowed more time to respond, let alone normally.  requiring user apps to poll the usb layer every 50usecs is far too heavy a requirement to impose on them.

tblah:
Thanks. I wonder if this is a problem with my host machine configuration then because I haven't seen the tests pass on any commit I have looked at since the delay was added.

tblah:
Above rebase is to pull in the CI changes on master, as suggested on IRC

PR #1253, opened 2020-10-01
Trivial makefile path change

denizzzka:


karlp:
Could you please provide some explanation of the motivation here?

denizzzka:
I just noticed that this can be replaced to more convient look. Apparently, the current version remained from the past.
(I used this code months ago with my own addon to generate libraries from Meson build system and cherryp-picked it to this PR.)

karlp:

PR #1256, opened 2020-10-11
Disable VBUS sensing

manuelbl:
Fixes #1119 
Partially fixes #1073 (*).

So far, the USB OTG peripheral has been initialized with active VBUS sensing. This doesn't make much sense as the driver is a device mode only driver. It does not support host mode let alone switching between modes. Furthermore, VBUS sensing requires VBUS be routed to a specific pin.

The unfortunate side effect was that the USB OTG driver would not work with several F4 series chips, in particular F411CE and F446RE (probably all F446xx versions).

This pull request disables VBUS sensing and fixes the issues on the above chips (*). It covers both USB core ID 0x1200 and 0x2000.

The code has been successfully tested on:
 - Chinese Black Pill with STM32F401CCU6
 - STM Nucleo-F411RE
 - STM Nucleo-F446RE 

The tests included my bulk endpoint speed test (source, sink, loopback) as well as *gadget0*. Note that three *gadget0* tests fail on all boards, before the code changes, after the code changes and with boards using a different USB driver.

If anyone would like to use VBUS sensing, it can be easily activated after initializing the USB driver.

I also tried to include a Nucelo-L476RG board into the tests. However, it seems that libOpenCM3 thinks that all L4 series chips have a USB FS device peripheral, when in fact only L41xx through L46xx have while L47xx through L4Axx have an USB OTG peripheral. That's will be a separate issue...

(*) As mentioned in #1073, the F446RE additionally requires reverting 26ab78a.

PR #1257, opened 2020-10-18
USB OTG driver for STM32L4

manuelbl:
So far libOpenCM3 assumes that all STM32L4 series chips have a USB FS device peripheral. However, this only holds for L41xx through L46xx while L47xx through L4Axx have an USB OTG peripheral.

This PR adds the necessary code so both the `otgfs_usb_driver` and `st_usbfs_v2_usb_driver` USB driver can be used.

The PR has been successfully tested with a NUCLEO-L476RG board.

To work without wiring the USB sensing pins), PR #1256 needs to be applied as well. In addition, 26ab78a needs to reverted (as seems to be the case for OTG peripherals with core ID 0x2000).


manuelbl:
@karlp Can you help me with the build problems? Why does it fail with the *No rule to make target '../../../include/libopencm3/usb/hid_usage_tables.h', needed by 'usb_hid.o'* even though their is no reference to `hid_usage_tables.h` nor a file of that name?

karlp:
don't stress the jenkins build failure, I'm having problems with that as well, I'm not sure of the cause right now.  whenever I try annd build manually on that machine it works normally. *shrugs*

include/libopencm3/stm32/l4/memorymap.h
```diff
@@ -106,6 +106,7 @@
 #define GPIO_PORT_H_BASE		(PERIPH_BASE_AHB2 + 0x1c00)
 /* Still AHB2, good job ST */
 #define OTG_FS_BASE			(0x50000000U + 0x00000)
+#define USB_OTG_FS_BASE     OTG_FS_BASE
```
karlp, line 109: as OTG was never enabled on l4, just drop the original defintion, and use the "standard" definitoin please.

manuelbl: Ok, done

manuelbl:
Does the Jenkins server have any special features (such as attached hardware) that couldn't be replaced by GitHub Actions? I would be willing to set it up.

karlp:
yes, it has attached hardware, that's the entire point :)  it's purpose is to run the gadget0 usb tests on actual hardware.  It only has two boards at present, but more are planned.  the bizzare build errors are blocking that at present.

karlp:
I'vee pulled this locally thanks.

karlp:
This has landed in master now thanks!

Issue #1242, opened 2020-08-03
USB dwc driver - improper endpoint initialization, can cause the very first IN packet to be discarded


stoyan-shopov:
I am doing some hacking for supporting usb high speed on stm32f723 devices.
I am not yet ready with a pull request, but I am describing this issue anyway, maybe some other people are having troubles with this as well.

Note that I am working on stm32f723, this issue may be non-present on other devices (but my suspicion is that it is present on at least some other devices) using the dwc driver.

Endpoint initialization in function `dwc_ep_setup()`, in file `usb_dwc_common.c` initializes endpoints with the `OTG_DIEPCTL0_EPENA ` bit set.

The setting of this flag for usb control endpoint 0 is here:
https://github.com/libopencm3/libopencm3/blob/24bef9c49eda109e92e926e065b246a71d454f2d/lib/usb/usb_dwc_common.c#L63-L64

For other endpoints, the setting of the flag is here:
https://github.com/libopencm3/libopencm3/blob/24bef9c49eda109e92e926e065b246a71d454f2d/lib/usb/usb_dwc_common.c#L89-L92

In my case, with this setup, after powering the device, the very first IN packet is **always** discarded. From inspecting the sources, writing IN packets is being done in function `dwc_ep_write_packet()`, in file `usb_dwc_common.c`, and these particular lines get interesting:
https://github.com/libopencm3/libopencm3/blob/24bef9c49eda109e92e926e065b246a71d454f2d/lib/usb/usb_dwc_common.c#L210-L213

This code assumes, that the endpoint is **NOT** enabled, which will not be true **only** for the **very first** IN packet. The st documentation explicitly states that the `DIEPTSIZ` register must be written with the size of the packet to be sent when the endpoint is **not enabled**, and only enable the endpoint **after** writing the `DIEPTSIZ` register.

Note that this issue is hard to notice for control endpoints, but it is obvious for other endpoints - in my case I am testing with a CDC-ACM usb class device, and I see that the very first packet that I am expecting, is always missed.

For me, the solution is deleting the `OTG_DIEPCTL0_EPENA` flag from the endpoint initialization code, on these lines:

https://github.com/libopencm3/libopencm3/blob/24bef9c49eda109e92e926e065b246a71d454f2d/lib/usb/usb_dwc_common.c#L64
https://github.com/libopencm3/libopencm3/blob/24bef9c49eda109e92e926e065b246a71d454f2d/lib/usb/usb_dwc_common.c#L90

[...]

PR #1258, opened 2020-10-23
Fix usb_dwc_common.c endpoint initialization

stoyan-shopov:
This commit addresses libopencm3 issue #1242:
https://github.com/libopencm3/libopencm3/issues/1242


dragonmux (approved):
Having checked this patch over and rebased it on the current state of things + checked against the datasheet - all looks good. This correctly fixes setup arming the IN endpoints for transmission before the firmware's actually ready and has something to send.

Thank you for this fix and apologies it's taken so long to get merged! Merging now.

PR #1259, opened 2020-10-23
Fix usb_dwc_common.c late IN usb interrupt flag clearing, #1241

stoyan-shopov:
This commit fixes issue #1241

Issue #1241, opened 2020-08-03
USB dwc driver - interrupt flag for IN packets cleared too late, can lead to device not functioning


stoyan-shopov:
I am doing some hacking for supporting usb high speed on stm32f723 devices.
I am not yet ready with a pull request, but I am describing this issue anyway, maybe some other people are having troubles with this as well. Note that this issue only occurs on 1 in 4 machines that I tested on, so it may be easy to miss. I spent a fair amount of time debugging this (without a hardware packet analyzer), so hopefully this can save other people time debugging this once more.

I found that the clearing of the interrupt flag for IN packet completion in `usb_dwc_common.c` is being done too late:

https://github.com/libopencm3/libopencm3/blob/24bef9c49eda109e92e926e065b246a71d454f2d/lib/usb/usb_dwc_common.c#L342-L353

I believe, line 351: `REBASE(OTG_DIEPINT(i)) = OTG_DIEPINTX_XFRC;` should be moved below line 344, just before the (eventual) invocation of the callback for handling the IN packet completion.

Here is what I am observing with the current source code, for the case of a usb control transfer that has multiple IN packets in the data stage of the control transfer.

To be more concrete, I am describing the case of reading a device descriptor, which is 71 bytes long - this means that, for a control endpoint which is 64 bytes in size, there will be two IN packets in the data stage of the `GET_DESCRIPTOR` usb control transfer - the first packet is 64 bytes long, and the second is 71 - 64 = 7 bytes long:
- (1) the usb host sends a `GET_DESCRIPTOR` setup packet to the device
- (2) libopencm3 decodes the setup request, and starts sending data - it sends the first packet from the data stage, which is a 64 byte IN packet
- (3) the usb host acknowledges reception of the first usb IN packet; this triggers an interrupt, which is handled in `usb_dwc_common.c`; the interrupt handler detects this, and invokes the correct transfer (`ctr`) callback, at line 347 from `usb_dwc_common.c`; in this particular case, this callback function is `_usbd_control_in()` on line 286 in file `usb_control.c`
- (4) the usb control state variable (`usbd_dev->control_state.state`) in this case is `DATA_IN`, so function `usb_control_send_chunk()` is invoked from `_usbd_control_in()`; this function sends the second (and last) data packet from the data stage of the control transfer (in this case - the 7 byte packet mentioned above)
[...]

Issue #873, opened 2018-01-19
usb: Regression of the FIFO word loss issue.


zyp:
Both @H2OBrain and @alibabashack are reporting (in comments on #785 and #600 respectively) that they are seeing a regression of the FIFO word loss issue originally reported as #600, caused by the introduction of #785.

I'm filing the regression as a new issue so we can keep all the discussion here, rather than spread around as comments on older tickets. Here's my preliminary analysis of the regression which I posted previously in a different comment:

> For reference, as far as I can recall at the moment, this issue were caused by a race condition in the usb hardware. It was fixed in 0c1ff68 which is rebased from #672, by rearranging the order of some operations related to flow control (CNAK/SNAK).
>
>#785 enables the use of flow control on EP0 to ensure that control events are handled in order. It is probably adding flow control operations at a point that hits the race condition again.

zyp:
After looking further into it, I've concluded that my preliminary analysis looks correct. The problem is that setting CNAK before the SETUP complete event is received causes a race condition. The SETUP callback is called when the SETUP packet event is received, which means that setting CNAK from the callback is too early.

Originally the problem was that CNAK was set by ep_read() which is called by the callback. #672 solved this by moving CNAK out of ep_read() and calling it after the SETUP complete event is received instead.

The regression by #785 is caused by the introduction of flow control calls into the SETUP callback. They also set CNAK.

To solve this properly, I propose changing the event handling code to only call the SETUP callback after the SETUP complete event is received. Unfortunately, this implies that the callback can't call ep_read() itself anymore, because the packet has to be read out of the FIFO before the SETUP complete event arrives. This implies a change of the API between the hardware drivers and _usbd_control_setup().

I have pushed a quick test fix here: https://github.com/zyp/libopencm3/tree/test_fix_otg_word_loss_regression

Please test with that branch and check if it appears to solve the problem. Note that I don't have access to hardware at the moment, so I haven't tested it myself.

H2OBrain:
@zyp, with your test_fix the control control-loopback test runs through (with and without interrupts enabled). I also adde a pull request for the gadget-zero stuff to your test_fix_otg_word_loss_regression branch.
```
[ gadget-zero (test_fix_otg_word_loss_regression *)]$ python3 -m unittest  test_gadget0.py 
Running tests for DUT:  stm32f429i-disco
.........ss.................
----------------------------------------------------------------------
Ran 28 tests in 0.359s

OK (skipped=2)
```

[...]

PR #1261, opened 2020-10-25
USB Core 0x2000 Fix

manuelbl:
Commit 26ab78a introduced a fix for issue #873 (*regression of the FIFO word loss issue*) but broke USB completely for all STM32s chips using the DWC USB core with ID 0x2000. This PR is a solution for both the word loss issue in 0x1200 cores as well as the broken USB for 0x2000 cores.

## Analysis

An analysis showed that the cores with ID 0x1200 and 0x200 behave slightly differently when it comes to OUT endpoints. Both are supposed to trigger two consecutive interrupts with two different packet status (see description of OTG_GRXSTSP register):

1. OUT data packet received
2. OUT transfer completed

Since the endpoints are configured for a single packet only, the core will set the NAK flag on the endpoint when a packet is received. On cores with ID 0x1200 the flag *MUST NOT* be cleared after the first interrupt. Otherwise four bytes are lost. On cores with ID 0x2000, the flag *MUST* be cleared after the first interrupt. Otherwise the second interrupt never occurs.

So on chips with core 0x2000, OUT endpoint 0 remained in NAK as the second interrupt, which reset the NAK flag, never occurred.

## Tests

This change has been successfully tested both with *gadget0* and source/sink/loopback test for bulk endpoints that makes heavy use of flow control (`usbd_ep_nak_set`) on different hardware:

- NUCLEO-F042K6
- Chinese BlackPill with STM32F103CB
- Chinese board with STM32F401CC
- NUCELO-F411RE
- NUCELO-F412ZG
- NUCELO-F446RE
- NUCELO-L476RG

The test has been executed with 4 additional PRs installed:
- PR #1259 Fix usb_dwc_common.c late IN usb interrupt flag clearing, #1241
- PR #1258 Fix usb_dwc_common.c endpoint initialization
- PR #1257 USB OTG driver for STM32L4 
- PR #1256 Disable VBUS sensing 

The tests use both USB polling in the main loop (gadget0) and USB processing in the USB interrupt handler (loopback).

## Additional Notes

This PR basically reverts two commits:

- 26ab78a as it broke the USB for cores with ID 0x2000 in the first place
- PR #785 (*Ensure control events are handled in order*) / commit 9b60f303b36d7fb97d0250582755751ddd680a99

If the latter one is still a problem even with this fix applied, it must be solved differently, e.g. by first fetching the RX FIFO and then the TX FIFO (instead of the other way round). Forcing NAK (as done by 9b60f303b36d7fb97d0250582755751ddd680a99) will at best solve part of the problem as the core ignores the NAK flag for SETUP data packets (see description of flag NAKSTS of register OTG_DIEPCTL0).

Most likely, the current implementation of `usbd_ep_nak_set` is not sufficiently robust. If the NAK flag is set between the *OUT data packet received* event and the *OUT transfer completed* event, data might be lost. But that is work for a future PR...

### DIFF
--- a/include/libopencm3/usb/dwc/otg_common.h
+++ b/include/libopencm3/usb/dwc/otg_common.h
@@ -233,6 +233,7 @@
 
 /* OTG FS Product ID register (OTG_CID) */
 #define OTG_CID_HAS_VBDEN	0x00002000
+#define OTG_CID_CNAK_EARLY	0x00002000
 
 /* Device-mode CSRs */
 /* OTG device control register (OTG_DCTL) */
--- a/lib/stm32/common/st_usbfs_core.c
+++ b/lib/stm32/common/st_usbfs_core.c
@@ -252,7 +252,6 @@ void st_usbfs_poll(usbd_device *dev)
 			/* OUT or SETUP? */
 			if (*USB_EP_REG(ep) & USB_EP_SETUP) {
 				type = USB_TRANSACTION_SETUP;
-				st_usbfs_ep_read_packet(dev, ep, &dev->control_state.req, 8);
 			} else {
 				type = USB_TRANSACTION_OUT;
 			}
--- a/lib/usb/usb_control.c
+++ b/lib/usb/usb_control.c
@@ -212,8 +212,6 @@ static void usb_control_setup_write(usbd_device *usbd_dev,
 	} else {
 		usbd_dev->control_state.state = LAST_DATA_OUT;
 	}
-
-	usbd_ep_nak_set(usbd_dev, 0, 0);
 }
 
 /* Do not appear to belong to the API, so are omitted from docs */
@@ -225,8 +223,6 @@ void _usbd_control_setup(usbd_device *usbd_dev, uint8_t ea)
 	(void)ea;
 
 	usbd_dev->control_state.complete = NULL;
-
-	usbd_ep_nak_set(usbd_dev, 0, 1);
 
 	if (usbd_ep_read_packet(usbd_dev, 0, req, 8) != 8) {
 		stall_transaction(usbd_dev);
@@ -299,7 +295,6 @@ void _usbd_control_in(usbd_device *usbd_dev, uint8_t ea)
 		break;
 	case LAST_DATA_IN:
 		usbd_dev->control_state.state = STATUS_OUT;
-		usbd_ep_nak_set(usbd_dev, 0, 0);
 		break;
 	case STATUS_IN:
 		if (usbd_dev->control_state.complete) {
--- a/lib/usb/usb_control.c
+++ b/lib/usb/usb_control.c
@@ -228,6 +228,11 @@ void _usbd_control_setup(usbd_device *usbd_dev, uint8_t ea)
 
 	usbd_ep_nak_set(usbd_dev, 0, 1);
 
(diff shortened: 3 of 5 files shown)


stoyan-shopov:
...
> Most likely, the current implementation of `usbd_ep_nak_set` is not sufficiently robust. If the NAK flag is set between the _OUT data packet received_ event and the _OUT transfer completed_ event, data might be lost. But that is work for a future PR...

This is my suspicion as well, but, to me, that looks like a defficiency of the `usb_dwc_common.c` driver. I have also reverted the `NAK` handling in `usb_control.c` in my branch, if I understand your message. As I understand it, USB `NAK` handling belongs to the upper USB layers, not to the driver code.

From the ST documentation, I understand `OUT data packet received` as: `data packet received by the usb core, *BUT* an ACK handshake not yet sent to the USB host`, and I understand `OUT transfer completed` as: `data packet received by the usb core, *AND* an ACK handshake successfully sent to the USB host`. My gut feeling is, that such behavior is easier to do in hardware - as is the single receive FIFO in the DWC core; but that means, more care from the DWC driver code is needed. Yet, I may be wrong, this is just a guess.

I am working on an STM32F723 device. The ST documentation is not too clear to me, on what consequences setting CNAK may have at arbitrary states of the DWC USB core, and I felt data loss is possible, if setting it between `OUT data packet received` and `OUT transfer completed`.

I did some inspection of the ST sources from the `CUBE MX` package. To be honest, it was difficult for me to follow those sources, they looked much more convoluted to me than the `libopencm3` sources, but otherwise, those sources seem to me that they circumvent this case, and the OUT packets are only delivered to the upper layers of the USB stack only when an `OUT transfer completed` is received.

I am currently working in this branch of `libopencm3`:
https://github.com/libopencm3/libopencm3/compare/master...stoyan-shopov:stm32f723-usbhs-support

This branch is unpolished, and not ready for submitting, but the code modifications work fine for me. One thing to beware of is the code on line 238 in `usb_dwc_common.c` in the branch I am working with - I believe I am not handling something right in the USB interrupt handler, but that workaround works fine for me. Note that the USB driver sources avoid this case, and I could not find any definitive information about this case in the ST documentation.

What I am basically doing, is that I stash the received packets, and only deliver them to the upper USB layers once that the `OUT transfer completed` notification is received. This works fine for me, but it noticeably increases the memory footprint of `libopencm3`, because buffers for the packets received are alloted. My suspicion is, that a single packet buffer may be sufficient, but I did not have time to investigate further.

I am just doing this as a hobby, I am trying to add support for the `stlinkv3-mini` probe to the `blackmagic` firmware, so I am probably taking some dirty shortcuts, but I simply lack the time to act more professionally, and perform proper investigations.

Hopefully, this information helps.

manuelbl:
@stoyan-shopov Where should I put my feedback about your work? I don't think it belongs here.

stoyan-shopov:
> 
> 
> @stoyan-shopov Where should I put my feedback about your work? I don't think it belongs here.

Thank you for your remarks. I have not investigated much deeper, once I could make libopencm3 work good for my application.
You have done a deeper investigation. I am generally not too familiar with USB, it is a complicated protocol.

Well, I believe you may create an issue containing your remarks in my fork of libopencm3 (I have just enabled issues), so that your remarks are visible to others as well:
https://github.com/stoyan-shopov/libopencm3/issues

Thanks, and regards
Stoyan

PR #1262, opened 2020-10-27
usb-dwc: clear STUP interrupt flag

TobleMiner:
Previously the STUP interrupt flag was never cleared.
On the STM32F722 the STUP interrupt flag seems to stops the internal
statemachine of the USB core from reaching the setup finished state.
This renders the USB controller unusable.

This commit ensures the STUP interrupt flag is cleared whenever the
driver is polled.

I do not have a whole lot of experience with DWC usb controllers, thus I'm unable to gauge whether this change breaks exisiting device support. Thus I'd love to see this change tested on as many devices as possible before it is merged,

I've tested this change on a STM32F722RET6 with cdcacm and it works great.

### DIFF
--- a/lib/usb/usb_dwc_common.c
+++ b/lib/usb/usb_dwc_common.c
@@ -350,6 +350,14 @@ void dwc_poll(usbd_device *usbd_dev)
 
 			REBASE(OTG_DIEPINT(i)) = OTG_DIEPINTX_XFRC;
 		}
+
+		/*
+		 * On some controllers the STUP interrupt flag seems to inhibit
+		 * the internal state machine from entering the setup completed
+		 * state again. Thus it must be cleared.
+		 */
+		REBASE(OTG_DOEPINT(i)) = OTG_DOEPINTX_STUP;
+
 	}
 
 	/* Note: RX and TX handled differently in this device. */


manuelbl:
I can test it on several STM32s with DWC cores 0x1200 and 0x2000. (Your STM32F722 has a core ID 0x3000.) Give me about 3 days for it.

At first sight, your bug sounds similar to the one affecting core IDs 0x2000. Have you tried the code in PR #1261? And didn't you have any problems with VBUS sensing?

TobleMiner:
Hi @manuelbl 

I did only notice your PR after opening this one. Very impressive work on your part. I've not tested your PR yet, though. I did not have any issues with VBUS sensing since VBUS is unconnected on my board. I'm currently using the force VBUS and B session valid bits.
I'd appreciate it a lot if you could test this change. Especially since the behavior displayed seems to weird to me.

manuelbl:
@TobleMiner I just wanted to test it on a Nucelo-F722ZE before moving on to the other boards. However, *libopencme* has no USB support for F7 boards. It's missing in the *Makefile* and it's missing in `otg_fs.h`. What other changes than the ones in this PR did you make to get it running?

manuelbl:
I have found that the PRs already in the pipeline also work for the STM32F7:

#1256 Disable VBUS sensing
#1258 Fix usb_dwc_common.c endpoint initialization
#1259 Fix usb_dwc_common.c late IN usb interrupt flag clearing
#1261 USB Core 0x2000 fix

I've added the missing pieces (extend Makefile, extend *ifdefs*) in a new PR: #1264 *USB OTG support for STM32F7*.

So most likely, this PR isn't need anymore. It seems to have the same effect as PR #1261. That's interesting. I'll probably investigate it some more.

PR #1266, opened 2020-11-16
Fix missing Ethernet registers + verify which registers are not available on the STM32F1

BOJIT:
Changes:
- ETH_PTPPPSCR is missing from the register list. Added register, and re-labelled incorrect PTPTSCR labels associated with this missing register.
- Added macro guards around the registers and offsets that are not available on the STM32F1 ethernet hardware.

Sources:
STM32F1: Page 1070 - [https://www.st.com/resource/en/reference_manual/cd00171190-stm32f101xx-stm32f102xx-stm32f103xx-stm32f105xx-and-stm32f107xx-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf]
STM32F4: Page 1236 - [https://www.st.com/resource/en/reference_manual/dm00031020-stm32f405-415-stm32f407-417-stm32f427-437-and-stm32f429-439-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf]
STM32F7: Page 1882 - [https://www.st.com/resource/en/reference_manual/dm00224583-stm32f76xxx-and-stm32f77xxx-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf]

### DIFF
--- a/include/libopencm3/ethernet/mac_stm32fxx7.h
+++ b/include/libopencm3/ethernet/mac_stm32fxx7.h
@@ -9,7 +9,7 @@
  *
  * @author @htmlonly &copy; @endhtmlonly 2013 Frantisek Burian <BuFran@seznam.cz>
  *
- * @date 16 November 2020
+ * @date 1 September 2013
  *
  * LGPL License Terms @ref lgpl_license
  */
@@ -58,10 +58,7 @@
 #define ETH_MACVLANTR			MMIO32(ETHERNET_BASE + 0x1C)
 #define ETH_MACRWUFFR			MMIO32(ETHERNET_BASE + 0x28)
 #define ETH_MACPMTCSR			MMIO32(ETHERNET_BASE + 0x2C)
-/* Following registers not available on F1 */
-#if !defined(STM32F1)
-	#define ETH_MACDBGR			MMIO32(ETHERNET_BASE + 0x34)
-#endif
+#define ETH_MACDBGR			MMIO32(ETHERNET_BASE + 0x34) /* Not on STM32F1 */
 #define ETH_MACSR			MMIO32(ETHERNET_BASE + 0x38)
 #define ETH_MACIMR			MMIO32(ETHERNET_BASE + 0x3C)
 
@@ -93,11 +90,8 @@
 #define ETH_PTPTSAR			MMIO32(ETHERNET_BASE + 0x718)
 #define ETH_PTPTTHR			MMIO32(ETHERNET_BASE + 0x71C)
 #define ETH_PTPTTLR			MMIO32(ETHERNET_BASE + 0x720)
-/* Following registers not available on F1 */
-#if !defined(STM32F1)
-	#define ETH_PTPTSSR			MMIO32(ETHERNET_BASE + 0x728)
-	#define ETH_PTPPPSCR		MMIO32(ETHERNET_BASE + 0x72C)
-#endif
+#define ETH_PTPTSSR			MMIO32(ETHERNET_BASE + 0x728) /* Not on STM32F1 */
+#define ETH_PTPPPSCR		MMIO32(ETHERNET_BASE + 0x72C) /* Not on STM32F1 */
 
 /* Ethernet DMA registers */
 #define ETH_DMABMR			MMIO32(ETHERNET_BASE + 0x1000)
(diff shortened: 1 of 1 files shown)


include/libopencm3/ethernet/mac_stm32fxx7.h
```diff
@@ -9,7 +9,7 @@
  *
  * @author @htmlonly &copy; @endhtmlonly 2013 Frantisek Burian <BuFran@seznam.cz>
  *
- * @date 1 September 2013
+ * @date 16 November 2020
```
karlp, line 12: these are original dates, we don't change them

include/libopencm3/ethernet/mac_stm32fxx7.h
```diff
@@ -58,8 +58,10 @@
 #define ETH_MACVLANTR			MMIO32(ETHERNET_BASE + 0x1C)
 #define ETH_MACRWUFFR			MMIO32(ETHERNET_BASE + 0x28)
 #define ETH_MACPMTCSR			MMIO32(ETHERNET_BASE + 0x2C)
-/* not available on F1 ?*/
-#define ETH_MACDBGR			MMIO32(ETHERNET_BASE + 0x34)
+/* Following registers not available on F1 */
+#if !defined(STM32F1)
```
karlp, line 62: we haven't normally done much work trying to prevent this sort of thing unless htere's been conflicts, as far as the _definitions_ go.  at least, the comment becomes pretty irrelevant :)

include/libopencm3/ethernet/mac_stm32fxx7.h
```diff
@@ -154,7 +163,10 @@
 
 #define ETH_MACCR_JD			(1<<22)
 #define ETH_MACCR_WD			(1<<23)
-#define ETH_MACCR_CSTF			(1<<25)
+/* Following registers not available on F1 */
```
karlp, line 166: the comment here is wrong, if you think it should be kept a tall.

include/libopencm3/ethernet/mac_stm32fxx7.h
```diff
@@ -247,45 +259,48 @@
...
+	#define ETH_MACDBGR_MMRPEA		(1<<0)
+	#define ETH_MACDBGR_MSFRWCS		(3<<1)
+	#define ETH_MACDBGR_RFWRA		(1<<4)
+
+	#define ETH_MACDBGR_RFRCS_SHIFT		5
+	#define ETH_MACDBGR_RFRCS		(3<<ETH_MACDBGR_RFRCS_SHIFT)
+	#define ETH_MACDBGR_RFRCS_IDLE		(0<<ETH_MACDBGR_RFRCS_SHIFT)
+	#define ETH_MACDBGR_RFRCS_RDATA		(1<<ETH_MACDBGR_RFRCS_SHIFT)
+	#define ETH_MACDBGR_RFRCS_RSTAT		(2<<ETH_MACDBGR_RFRCS_SHIFT)
+	#define ETH_MACDBGR_RFRCS_FLUSH		(3<<ETH_MACDBGR_RFRCS_SHIFT)
```
karlp, line 273: please don't change the style like this, nowhere else do we have indented defines.

include/libopencm3/ethernet/mac_stm32fxx7.h
```diff
@@ -247,45 +259,48 @@
...
-#define ETH_MACDBGR_TFRS_IDLE		(0<<ETH_MACDBGR_TFRS_SHIFT)
-#define ETH_MACDBGR_TFRS_READ		(1<<ETH_MACDBGR_TFRS_SHIFT)
-#define ETH_MACDBGR_TFRS_WAIT		(2<<ETH_MACDBGR_TFRS_SHIFT)
-#define ETH_MACDBGR_TFRS_FLUSH		(3<<ETH_MACDBGR_TFRS_SHIFT)
-
-#define ETH_MACDBGR_TFWA		(1<<22)
-#define ETH_MACDBGR_TFNE		(1<<24)
-#define ETH_MACDBGR_TFF			(1<<25)
+/* Following registers not available on F1 */
+#if !defined(STM32F1)
```
karlp, line 263: if you switch this to checking on the definition of hte register, it will automatically work if/when more parts with /without the registers appear.  If this is actually a lot, we might want to add it as an extra header

include/libopencm3/ethernet/mac_stm32fxx7.h
BOJIT: Should I just change all of these exceptions to notes rather than checking if STM32F1 is defined?

As far as I can tell, there are no register conflicts between the F1 and F4/7 peripheral registers. In fact, I think it's possible that the registers may work on the F1 and they are just not documented. In every other respect the hardware is identical, but I only have F4/F7 here to test.

This PR is only fixing one register definition that is actually wrong, the rest is just flagging where certain registers and register locations are not documented in the STM32F1 reference manual. As I have no way of testing whether these undocumented registers work or not on F1 targets, is it worth me just restricting this PR to the one incorrect register and removing the `#if !defined(STM32F1)` parts?

karlp: Sounds good to me.

include/libopencm3/ethernet/mac_stm32fxx7.h
```diff
@@ -91,8 +90,8 @@
 #define ETH_PTPTSAR			MMIO32(ETHERNET_BASE + 0x718)
 #define ETH_PTPTTHR			MMIO32(ETHERNET_BASE + 0x71C)
 #define ETH_PTPTTLR			MMIO32(ETHERNET_BASE + 0x720)
-/* not available on F1 ?*/
-#define ETH_PTPTSSR			MMIO32(ETHERNET_BASE + 0x728)
+#define ETH_PTPTSSR			MMIO32(ETHERNET_BASE + 0x728) /* Not on STM32F1 */
+#define ETH_PTPPPSCR		MMIO32(ETHERNET_BASE + 0x72C) /* Not on STM32F1 */
```
karlp, line 94: heh, I had fun valdating this, in my revision of the f7 ref man, it's in the regster section, but not the register _map_ section!

karlp:

Issue #1265, opened 2020-11-06
Either avoid using `python` (without version) or suggest using a virtualenv.


sumpfralle:
I posted the following text one week ago as [comment for #1075](#1075#718045397). But that issue was already closed for some time, thus my comment was probably more or less invisible. Feel free to close this new issue and re-open the old one, if you think, that the discussion should continue there instead of here.

Summary: I would like to re-open the discussion regarding the most suitable python-shebang in the four python scripts used in libopencm3. The previous discussion (#1075) seems to contain a few misinterpretations of the relevant [PEP-0394](https://www.python.org/dev/peps/pep-0394/), which I would like to clarify below in order to smoothen the usage experience of libopencm3.

> According to https://www.python.org/dev/peps/pep-0394/ we should keep using "python" if we want to support both 2 and 3, and just make sure our script works with both, which I believe it does, but feel free to correct me on that.

Indeed [PEP-394](https://www.python.org/dev/peps/pep-0394/#for-python-script-publishers) allows `#!/usr/bin/env python`, but it should *only be used in a virtualenv*:

> For scripts that are only expected to be run in an activated virtual environment, shebang lines can be written as #!/usr/bin/env python, as this instructs the script to respect the active virtual environment.

I guess, that we do not really expect the users of libopencm3 to run `make` (for compiling libopencm3) in a python virtualenv, or?

PEP-394 further says:

> Scripts targeting both “old systems” and systems without the default python command need to make a compromise and document this situation. Avoiding shebangs (via the console_scripts Entry Points ([9]) or similar means) is the recommended workaround for this problem.

Thus the minimum effort recommended by PEP-394 would be to document, that `make` should be typed in a Python virtualenv. Other approaches could be a new pre-compilation step converting the input python scripts into *locally usable* python scripts with a proper (detected) shebang. But this feels unreasonably cumbersome.

> If fedora is breaking the PEP by simply having no python executable, then, respectfully, I feel that's a fedora bug. (And thanks for hitting it before me, I'm still on fed29)

No, I think, Fedora's removal of `python` does not break PEP-394, but follows its wording perfectly.
See one of the [proposed options](https://www.python.org/dev/peps/pep-0394/#for-python-runtime-distributors):

> Distributors may choose to set the behavior of the python command as follows:
> * ...
> * not provide python command,
> * ...

[...]

PR #1269, opened 2020-11-24
Update python references to python3

sumpfralle:
Python2 is end-of-life [1] since the 1st of January 2020.

Some distributions (most notably: Debian and its derivatives) will stop
providing a `python` executable in order to encourage users to specify
the interpreter language of local scripts explicitly.  Users of such
environments will be forced to work around this in one of these ways:

* create a virtual environment or
* manipulate the shebangs of the scripts or
* install the python2 package (as long as it is provided by distributions)

All currently maintained distribution releases provide python3.
In the near future distributions will need to remove python2, since it
is not maintained anymore.

PEP-394 [2] recommends to reference a specific python version
(python2 or python3), if the script is not expected to run in a virtual
environment.

Closes: #1265

[1] https://www.python.org/dev/peps/pep-0373/#update-april-2014
[2] https://www.python.org/dev/peps/pep-0394/#for-python-script-publishers

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,12 @@ Building requires Python (some code is generated).
  Download and install:
 
  - msys - http://sourceforge.net/projects/mingw/files/MSYS/Base/msys-core/msys-1.0.11/MSYS-1.0.11.exe
- - Python - http://www.python.org/ftp/python/2.7/python-2.7.msi (any 2.7 release)
+ - Python - https://www.python.org/downloads/windows/ (any release)
  - arm-none-eabi/arm-elf toolchain (for example this one https://launchpad.net/gcc-arm-embedded)
 
-Run msys shell and set the path without standard Windows paths, so Windows programs such as 'find' won't interfere:
+Run msys shell and set the path without standard Windows paths (adjusting to your version of Python), so Windows programs such as 'find' won't interfere:
 
-    export PATH="/c//Python27:/c/ARMToolchain/bin:/usr/local/bin:/usr/bin:/bin"
+    export PATH="/c//Program Files/Python 3.9:/c/ARMToolchain/bin:/usr/local/bin:/usr/bin:/bin"
 
 After that you can navigate to the folder where you've extracted libopencm3 and build it.
 
--- a/scripts/data/lpc43xx/csv2yaml.py
+++ b/scripts/data/lpc43xx/csv2yaml.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 import sys
 import yaml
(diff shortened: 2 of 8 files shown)


scripts/data/lpc43xx/csv2yaml.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
karlp, line 1: did you test this on py3? this file hasn't been touched in ~forever, and isn't guaranteed to be python3 safe

scripts/data/lpc43xx/gen.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
karlp, line 1: same here, lpc43xx is ~unmaintained

scripts/lpcvtcksum
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
```
karlp, line 1: and this one, all the others look good

tests/gadget-zero/Jenkinsfile
```diff
@@ -64,7 +64,7 @@ pipeline {
         		. .env3/bin/activate
         		cd tests/gadget-zero
         		rm -rf tests
-        		python test_gadget0.py -X
+        		python3 test_gadget0.py -X
```
karlp, line 67: this actually runs in a venv, see line 26 and 64, so this is "fine" but unnecessary

README.md
```diff
@@ -55,12 +55,12 @@ Building requires Python (some code is generated).
...
  - msys - http://sourceforge.net/projects/mingw/files/MSYS/Base/msys-core/msys-1.0.11/MSYS-1.0.11.exe
- - Python - http://www.python.org/ftp/python/2.7/python-2.7.msi (any 2.7 release)
+ - Python - https://www.python.org/downloads/windows/ (any release)
  - arm-none-eabi/arm-elf toolchain (for example this one https://launchpad.net/gcc-arm-embedded)
 
-Run msys shell and set the path without standard Windows paths, so Windows programs such as 'find' won't interfere:
+Run msys shell and set the path without standard Windows paths (adjusting to your version of Python), so Windows programs such as 'find' won't interfere:
 
-    export PATH="/c//Python27:/c/ARMToolchain/bin:/usr/local/bin:/usr/bin:/bin"
+    export PATH="/c//Program Files/Python 3.9:/c/ARMToolchain/bin:/usr/local/bin:/usr/bin:/bin"
```
karlp, line 63: thank you for doing this so nicely!

karlp:
I've pulled this locally with changes to the lpc43xx stuff as mentioned, there's no need for you to do anything at this stage, thanks!

sumpfralle:
Thank you for taking a look at my proposal!

Indeed I overlooked two incompatibilities with python3 in the scripts you mentioned. These are fixed now.

In general all of these scripts are rather trivial and do no use features, that could be problematic for a python3 conversion. Thus I am quite confident, that even the rarely used (and thus not easily testable) scripts are fine.

Most of the scripts deviate quite significantly from the recommended python code (see `python3 -m flake8 FILENAME`). I would be happy to clean these up (and maybe add a style check for travis), if you think, it is worth the (small) effort. Just tell me ...
(and maybe adding `codespell` to travis, too)

karlp:
landed in master now thanks!

PR #1274, opened 2020-12-10
stm32g4: Add support for OPAMP

ben-brewer-codethink:
Move implementations into common and split into v1 and v2.


karlp:
looks good, merged thanks.

PR #1275, opened 2020-12-11
stm32g4: add SPI

kplus:
The SPI peripheral on G4 is identical to F3, this patch copies the
header files directly from F3


karlp:
Merged thanks.  I dropped the misleading/wrong 2012 date line

PR #1281, opened 2020-12-16
stm32/dac: Update API to support multiple DACs and add v2 support

ben-brewer-codethink:
This patch should enable use of v2 DACs and also support the F3's multiple v1 DAC at the cost of breaking the API. However the new interface is more in-keeping with interfaces for other peripherals now as was agreed in #libopencm3 on freenode.

Patch to the examples repository to follow.


ben-brewer-codethink:
Example fixes can be found here: https://github.com/libopencm3/libopencm3-examples/pull/218

include/libopencm3/stm32/common/dac_common_all.h
```diff
@@ -322,105 +132,114 @@ Unmask bits [(n-1)..0] of LFSR/Triangle Amplitude equal to (2**(n+1)-1
...
-void dac_set_trigger_source(uint32_t dac_trig_src);
-void dac_set_waveform_generation(uint32_t dac_wave_ens);
-void dac_disable_waveform_generation(data_channel dac_channel);
-void dac_set_waveform_characteristics(uint32_t dac_mamp);
-void dac_load_data_buffer_single(uint16_t dac_data, data_align dac_data_format,
-				 data_channel dac_channel);
-void dac_load_data_buffer_dual(uint16_t dac_data1, uint16_t dac_data2,
-			       data_align dac_data_format);
-void dac_software_trigger(data_channel dac_channel);
+void dac_enable(uint32_t dac, uint8_t channel);
```
karlp, line 226: hrmm, you _probably_ just want "int" for the channel, there's no need for it to be explicitly 8bit, which can require more work

include/libopencm3/stm32/f0/dac.h
```diff
@@ -33,37 +33,20 @@
...
 #define LIBOPENCM3_DAC_H
 
-#include <libopencm3/stm32/common/dac_common_all.h>
+#include <libopencm3/stm32/common/dac_common_v1.h>
 
 /*****************************************************************************/
 /* Module definitions                                                        */
 /*****************************************************************************/
 
 #define DAC				DAC_BASE
```
karlp, line 42: seeing as we're breaking things, do we wnat to move this to DAC1 too?

karlp: (Or at least, DAC _and_ DAC1?  (we'v edone that with ADC for  many parts already too)

ben-brewer-codethink: I guess so yeah, DAC1 makes more sense as it makes it more obvious that the API is designed to support multiple rather than it being repetitive boilerplate, you agree?

include/libopencm3/stm32/common/dac_common_all.h
ben-brewer-codethink: Ok, will fix.

include/libopencm3/stm32/f0/dac.h
ben-brewer-codethink: Ok, switched to just DAC1, no need for backward compat given we're breaking the API anyway and the DAC parameter was previously unused.

karlp:
I'm  just merging this now, where did you get "DAC5" from for g4?

karlp:
Also, you appear to have lost the MAMPxxx definitions? are these common or what?  It looks the same for L1 (old) and G4 (new) so I've restored them to the "common_all" header, but let me know if that's wrong.

karlp:
actually, after looking at them, fuck that, I'm going to fix teh documentation references that tried to use them :)

karlp:
actually, I think the "set waveform characteristics" function probably needs a new API, like...

set_waveform_char(dac, mamp1, mamp2)
and that can handle the correct shifting and masking, instead of this gross "jsut OR whatever" we have now...

karlp:
See https://github.com/libopencm3/libopencm3/pull/1297 for a new api, feedback welcome before I slam it in...

karlp:
This (now merged) PR also breaks the dac_set_waveform_generation() funciton, as the enums aren't viable for use as is.  ~same problem as with the waveform charactersitics really.  One of us has to handle it ;)

ben-brewer-codethink:
> I'm just merging this now, where did you get "DAC5" from for g4?

Looks like a mistake, I guess I still had ADC's in my head.

ben-brewer-codethink:
> Also, you appear to have lost the MAMPxxx definitions? are these common or what? It looks the same for L1 (old) and G4 (new) so I've restored them to the "common_all" header, but let me know if that's wrong.

I removed the MAMP defines since they were just defining numbers with no meaning, but it's fine for them to stay too.

karlp:
yeah, dropping the mamp defintions was fine, just some doc refs broke, and it showed me that the api was bad.  waveform generation has a similar problem

ben-brewer-codethink:
> yeah, dropping the mamp defintions was fine, just some doc refs broke, and it showed me that the api was bad. waveform generation has a similar problem

Anything I can do to help?

karlp:
@ben-brewer-codethink sure! have a review of https://github.com/libopencm3/libopencm3/pull/1297 for me?

PR #1285, opened 2020-12-26
Refactored genlink.py to expose some kind of device info API

Maldus512:
When developing STM32 firmware I like to structure my Projects with Scons as the build system instead of Make.
The integration with libopencm3 is quite straightforward since the default Makefile compiles a static library that I can then link easily into my binary.

I'd also like to access to the same compilation flags and maybe generate the linker script on my own. For this Libopencm3 uses `genlink.py`, and it works perfectly even outside of Make. However, since Scons is basically a Python framework it would be better to import the script instead of invoking it as a shell command. I changed the internals of `genlink.py` to expose a bare Python API to get the device information.

For example, while before I had to invoke a separate Python process and read its output:
```
from subprocess import check_output

(FAMILY, CPU, FPU) = tuple(
    map(
        lambda x: check_output([
            f'{LIBOPENCM3}/scripts/genlink.py',
            f'{LIBOPENCM3}/ld/devices.data', DEVICE, x
        ]).decode('utf-8'), ['FAMILY', 'CPU', 'FPU']))
``` 

Now I can just import the script and use its functions:
```
sys.path.insert(1, f'{LIBOPENCM3}/scripts')
from genlink import get_device_family, get_device_info

data_file = f'{LIBOPENCM3}/ld/devices.data'
FAMILY = get_device_family(data_file, DEVICE)
(CPU, FPU) = tuple(map(lambda x: get_device_info(data_file, DEVICE, x), ['CPU', 'FPU']))
```

This refactoring should leave the previous functioning of the generator untouched; I just moved the information extraction into importable functions.

### DIFF
--- a/scripts/genlink.py
+++ b/scripts/genlink.py
@@ -23,103 +23,162 @@ import sys
 import re
 import os
 
-if len(sys.argv) != 4:
-    print("usage: %s <device-data-file> <device> <mode>" % sys.argv[0], file=sys.stderr)
-    sys.exit(1)
 
-data_file_path = sys.argv[1]
-find = sys.argv[2].lower()
-mode = sys.argv[3].upper()
+def get_device_data(data_file_path, device_id):
+    device = {
+        'info': {},
+        'defs': [],
+        'family': [],
+    }
 
-device = {
-    'info': {},
-    'defs': [],
-    'family': [],
-}
+    # open device data file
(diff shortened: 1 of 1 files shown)


karlp:
This looks like a very sensible idea :)

karlp:
merged via rebase in 0817f917 thanks!

Issue #1263, opened 2020-10-29
Building libopencm3 for a specific target


krish2487:
Hello,
I know this look like a rather silly issue.
A while ago, I was able to compile libopencm3 for a specific stm32 family (F1 to be precise)  to keep the footprint low. 
I do not remember for the life of me, what I make file target I gave it.

Can anyone pitch in and help me with this?

To be clear, I am using the libopencm3 template and would like to build the libopencm3 library - but only for the STM32 F1 family. not the rest. I know it can be done and I have done it before from the makefile in the libopencm3 top level directory.. but I do not remember how..

krish2487:
Never mind,

As is the case, I figured out how to do so a couple of minutes after I posted the question.
🤦 

For anyone else who might want to do the same, this is how.

` make $TARGETS=stm32/f1`

I am closing the issue now.

karlp:
No, it' sok, we should add that to the front readme page under "building" seeing as we support it.

karlp:
fixed via the linked issue being merged

PR #1287, opened 2020-12-29
README: added info about partial build

a-v-v:
This commit resolves issue #1263.


a-v-v:
I can close this pull request, if it is unecessary.

karlp:
no, it fixes an issue, there's no reason to close it.

schodet:
There seems to be no way to push additional commits to someone else PR, I made some suggestion here:

https://github.com/schodet/libopencm3/tree/readme_build

- split white space change in another commit,
- remove list of targets from README to avoid duplicated work and add a list-targets in Makefile,
- replace “footprint of the library” by “build time”, as someone could wrongly understand that the compiled size of the lib depends on the list of built targets.

karlp:
resolved via ed2aada3..8f016208 thanks!

PR #1290, opened 2020-12-30
STM32: Use const qualifier in i2c_transfer7

jeanthom:
The write buffer (data being sent from the I2C controller to the I2C peripheral) is read-only, so I thought it would be nicer if we made it const.

### DIFF
--- a/include/libopencm3/stm32/common/i2c_common_v1.h
+++ b/include/libopencm3/stm32/common/i2c_common_v1.h
@@ -415,7 +415,7 @@ void i2c_enable_dma(uint32_t i2c);
 void i2c_disable_dma(uint32_t i2c);
 void i2c_set_dma_last_transfer(uint32_t i2c);
 void i2c_clear_dma_last_transfer(uint32_t i2c);
-void i2c_transfer7(uint32_t i2c, uint8_t addr, uint8_t *w, size_t wn, uint8_t *r, size_t rn);
+void i2c_transfer7(uint32_t i2c, uint8_t addr, const uint8_t *w, size_t wn, uint8_t *r, size_t rn);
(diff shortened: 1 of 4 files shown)


karlp:
I'm not generally opposed to this sort of thing, but it seems very incomplete.  What about the the i2c periph itself, what about the write size? what about the read size? what about the writebuffer you're passing down to the next function?  I generally feel that this sort of "constifying" projects are only valuable when they're complete, so you can actually rely on it.  I'm not.... realllly convinced that this is the sort of project that should just be done one parameter at a time?

jeanthom:
Hi,

>  What about the the i2c periph itself
It doesn't matter here. Both v1 and v2 functions are **NOT** feeding a pointer to the peripheral, instead each byte of the write buffer is read by the C code and **then** fed into the peripheral. You can check for yourself in the source files of each function.

> I generally feel that this sort of "constifying" projects are only valuable when they're complete, so you can actually rely on it. I'm not.... realllly convinced that this is the sort of project that should just be done one parameter at a time?
Could you elaborate on that? I don't understand what you mean. I'm only adding the const qualifier to one public function, and because we have two similar implementations of that function, I thought it would be more sensible to edit both functions at once instead of having mixed API depending on the version of the peripheral.

emersion:
You can also think of `const` as part of the docs: it's easier to understand how a function will handle parameters if some of them have a `const` qualifier.

>what about the write size? what about the read size?

`const` should only be used with arguments that are pointers. Values that aren't pointers are copied when the function is called anyways.

jeanthom:
Ping!

karlp:
ok ok, this is merged in a9cc6953  this seems like a trivial thing to have so many of you pushing on, but fine, we'll const correctness this single api right now.  continuous improvement sure.... always on the road forwards.

jeanthom:
Thanks!

PR #745, opened 2017-02-14
rewrite usb_msc.c

cupkee:
support write

PR #1294, opened 2021-01-19
MSC enhancements

dhoepfl:
I implemented some enhancements to the MSC (mass storage class) implementation. Some changes are inspired by [this article](https://medium.com/@ly.lee/stm32-blue-pill-usb-bootloader-how-i-fixed-the-usb-storage-serial-dfu-and-webusb-interfaces-36d7fe245b5c). I hope it is okay to have several changes in one pull request but since later commits depend on earlier ones, I decided to got with it. The changes I made are:

 * d6ac26b: Adds handling for vital product data pages. This omission was already marked as TODO. VPD-pages are mandatory according to SCSI and not supporting them caused long timeouts on Windows 7 when mounting my device. The commit also includes a dummy implementation of the unit serial number VPD-page. While not mandatory by the SCSI standard, not having it, too, caused delays when mounting the device.
 * 4430735: Adds support for the READ FORMAT CAPACITIES scsi command. Again, Windows wants this call to work, so I implemented a minimal version.
 * 89d0841: Not strictly mandatory but if you want to build a MSC device that is removable, you need to support the PREVENT ALLOW MEDIUM REMOVAL command. The implementation accepts all medium lock calls (and exposes the current state to the device firmware). There is no callback support though.
 * 0affe8f: Implements handling of START STOP UNIT, that is the second part of removable MSC support. This call includes a callback that can be used to set the medium loaded state. Setting the medium to not loaded also changes the device status to reflect this state.

Combining all of these changes enables you to write a MSC device that mounts (quickly) on all OSes I could test and that also correctly ejects itself when you eject it in the OS.

### DIFF
--- a/include/libopencm3/usb/msc.h
+++ b/include/libopencm3/usb/msc.h
@@ -78,6 +78,12 @@ typedef struct _usbd_mass_storage usbd_mass_storage;
 #define USB_MSC_REQ_BULK_ONLY_RESET	0xFF
 #define USB_MSC_REQ_GET_MAX_LUN		0xFE
 
+/* SPC-2, 7.12 PREVENT ALLOW MEDIUM REMOVAL, Table 77 */
+#define USB_MSC_SPC_2_MEDIUM_EJECT_LOCKED_NONE		0x00
+#define USB_MSC_SPC_2_MEDIUM_EJECT_LOCKED_REMOTE	0x01
+#define USB_MSC_SPC_2_MEDIUM_EJECT_LOCKED_DEVICE	0x02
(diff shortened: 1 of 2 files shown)


karlp:
Are these meant to replace, or complement some of the existing MSC PRs?

dhoepfl:
I found two related pull requests:

#923 which I guess gets replaced by this pull request.
#745 seems to be a complete rewrite of the MSC code. My pull request is unrelated.

stefaandesmet2003:
This PR is useful. Can it be merged please?

PR #1297, opened 2021-01-24
stm32: dac: new api for waveform characteristics

karlp:
Old API required users to manually construct bit maps frm opaquely named
defines, with little help.  It also was a pure OR operation, with no way
to ever clear bits.

### DIFF
--- a/include/libopencm3/stm32/common/dac_common_all.h
+++ b/include/libopencm3/stm32/common/dac_common_all.h
@@ -103,9 +103,11 @@ specific memorymap.h header before including this header file.*/
 /** MAMP2[3:0]: DAC channel2 mask/amplitude selector */
 #define DAC_CR_MAMP2_SHIFT		24
 
-/** WAVE2[1:0]: DAC channel2 noise/triangle wave generation enable */
+/** Wave generation mode mask size */
+#define DAC_CR_WAVEx_MASK		0x3
+
+/** WAVE2[1:0]: DAC channel2 wave generation mode*/
 #define DAC_CR_WAVE2_SHIFT		22
-#define DAC_CR_WAVE2_MASK		0x3
 
 /** EN2: DAC channel2 enable */
 #define DAC_CR_EN2			(1 << 16)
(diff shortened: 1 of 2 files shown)


ben-brewer-codethink:
Looks much better to me.

karlp:
Hrm, so I currently still have waveform_char and waveform_gen using two different styles.  I think they should be consistent, either...
```
// set single
dac_set_waveform_characteristics(DAC1, 0x123, 0);
dac_set_waveform_generation(DAC1, DAC_WAVE_TRIANGLE, DAC_WAVE_DISABLE);
// set two different
dac_set_waveform_characteristics(DAC1, 0x123, 0x3f);
dac_set_waveform_generation(DAC1, DAC_WAVE_TRIANGLE, DAC_WAVE_NOISE);
// set two same
dac_set_waveform_characteristics(DAC1, 0x69, 0x69);
dac_set_waveform_generation(DAC1, DAC_WAVE_TRIANGLE, DAC_WAVE_TRIANGLE);
```
or
```
// set single
dac_set_waveform_characteristics(DAC1, DAC_CHANNEL1, 0x123);
dac_set_waveform_generation(DAC1, DAC_CHANNEL1, DAC_WAVE_TRIANGLE);
// set two different
dac_set_waveform_characteristics(DAC1, DAC_CHANNEL1, 0x123);
dac_set_waveform_generation(DAC1, DAC_CHANNEL1, DAC_WAVE_TRIANGLE);
dac_set_waveform_characteristics(DAC1, DAC_CHANNEL2, 0x43);
dac_set_waveform_generation(DAC1, DAC_CHANNEL2, DAC_WAVE_NOISE);
// set two same
dac_set_waveform_characteristics(DAC1, DAC_CHANNEL_BOTH, 0x69);
dac_set_waveform_generation(DAC1, DAC_CHANNEL_BOTH, DAC_WAVE_TRIANGLE);

```

The former is, I think, a little nicer, even if you are required to consider the other channel all the time.  The "with channel" API lets you only set one channel if you only want to consider one channel, but it's more code in the library, and is it really any better?

ChuckM:
My question would be this, when used in practice are you always talking to both channels? Or is it common to talk to just one? The latter suggests that the second API would add code but be easier in the general case.  I suppose varargs if off the  table right? Basically the "hit" here is that when you want to set two different wave forms you end up setting each individually vs being able to do it in one call. If that is the "common" case then #1 is better, but if setting one or the other is the common case then #2 would be better. Helpful?

karlp:
not really :)  I don't know what is most common.  "with channel" is probably most easy to read after the fact, even if it's wordier, and is more code.  "always both" is compact, but not as obvious what's happening, and you're _always_ provideing two channels, wehther you want it or not.

karlp:
merged the "with channel" form for both..

PR #1298, opened 2021-01-28
g4:flash add missing defines

ChuckM:
The flash defines were missing a couple so they are added in this commit

### DIFF
--- a/include/libopencm3/stm32/g4/flash.h
+++ b/include/libopencm3/stm32/g4/flash.h
@@ -73,6 +73,10 @@
 #define FLASH_ACR_DBG_SWEN		(1 << 18)
 #define FLASH_ACR_SLEEP_PD		(1 << 14)
 #define FLASH_ACR_RUN_PD		(1 << 13)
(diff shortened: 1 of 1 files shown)


include/libopencm3/stm32/g4/flash.h
```diff
@@ -73,6 +73,10 @@
 #define FLASH_ACR_DBG_SWEN		(1 << 18)
 #define FLASH_ACR_SLEEP_PD		(1 << 14)
 #define FLASH_ACR_RUN_PD		(1 << 13)
+#define FLASH_ACR_DCRST			(1 << 12)
+#define FLASH_ACR_ICRST			(1 << 11)
+#define FLASH_ACR_DCEN			(1 << 10)
+#define FLASH_ACR_ICEN			(1 << 9)
```
karlp, line 79: these are in flash_common_idcache.h, which is included by the g4 header?  Are you sure you need this?

karlp:
after conversations on IRC, this was fixed by simply updating locally.

PR #1301, opened 2021-01-30
genlink: Provide LIBNAME even if library isn't built yet

DanielG:
Currently genlink-config.mk tries to look for the device library by first
looking for the family and then subfamily suffix. This has two problems:

1) If the library wasn't built yet we get a warning and LIBNAME isn't
set. This makes building libopencm3 automatically as part of another
project's makefile uneccesarily difficult as LIBNAME has to be hardcoded as a
workaround.

2) In the special case of the lpc43xx_m0 "device" the correct LIBNAME is
actually the one with subfamily suffix, but currently we prefere the family
one.

Both of these issues are fixed by makeing the canonical LIBNAME part of the
genlink's database. When asking genlink for a LIBNAME we default to the
family suffix so as to not require annotating evey device manually but
allow overriding this by specifying a LIBNAME as a definition in
devices.data.

Currently lpc43xx_m0 is the only odd one out that requires this though.

---
This was submitted as part of #1137 previously. Commit d60d7802 seems to have fixed some of the GENHDR problems since I submitted my fix for that.

Tested by building libopencm3-template and libopencm3-examples with this. Still seems to work fine apart from a small LDSCRIPT fix due to e15071dbf8 I assume.

### DIFF
--- a/ld/devices.data
+++ b/ld/devices.data
@@ -509,7 +509,7 @@ lpc13u lpc13xx USBRAM_OFF=0x20004000
 lpc17[56]x lpc17xx RAM1_OFF=0x2007C000 RAM2_OFF=0x20080000
 lpc17[78]x lpc17xx RAM1_OFF=0x20000000 RAM2_OFF=0x20004000
 
-lpc43xx_m0 lpc43xx CPU=cortex-m0 FPU=soft
+lpc43xx_m0 lpc43xx CPU=cortex-m0 FPU=soft LIBNAME=lpc43xx_m0
 lpc43xx_m4 lpc43xx CPU=cortex-m4 FPU=hard-fp4-sp-d16
 
 ################################################################################
--- a/mk/genlink-config.mk
+++ b/mk/genlink-config.mk
@@ -25,6 +25,7 @@ LDSCRIPT	= generated.$(DEVICE).ld
 DEVICES_DATA = $(OPENCM3_DIR)/ld/devices.data
 
 genlink_family		:=$(shell $(OPENCM3_DIR)/scripts/genlink.py $(DEVICES_DATA) $(DEVICE) FAMILY)
+genlink_libname		:=$(shell $(OPENCM3_DIR)/scripts/genlink.py $(DEVICES_DATA) $(DEVICE) LIBNAME)
 genlink_subfamily	:=$(shell $(OPENCM3_DIR)/scripts/genlink.py $(DEVICES_DATA) $(DEVICE) SUBFAMILY)
 genlink_cpu		:=$(shell $(OPENCM3_DIR)/scripts/genlink.py $(DEVICES_DATA) $(DEVICE) CPU)
 genlink_fpu		:=$(shell $(OPENCM3_DIR)/scripts/genlink.py $(DEVICES_DATA) $(DEVICE) FPU)
@@ -52,18 +53,7 @@ ifeq ($(genlink_family),)
 $(warning $(DEVICE) not found in $(DEVICES_DATA))
 endif
 
-# only append to LDFLAGS if the library file exists to not break builds
-# where those are provided by different means
-ifneq (,$(wildcard $(OPENCM3_DIR)/lib/libopencm3_$(genlink_family).a))
-LIBNAME = opencm3_$(genlink_family)
-else
-ifneq (,$(wildcard $(OPENCM3_DIR)/lib/libopencm3_$(genlink_subfamily).a))
-LIBNAME = opencm3_$(genlink_subfamily)
-else
-$(warning $(OPENCM3_DIR)/lib/libopencm3_$(genlink_family).a library variant for the selected device does not exist.)
-endif
-endif
-
+LIBNAME ?= opencm3_$(genlink_libname)
 LDLIBS += -l$(LIBNAME)
 LIBDEPS += $(OPENCM3_DIR)/lib/lib$(LIBNAME).a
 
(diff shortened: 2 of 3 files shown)


DanielG:
Note this causes conflicts with #1285 currently.

DanielG:
@karlp ping. Review would be appreciated. I just ran into the issue this PR fixes again, gah.

karlp:
This looks reasonable, let me play with a bit and see if I'm missing any history of why it's like it is.

karlp:
So, what exactly is this solving again?  I've got a project locally that uses the genlink-config and genlink-rules, and with or without this, you still need to build the library first by hand, or things like nvic generated files don't exist.

karlp:
I've modified the "rules.mk" that's used in the gadget zero tests, for isntance, to build the LIBNAME on demand, but it will still never rebuild the library if it's touched internally.  That's a whole soup of further dependency tracking that's kinda tangential to building a .a library in the first place.  So, yeah, this... doesn't do any particular harm beyond yet another re-parsing of the file but I'm also not entirely sure what it saves still

DanielG:
Right so I have a couple of changes to rules.mk in my project to make this all fit together.

Indeed you are right that internal dependencies aren't handled by this approach but that really isn't a big deal for the downstream use-case. Ideally I'd like to be able to include lcm3's makefiles into my own or include some *.d files to make this work but that's a different can of worms. baby steps.

The main pain point is LIBNAME not getting set when the library doesn't exist yet. That presents a staging problem as I can't (easily) build it before calling my build system just to know which library to build in the first place :)

karlp:
I'm still not seeing that though.  You just do "$(MAKE) -C $(OPENCM3_DIR)"  and it's done....  you can even make that in your bootstrap scripts, libopencm3 doesn't have make targets that build based on .a names.

I'm not seeing yet how this is _actually_ blocking you in the way that you seem to be making it out to be?  Can you link a project that "breaks" or "is unfriendly" or something that you needed this for?

DanielG:
Take for example libopencm3-template, it has this very same problem:
```
$ git clone https://github.com/libopencm3/libopencm3-template.git
$ cd libopencm3-template/
$ git clone https://github.com/libopencm3/libopencm3.git libopencm3/ # use master rather than ancient submodule
$ make -C my-project
make: Entering directory '/tmp/libopencm3-template/my-project'
../libopencm3/mk/genlink-config.mk:63: ../libopencm3/lib/libopencm3_stm32f4.a library variant for the selected device does not exist.
  CC	my-project.c
  CC	../my-common-code/api.c
  AS	../my-common-code/api-asm.S
  GENLNK  stm32f407vgt6
make: *** No rule to make target '../libopencm3/lib/lib.a', needed by 'awesomesauce.elf'.  Stop.
make: Leaving directory '/tmp/libopencm3-template/my-project
```

And there you go. The library wasn't built yet so it doesn't work. It generates this weird "lib.a" name precisely because LIBNAME isn't set when genlink-config.mk tries to do this:

    LIBDEPS += $(OPENCM3_DIR)/lib/lib$(LIBNAME).a

If we go and build the library first it does work ofc. but the whole point of this PR is to get rid of this first step! I don't want users to have to build the library first and/or keep in mind to rebuild the library by hand when I update the submodule. That's just a recipe for problems.

aklomp:
@karlp I have ran into this same issue in my [rplatform](https://github.com/aklomp/rplatform) project. I tried hard to leverage Libopencm3's GCC extensions (in `mk/`) to keep my own `Makefile` as minimal as possible, but I ran into similar issues.

The `Makefile` is [here](https://github.com/aklomp/rplatform/blob/master/Makefile). The project should compile cleanly from the root with a `make`. I think it's a nice real-world example of this issue, and of where the `mk/` stuff is lacking currently.

First of all, @DanielG is right, my Makefile also needs to [define `LIBNAME`](https://github.com/aklomp/rplatform/blob/master/Makefile#L10) explicitly in its setup section, or I get this error message:

```
lib/libopencm3/mk/genlink-config.mk:63: lib/libopencm3/lib/libopencm3_stm32f1.a library variant for the selected device does not exist.
```
The build ultimately predictably fails with:
```
/usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/bin/ld: cannot find -l-static                              
```
What I would expect is that `mk/genlink-config.mk` sets `LIBNAME` from the information it knows: the device family and subfamily. Instead what does is check for the existence of `$(OPENCM3_DIR)/lib/libopencm3_$(genlink_family).a` and use *that* to set the variable. Or not set it, as is the case in my `Makefile`, because at the time of evaluation, at the very top of the definition section of the `Makefile` the library isn't built yet!

So to resolve this circular dependency, I have to manually set `LIBNAME` to `opencm3_stm32f1` to make things work. That's the chicken/egg problem that's also alluded to in #1053.

The second issue is that `genlink-config.mk` sets `LIBDEPS`, which is great, but neither `genlink-rules.mk` or `gcc-rules.mk` actually define a `make` recipe for `$(LIBDEPS)`. So I need to do that manually [here](https://github.com/aklomp/rplatform/blob/master/Makefile#L10). I think it would be fair if `gcc-rules.mk` would include a recipe to build the library, but maybe it would be too restrictive?

This is made worse by the fact that the anemic [`mk/README`](https://github.com/libopencm3/libopencm3/blob/master/mk/README) does not mention needing to set up a custom recipe for `$(LIBDEPS)`. In fact, copypasting the example from that file will not result in a working setup; I could only figure the whole thing out by looking at the sources directly.

The third issue, which is small, is that it's not documented anywhere how to tie in to `libopencm3`'s verbosity flags. I reverse engineered this from the code and [added it](https://github.com/aklomp/rplatform/blob/master/Makefile#L15) myself to my own `Makefile`. I think it would be nice if this was also handled by some optional `*-rules.mk` file.

Issue #1302, opened 2021-02-01
No clock config for WeAct BlackPill 2.0 (stm32f411xe) 100Mhz


sker65:
For this board https://docs.zephyrproject.org/latest/boards/arm/blackpill_f411ce/doc/index.html
there is no clock config for rcc_clock_setup_pll(). The board uses a stm32f411xe and can run up to 100Mhz.

According to CubeMx it should be: 
  RCC_OscInitStruct.PLL.PLLM = 12;
  RCC_OscInitStruct.PLL.PLLN = 96;
  RCC_OscInitStruct.PLL.PLLP = RCC_PLLP_DIV2;
  RCC_OscInitStruct.PLL.PLLQ = 4;

and
  RCC_ClkInitStruct.AHBCLKDivider = RCC_SYSCLK_DIV1;
  RCC_ClkInitStruct.APB1CLKDivider = RCC_HCLK_DIV2;
  RCC_ClkInitStruct.APB2CLKDivider = RCC_HCLK_DIV1;

with flash latency 3.
Can you please provide an additional config. Thx

ChuckM:
It is possible to build the clock_scale structure in your code you know this right?

sker65:
Yes, of course. Just want to mention that this on eis missing as the board is quite popular these days and not everybody can do the magic clock math P, Q, M, N ....

ChuckM:
What I'm saying is that since you're providing documentation, you could also provide example code that sets the clock. The libopencm3 project has gone back and forth at least a couple of times of providing a "higher layer" clock frequency setting API that is more generic (and so doesn't need a custom structure) but the argument against that has been "it has to make too many assumptions and when it doesn't work they will blame the project." 

It is a pretty compelling argument, clocks are one of the more complicated things here. 

That said, you could always submit a pr.

sker65:
> That said, you could always submit a pr.
Will do.

PR #1303, opened 2021-02-01
Add 100MHz clock config

sker65:
Added config, see issue https://github.com/libopencm3/libopencm3/issues/1302

### DIFF
--- a/include/libopencm3/stm32/f4/rcc.h
+++ b/include/libopencm3/stm32/f4/rcc.h
@@ -796,6 +796,7 @@ extern uint32_t rcc_apb2_frequency;
 
 enum rcc_clock_3v3 {
 	RCC_CLOCK_3V3_84MHZ,
+	RCC_CLOCK_3V3_100MHZ,
 	RCC_CLOCK_3V3_168MHZ,
 	RCC_CLOCK_3V3_180MHZ,
 	RCC_CLOCK_3V3_END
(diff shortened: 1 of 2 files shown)


karlp:
Commit message needs to have all the information in it, why is this config relevant, for instance.  it can't just be "add 100meg" like it is now.  given that the f411 familiy does introduce a new top speed, I'm ok with adding it, but do it right :)

sker65:
Did not find a CONTRIBUTE.md or something that describe that type of requirements on commit comments. What now? Could you not just add an additional comment line when merging it?

karlp:
commit comments should _always_ say why, not what.  the what is in the code commit, the comment is to explain the why.  It also must stand alone, not rely on an external source like "fixes github 123" being available.

no, I will not do work for you.  My time is limited, and I absolutely do not scale to "fixing" commits for everyone who wants to contribute.

sker65:
Oh man, my time is limited as well. The why is obvious: there are a couple of boards out their, that support the new timing configuration. I've created the structure and PR . Take it or leave it.

ALTracer:
Please never ever consider merging this.
I'll add my 2 cents since I bought a couple of these boards.

1. It's impossible to clock both SYSCLK/HCLK at 100MHz and USB at 48MHz simultaneously, the Main PLL dividers pllp & pllq are not flexible enough. Best you can do is 96MHz, which is 4% off (remember the 24MHz Flash in these parts) and already supported.

2. This commit adds the struct into `rcc_hse_8mhz_3v3` array, but the WeAct MiniF4 boards have a 25MHz HSE. Also the dividers are calculated for 25MHz HSE (25/12*96/2=100). And it breaks indexing for all the other `rcc_hs*` arrays.

PR #1306, opened 2021-02-04
Added STM32F1 PDF documentation

aht2000:


karlp:
No, we're not committing generated doxygen outputs.  You can generate pdfs locally if you like, and daily html builds are available at http://libopencm3.org/docs/latest/html/

PR #1307, opened 2021-02-04
stm32h7: document optional parameters

karlp:
Provide linking and explanations of optional values
@vielster  is this the correct way of doing the optionals?


vielster:
LGTM

Issue #1299, opened 2021-01-28
STM32H7 SPI implementation is not as per the reference manual


donotnoot:
Recently, I got my hands on a STM32H735G-DK board. Obviously there's no way I'm using the abomination that CubeMX is so I'm using libopencm3 as much as I can, writing directly to registers and such when things are not covered by this fantastic lib.

I was trying to get SPI up and running since it seemed like libopencm3 supports it, but the implementation seems to be using the "classic" SPI peripheral that other chips are using, and it is not compatible with the fancier SPI peripheral ST put on this line.

I checked the reference manuals for both the h7[23] and h7[45] lines and they both have this new peripheral:

Page 2261 and 2236 respectively.
https://www.st.com/resource/en/reference_manual/dm00603761-stm32h723733-stm32h725735-and-stm32h730-value-line-advanced-armbased-32bit-mcus-stmicroelectronics.pdf
https://www.st.com/resource/en/reference_manual/dm00314099-stm32h742-stm32h743-753-and-stm32h750-value-line-advanced-arm-based-32-bit-mcus-stmicroelectronics.pdf

As you can see, there are quite a few new registers and the way the peripheral works is considerably different than the rest of the lines. I've not combed through them both to make sure they're 100% the same but it looks like they are.

I'd like to know whether I'm missing something and whether a contribution for this new SPI peripheral would be useful. I'm porting some software that runs on F4 to H7 and need the SPI working, so at some point I might stop writing to the registers raw and move it to the opencm3, but only if nobody is already working on that already.

There's another thing that I noticed too, and it is that the current RCC implementation doesn't allow to use the SMPS because initialising the LDO is hardcoded into the RCC PLL config function. Should I open a separate issue for that? It's fairly simple to fix, one just needs to separate the PWR stuff out of the RCC function and set up the SMPS before setting up the PLL.

donotnoot:
As I said I'm kinda more or less porting it already. As for compatibility, it's very different. The CR1/2 registers have the same names but very different configs and there are two entirely new registers to configure it further. Also there's no DR, there are separate registers for TX/RX. Anyway, if/when I do port it properly I'll try to make the API similar, but the hardware is rather different so it might not be totally the same.

Fair enough on closing it. I'll keep playing with the dev board that I got and see what the best solution for the SMPS would be in terms of the API, but yeah as you said adding a config to the struct seems like a sensible idea.

karlp:
[...]

PR #1308, opened 2021-02-04
stm32h7: Use enums for the prescalers in the RCC PLL config struct

donotnoot:
This had me confused for a little while until I realised that the fields don't take an integer value representing the clock dividers but actually a value that goes into the relevant registers.

It would have saved me a little bit of time if the documentation was a bit more explicit, so hoping this makes it a bit clearer for the next user :)

### DIFF
--- a/include/libopencm3/stm32/h7/rcc.h
+++ b/include/libopencm3/stm32/h7/rcc.h
@@ -182,29 +182,37 @@ LGPL License Terms @ref lgpl_license
 /** @defgroup rcc_d1cfgr_values RCC_D1CFGR Values
  * @ingroup rcc_registers
  * @{*/
-#define RCC_D1CFGR_D1CPRE_BYP       0x0
-#define RCC_D1CFGR_D1CPRE_DIV2      0x8
-#define RCC_D1CFGR_D1CPRE_DIV4      0x9
-#define RCC_D1CFGR_D1CPRE_DIV8      0xA
-#define RCC_D1CFGR_D1CPRE_DIV16     0xB
-#define RCC_D1CFGR_D1CPRE_DIV64     0xC
-#define RCC_D1CFGR_D1CPRE_DIV128    0xD
-#define RCC_D1CFGR_D1CPRE_DIV256    0xE
-#define RCC_D1CFGR_D1CPRE_DIV512    0xF
-#define RCC_D1CFGR_D1PPRE_BYP       0x0
-#define RCC_D1CFGR_D1PPRE_DIV2      0x4
-#define RCC_D1CFGR_D1PPRE_DIV4      0x5
-#define RCC_D1CFGR_D1PPRE_DIV8      0x6
-#define RCC_D1CFGR_D1PPRE_DIV16     0x7
-#define RCC_D1CFGR_D1HPRE_BYP       0x0
-#define RCC_D1CFGR_D1HPRE_DIV2      0x8
-#define RCC_D1CFGR_D1HPRE_DIV4      0x9
-#define RCC_D1CFGR_D1HPRE_DIV8      0xA
-#define RCC_D1CFGR_D1HPRE_DIV16     0xB
-#define RCC_D1CFGR_D1HPRE_DIV64     0xC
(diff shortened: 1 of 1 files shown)


donotnoot:
Ah, just saw https://github.com/libopencm3/libopencm3/issues/1299#issuecomment-773631832. I can give that a shot too? I do think enums would be even better.

vielster:
Assuming updating to enums doesn't break any backward compatibility (i.e. you don't have to rename anything) then I'm all for it.

donotnoot:
Cool, that's it with enums and backwards compatible (if we ignore the fact that the struct is not binary compatible, which I guess is fine...). I generated the Doxygen docs and they look pretty good too. Everything is dead obvious now.

I'm not too sure about the flash waitstates, we could do the same? But that's a slightly bigger change and honestly not a big deal since a plain integer in that field works.

karlp:
what, no, this looks horrible!  now youv'e got splatter of everything!  I'm not ok with duplicating it all, no thanks.  you can make a case that this is new enough that you can replace the defines with enums, and you can easily add doxygen links to the existing groupings that might be enough without adding enums, but I'm not having _both_ the defines _and_ enums!

donotnoot:
Alright, that's what you meant! Now I know. See the update :+1:

I checked the generated documentation and it looks alright. I'm not familiar enough with Doxygen to know if that's correct/acceptable, please do let me know.

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -182,29 +182,37 @@ LGPL License Terms @ref lgpl_license
...
-#define RCC_D1CFGR_D1HPRE_BYP       0x0
-#define RCC_D1CFGR_D1HPRE_DIV2      0x8
-#define RCC_D1CFGR_D1HPRE_DIV4      0x9
-#define RCC_D1CFGR_D1HPRE_DIV8      0xA
-#define RCC_D1CFGR_D1HPRE_DIV16     0xB
-#define RCC_D1CFGR_D1HPRE_DIV64     0xC
-#define RCC_D1CFGR_D1HPRE_DIV128    0xD
-#define RCC_D1CFGR_D1HPRE_DIV256    0xE
-#define RCC_D1CFGR_D1HPRE_DIV512    0xF
+enum rcc_core_pre {
```
karlp, line 185: I think the enum naming should stick with the existing naming convention, something like rcc_d1cfgr_[your-free-text-here]  things like "rcc_core_" are going to start colliding or being non-visible quite quickly if they're not standardized.

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -455,13 +468,13 @@ struct rcc_pll_config {
...
-  uint8_t ppre3;                    /**< APB3 Peripheral prescaler note: domain 1. */
-  uint8_t ppre4;                    /**< APB4 Peripheral prescaler note: domain 3. */
-  uint8_t flash_waitstates;         /**< Latency Value to set for flahs. */
+  enum rcc_core_pre core_pre;       /**< Core prescaler. */
+  enum rcc_domain1_hpre hpre;       /**< HCLK3 prescaler. */
+  enum rcc_domain2_ppre ppre1;      /**< APB1 Peripheral prescaler. */
+  enum rcc_domain2_ppre ppre2;      /**< APB2 Peripheral prescaler. */
+  enum rcc_domain1_ppre ppre3;      /**< APB3 Peripheral prescaler. */
+  enum rcc_domain3_ppre ppre4;      /**< APB4 Peripheral prescaler. */
+  uint8_t flash_waitstates;         /**< Latency Value to set for flash. FLASH_ACR_LATENCY_xWS value. */
```
karlp, line 477: this is out of scope, and aren't the flash wait states pure numeric anyway?

vielster: These shift value for this macro is zero, so it works either way, but I agree it's sorta outside of the scope of changing the rcc regs to enums

karlp:
still needs the agreed enum name changes.

donotnoot:
Closing due to lack of time/energy.

PR #1315, opened 2021-02-17
stm32g4: Add support for selecting peripheral clock source

ventZl:
Implemented rcc_set_peripheral_clk_sel function for STM32G4. Again.


ventZl:
This one is kind-of functional prerequisity for #1317, as FDCAN has two clocks and FDCAN_clk can only be configured using RCC clock select. Thus FDCAN is defunct unless one configured HSE properly or uses RCC clock select to opt for different FDCAN_clk.

karlp:
I suspect this is incomplete as it ignores CCIPR2?

PR #1316, opened 2021-02-23
Fix ADD1S bit definition in the RTC SHIFTR register

iqyx:


karlp:
merged in b4c03a84 thanks

PR #1317, opened 2021-02-23
STM32G4: Add support for FDCAN

ventZl:
Add stm32g4 support for FDCAN peripheral. Normal / FDCAN operation
supported, bitrate switching and filtering supported via API. Reception and
transmission possible both in interrupt driven and polling mode.
Timestamping and transmit event buffer support in API are TBD.


include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,855 @@
...
+
+/* NTSEG1[7:0]: Nominal timing segment before sample point length */
+#define FDCAN_NBTP_NTSEG1_SHIFT			8
+#define FDCAN_NBTP_NTSEG1_MASK			(__BITS(8) << FDCAN_NBTP_NTSEG1_SHIFT)
+
+/** Obtain value for use as TSEG1 for normal speed.
+ * @param ts amount of tq
+ * @returns value suitable for use with fdcan_init
+ */
+#define FDCAN_NBTP_NTSEG1_TQ(tq)		((tq) << FDCAN_NBTP_NTSEG1_SHIFT)
```
karlp, line 206: why would you not just let the user pass "tq" to fdcan_init directly and do the shift internally?

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,855 @@
...
+#define FDCAN_TXBCIE(can_base)			MMIO32(can_base + 0x00E0)
+#define FDCAN_TXEFS(can_base)			MMIO32(can_base + 0x00E4)
+#define FDCAN_TXEFA(can_base)			MMIO32(can_base + 0x00E8)
+#define FDCAN_CKDIV(can_base)			MMIO32(can_base + 0x0100)
+
+/** @brief Helper macro for defining bitmasks
+ * Will return number having n LSB set
+ * @param n how many bits mask occupies
+ */
+#define __BITS(n)						((1 << n) - 1)
```
karlp, line 103: please no, follow the existing style.

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,855 @@
...
+
+/* SFEC = 0x7 is unused */
+
+#define FDCAN_SFID1_SHIFT				16
+#define FDCAN_SFID1_MASK				(__BITS(11) << FDCAN_SFID1_SHIFT)
+
+#define FDCAN_SFID2_SHIFT				0
+#define FDCAN_SFID2_MASK				(__BITS(11) << FDCAN_SFID2_SHIFT)
+
+#define __CHANGE_BITS(tgt, value, mask) (tgt) = (tgt & ~(mask)) | (value & mask)
```
karlp, line 596: none of this is emant to be used by users right?  put it in the c file.

if you were planning on having this be user API, please reconsider, we've never mixed macros and functions before, and it's not going to start now :)

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,855 @@
...
+	 */
+	struct FDCAN_TxBufferElement TxBuffer[3];
+};
+
+
+/* --- FD-CAN error returns ------------------------------------------------- */
+
+/** @defgroup fdcan_errors FDCAN error return values
+ * @{
+ */
```
karlp, line 792: why not just make this an enum?

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
+/** @defgroup fdcan_file FDCAN peripheral API
+ * @ingroup peripheral_apis
+ * 
+ * @brief <b>libopencm3 STM32G4xx CAN</b>
```
karlp, line 4: fix this "G4" bit, it's not.  Also, this sort of common file normally goes into the common directory, not the root.  (the bxcan file is iirc, the _only_ file that isn't in that style, and not a good example to follow :)

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+ */
+#define FDCAN_CCCR_INIT_TIMEOUT		0x0000FFFF
+
+static int fdcan_cccr_init_cfg(uint32_t canport, bool set);
+static int fdcan_get_fill_rxfifo(uint32_t canport, uint8_t fifo_id, 
+		unsigned * get_index, unsigned * pending_frames);
+static struct FDCAN_MessageRAM * fdcan_get_msgram_addr(uint32_t canport);
+static int fdcan_get_free_txbuf(uint32_t canport);
+static uint32_t fdcan_length_to_dlc(uint8_t length);
+static uint8_t fdcan_dlc_to_length(uint32_t dlc);
```
karlp, line 57: just put the static impls up the top.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+
+/** @brief FDCAN Reset
+ *
+ * This will reset the FDCAN peripheral. All registers will be placed in
+ * the reset state. Note that it is not possible to reset FDCAN per-CAN
+ */
+void fdcan_reset(void)
+{
+	rcc_periph_reset_pulse(RST_FDCAN);
+}
```
karlp, line 71: I would suggest that this type of api duplication not be encouraged, just drop the function entirely. peripheral reset is already handled (twice) in the rcc apis.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+ * Initialize the selected CAN peripheral block. This function will switch CAN block
+ * into initialization mode and set crucial configuration flags and timing registers.
+ * CAN block is then left in initialization mode in order to 
+ * 
+ * @param[in] canport Unsigend int32. CAN register base address @ref can_reg_base.
+ * @param[in] brs bool. Enable bitrate switching during Tx (only avaiable if FDCAN operation is also enabled)
+ * @param[in] fdoe Enable FDCAN operation
+ * @param[in] nart Disable automatic retransmission
+ * @param[in] rflm Turn on receive FIFO locked mode, otherwize FIFO is in overwrite mode
+ * @param[in] txqm Switch Tx into queue mode, if false then Tx works in FIFO mode
```
karlp, line 84: you should be getting doxygen warnings here on none of these parameters matching :)

include/libopencm3/stm32/fdcan.h
ventZl: I wanted to make API semantically as close to can.c as possible. can.c defines a lot of NTSEG_xTQ defines. But FDCAN is much more flexible with some prescalers as wide as 7 bits and having 256 defines would be weird. If API similarity is not desired, I may change it to accept bare numbers. Both here and on other places around the code, which behave similarly.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+ * @param[in] sjw Resynchronization time quanta jump width.
+ * @param[in] ts1 Time segment 1 time quanta width.
+ * @param[in] ts2 Time segment 2 time quanta width.
+ * @param[in] brp Baud rate prescaler.
+ * @param[in] loopback Turn on loopback mode
+ * @param[in] silent Turn on silent (monitor) mode
+ * @returns int FDCAN_E_OK on success, FDCAN_E_TIMEOUT on timeout waiting for INIT bit to flip.
+ */
+int fdcan_init(uint32_t canport, bool brs_enable, bool fd_op_enable, 
+		 bool auto_retr_disable, bool rx_fifo_locked, bool tx_queue_mode,
```
karlp, line 94: cmon, just add the y in retry?  that doesn't worth trying to save a character over.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+ * @param[in] nart Disable automatic retransmission
+ * @param[in] rflm Turn on receive FIFO locked mode, otherwize FIFO is in overwrite mode
+ * @param[in] txqm Switch Tx into queue mode, if false then Tx works in FIFO mode 
+ * @param[in] sjw Resynchronization time quanta jump width.
+ * @param[in] ts1 Time segment 1 time quanta width.
+ * @param[in] ts2 Time segment 2 time quanta width.
+ * @param[in] brp Baud rate prescaler.
+ * @param[in] loopback Turn on loopback mode
+ * @param[in] silent Turn on silent (monitor) mode
+ * @returns int FDCAN_E_OK on success, FDCAN_E_TIMEOUT on timeout waiting for INIT bit to flip.
```
karlp, line 91: this is exactly why you want to use an enum here, not your defines.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+
+
+/* Timeout for FDCAN_CCCR register INIT bit to accept set value.
+ * This timeout is required because FDCAN uses two different clocks
+ * feeding two different portions of block. There can be slight delay
+ * based on how clocks are set up. While amount of FDCAN_clk /
+ * FDCAN_pclk combinations is high and clock speeds may vary a lot,
+ * following value has been choosen as sane default.
+ */
+#define FDCAN_CCCR_INIT_TIMEOUT		0x0000FFFF
```
karlp, line 49: people are almost definitely going to want to tune that themselves, if you're introducing a world where functions have a timeout.  Also, please definitely consider adding api methods that allow the functions to be used in non-blocking fashion.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+	 * FDCAN blocks, message RAM for n-th FDCAN starts at address 
+	 * end of (n-1)-th block + 4 (explicitly, offset 0x354).
+	 *
+	 * It turns out, that this statement is also false! In fact FDCAN
+	 * message RAMs are packed tightly and n-th block starts immediately
+	 * after (n-1)-th block ends. Thus offset is going to be computed
+	 * using formula:
+	 *
+	 * FDCAN1_RAM_BASE + (block_id * sizeof(struct FDCAN_MessageRAM))
+	 */
```
karlp, line 822: why is this not just 3 constants?

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+int fdcan_filter_init(uint32_t canport, uint8_t std_filt, uint8_t ext_filt)
+{
+	struct FDCAN_MessageRAM * ram = fdcan_get_msgram_addr(canport);
+
+	if ((FDCAN_CCCR(canport) & FDCAN_CCCR_INIT) == 0)
+	{
+		return FDCAN_E_BUSY;
+	}
+
+	if (!ram)
```
karlp, line 235: aiui, this can never happen?  we don't need to account for people passing made up random can ports (and no other api tries to protect from that case.  YOu're really breaking new ground introducing return codes, and that's _bad_ but it's a very quick path to code bloat with error checking.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+	}
+
+	/* Only perform initialization of message RAM if there are
+	 * any filters required
+	 */
+	if (std_filt > 0)
+	{
+		if (std_filt > FDCAN_SFT_MAX_NR)
+		{
+			return FDCAN_E_OUTOFRANGE;
```
karlp, line 247: same here, at a certain point, you should trust the users instead of trying to validate all inputs.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+			| (((std_filt + 1) << FDCAN_RXGFC_LSS_SHIFT) & FDCAN_RXGFC_LSS_MASK);
+
+
+		for (int q = 0; q < FDCAN_SFT_MAX_NR; ++q)
+		{
+			ram->LFSSA[q].type_id1_conf_id2 = 0;
+		}
+	}
+	else
+	{
```
karlp, line 260: I mentioned that  it's if (blah) {  so just for compelteness, it's also ``` } else { ``` style wise.  there's a "make stylecheck" target you can use.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+ * depending on matching mode selected
+ * @param action Action performed if filtering rule matches frame ID, see 
+ * FDCAN_SFEC_<action> defines.
+ * @returns FDCAN_E_OK if filter was set up properly, 
+ * FDCAN_E_INVALID if FDCAN block base address is unknown and 
+ * FDCAN_E_OUTOFRANGE is filter number is larger than previosly configured.
+ */
+int fdcan_std_filter_set(uint32_t canport, uint32_t nr, 
+             uint8_t id_list_mode, uint32_t id1, uint32_t id2,
+		     uint8_t action)
```
karlp, line 310: whitespace looks odd here?

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+{
+	struct FDCAN_MessageRAM * ram = fdcan_get_msgram_addr(canport);
+
+	if (!ram)
+	{
+		return FDCAN_E_INVALID;
+	}
+
+	uint32_t rxgfc_lse_current = ((FDCAN_RXGFC(canport) & FDCAN_RXGFC_LSE_MASK) 
+			>> FDCAN_RXGFC_LSE_SHIFT);
```
karlp, line 371: this is just reading back what the user set themselves earlier right?  this is nice big secure safe defensive never trust anyone code, but it's very much not what we've done so far with "do what the users asked to do, with a neat api and as little fuss as possible"  I would consider whether this is really necessary or beneficial.

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+ * @param data Unsigned int8[]. Message payload data.
+ * @returns int 0, 1 or 2 on success and depending on which outgoing mailbox got
+ * selected. On error returns FDCAN_E_BUSY if no mailbox was available,
+ * FDCAN_E_INVALID if CAN block base address and/or message length are invalid.
+ */
+int fdcan_transmit(uint32_t canport, uint32_t id, bool ext, bool rtr,
+        bool fdcan_fmt, bool btr_switch, uint8_t length, const uint8_t *data)
+{
+	int mailbox;
+	uint32_t dlc, flags = 0;
```
karlp, line 408: personally, I don't see any benefit to this sort of misra style, but I'm not going to complain about it.   (... much ;)

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+ * @param fifo FIFO id.
+ * @param release Release the FIFO automatically after coping data out.
+ * @param id Message ID.
+ * @param ext The message ID is extended?
+ * @param rtr Request of transmission?
+ * @param fmi ID of the matched filter.
+ * @param length Length of message payload.
+ * @param data Message payload data.
+ * @param timestamp Pointer to store the message timestamp.
+ * 			Only valid on time triggered CAN. Use NULL to ignore.
```
karlp, line 485: no return val doc?

lib/stm32/fdcan.c
```diff
@@ -0,0 +1,891 @@
...
+		{
+			/* Already there, sir */
+			return FDCAN_E_OK;
+		}
+
+		FDCAN_CCCR(canport) &= ~FDCAN_CCCR_INIT;
+		expected = 0;
+	}
+
+	/* Wait, until INIT bit is acknowledged */
```
karlp, line 723: if you split this check into an _is_ok, or _is_init_complete or something, then you can have a function that sets and returns, and the users then polls _is_ok() themselves, and also an _init_blocking() that does the default

lib/stm32/g4/Makefile
```diff
@@ -56,6 +56,7 @@ OBJS += usb_hid.o
 OBJS += usb_midi.o
 OBJS += usb_msc.o
 OBJS += st_usbfs_core.o st_usbfs_v2.o
+OBJS += fdcan.o
```
karlp, line 59: this list is kept sorted, with the USB grouped separately.  This shouldgo up between dmamux.o and flash.o...

karlp:
I generally like the look of this. it looks well thought out and generally sane.  I certainly, at least superficially, prefer this implementation to the implementaiton in https://github.com/libopencm3/libopencm3/pull/1166  That said, I don't use any can peripherals, bx or fd, and I don't have any equipment.  Could you guys talk together and see if there's strengths / weaknesses  or anything else we could use to get the whole thing moving forward? talk to @mattofak about it maybe?  It certainly avoids these extra "libraries" that seemed ... dubious.

include/libopencm3/stm32/fdcan.h
karlp: I'm not entirely sure that following the existing can.c api is the best idea.  it's an extremely old piece of code, and very out of style with almost everything else in the library.  we receive almost no feedback on it, and it receies almost no changes.  Might be perfect, might be unusable garbage :)

lib/stm32/fdcan.c
ventZl: In certain controllers, such as H7, size of Message RAM per CPU block is configurable. Constants probably wouldn't work well here. I originally expected that different MCU families will have different fixed sizes, but user-configurable size calls for completely different approach.

karlp: ok, but this function is constant, there's no support for configurable message ram here?

ventZl: No, as this is written and tested, it only supports constant mapping CANFDx_BASE => CANFDx_MSGRAM_BASE. Code below should end up as returning constants as all value are compile-time deducible.

include/libopencm3/stm32/fdcan.h
ventZl: I think it is actually not even used as macros below are really a performance penalty if being used. And nobody sane enough would probably want to deal with FDCAN msgram on his own anyway. I will remove this & friends.

lib/stm32/fdcan.c
ventZl: Will do

include/libopencm3/stm32/fdcan.h
ventZl: Removed, all uses converted to plain hexa values.

include/libopencm3/stm32/fdcan.h
ventZl: Removed

ventZl:
I have reworked how initialization works. fdcan_init() and fdcan_start() now both accept user timeout value and are non-blocking if 0 is passed. FDCAN initialization itself is broken into multiple functions, because I don't like that bunch of arguments for old fdcan_init(). It also simplifies addition of support for new features. Function naming is changed to be fdcan_<action>_<subject>(). No more Yoda. User errors are not sanitized / checked except of CAN IDs, which could spill into flags and this kind of error could be tricky to debug. Many functions now return nothing.

For now I have decided not to touch how message RAM is implemented. Devices such as F7/H7 have much more complicated message RAM organization and contain additional configuration register. I don't possess any such device to test proper functionality.

Waiting to hear from @mattofak

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,793 @@
...
+/** @{ */
+
+/* FDCAN block base addresses. Used in functions to identify FDCAN block being manipulated. */
+
+/** @defgroup fdcan_block FDCAN block base addresses
+ * @{
+ */
+#define CAN1							FDCAN1_BASE
+#define CAN2							FDCAN2_BASE
+#define CAN3							FDCAN3_BASE
```
karlp, line 47: do any parts have both bxcan and fdcan? is this a safe idea to just have "can" be the name? (not an expert in this field, just curious on this one, as we're using fdcan for everything else in this file)

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,793 @@
...
+#define FDCAN_TXEFA(can_base)			MMIO32(can_base + 0x00E8)
+#define FDCAN_CKDIV(can_base)			MMIO32(can_base + 0x0100)
+
+/* DAY[7:0]: FDCAN core revision date */
+#define FDCAN_CREL_DAY_SHIFT			0
+#define FDCAN_CREL_DAY_MASK				(0xFF << FDCAN_CREL_DAY_SHIFT)
+
+/* MON[7:0]: FDCAN core revision month */
+#define FDCAN_CREL_MON_SHIFT			8
+#define FDCAN_CREL_MON_MASK				(0xFF << FDCAN_CREL_MON_SHIFT)
```
karlp, line 105: has this changed frrom last rev?  masks are normally specified unshifted, so they can be used by users to check their own values first if desired.  there are apis using pre-shifted, but all new stuff is unshifted, and other stuff slowly moves.  I'm sorry if I mised this the first time?

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,793 @@
...
+ * and 16 32bit values for message payload.
+ * This area of memory can only be accessed in 32bit quantities
+ */
+struct fdcan_rx_fifo_element {
+	/** Aggregate of message identifier and flags. */
+	uint32_t identifier_flags;
+	/**< Aggregate of filter match ID, transfer format, DLC and timestamp */
+	uint32_t filt_fmt_dlc_ts;
+	/**< Message payload data */
+	uint32_t data[64 / sizeof(uint32_t)];
```
karlp, line 638: why.... not just put 16 here?  like yu say, it has space for 16 32bit values, and it's defined as a uint32_t array.... so? just make it clear and simple?

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,793 @@
...
+ * This area of memory can only be accessed in 32bit quantities
+ */
+struct fdcan_tx_buffer_element {
+	/**< Aggregate of message identifier and flags. */
+	uint32_t identifier_flags;
+
+	/**< Aggregate of event ID, transfer format and DLC */
+	uint32_t evt_fmt_dlc_res;
+	/**< Message payload data */
+	uint32_t data[64 / sizeof(uint32_t)];
```
karlp, line 665: same

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,793 @@
...
+/**@}*/
+
+/* --- FD-CAN functions ----------------------------------------------------- */
+
+BEGIN_DECLS
+
+int fdcan_init(uint32_t canport, uint32_t timeout);
+
+void fdcan_set_can(uint32_t canport, bool auto_retry_disable, bool rx_fifo_locked,
+		bool tx_queue_mode,	bool silent, uint32_t n_sjw, uint32_t n_ts1, uint32_t n_ts2,
```
karlp, line 755: you have a stray tab in there I think

karlp: is this a method for setting "basic" can mode? vs fdcan_set_fdcan?  would "fdcan_[set_]mode_can/fdcan_[set_]mode_fdcan" be better or is this clear enough for users as is?

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,793 @@
...
+
+void fdcan_set_test(uint32_t canport, bool testing, bool loopback);
+
+void fdcan_filter_init(uint32_t canport, uint8_t std_filt, uint8_t ext_filt);
+
+int fdcan_start(uint32_t canport, uint32_t timeout);
+
+void fdcan_set_std_filter(uint32_t canport, uint32_t nr,
+		uint8_t id_list_mode, uint32_t id1, uint32_t id2,
+		uint8_t action);
```
karlp, line 769: unless you _need_ this to be 8 bit, just use "int" here, so the compiler isn't being forced to mask things down to 8 bits for no reason.  (in general, this applies to all api places.  just "int" and "unsigned" will do the best thing for you normally, though you can go crazy with uint_fast8_t if you want to go nuts.... (I'd not really recommend using the "special" types in public apis though honestly.)

karlp: (you know what, don't worry about this, I just went and played on godbolt for a while, and this advice is outdated these days, just leave this as is)

lib/stm32/common/fdcan.c
```diff
@@ -0,0 +1,763 @@
...
+/* --- FD-CAN internal functions -------------------------------------------- */
+
+/** Routine implementing FDCAN_CCCR's INIT bit manipulation.
+ * This routine will change INIT bit and wait for it to actually
+ * change its value. If change won't happen before timeout,
+ * error is signalized. If INIT bit already has value which
+ * should be set, this function will return immediately.
+ *
+ * @param [in] canport FDCAN block base address
+ * @param [in] set new value of INIT, true means set
```
karlp, line 51: it kinda feels like this should be fdcan_ccr_init and fdcan_cccr_deinit?

lib/stm32/common/fdcan.c
```diff
@@ -0,0 +1,763 @@
...
+ * @param [out] get_index Address of buffer where next get index will be stored
+ * @param [out] pending_frames Address of buffer where amount of pending frame will be stored.
+ * @returns Operation error status.
+ */
+static int fdcan_get_fill_rxfifo(uint32_t canport, uint8_t fifo_id, unsigned *get_index,
+		unsigned *pending_frames)
+{
+	if (fifo_id == 0) {
+		*get_index = (FDCAN_RXF0S(canport) & FDCAN_RXF0S_F0GI_MASK)
+			>> FDCAN_RXF0S_F0GI_SHIFT;
```
karlp, line 127: eg, using unshifted masks, all that changes is that you shift before hand.
(leaving review here, have to siwtch tasks)

lib/stm32/common/fdcan.c
```diff
@@ -0,0 +1,763 @@
...
+		return (struct fdcan_message_ram *) (FDCAN1_RAM_BASE + 0);
+	} else if (canport == CAN2) {
+		return (struct fdcan_message_ram *)
+			(FDCAN1_RAM_BASE + sizeof(struct fdcan_message_ram));
+	} else if (canport == CAN3) {
+		return (struct fdcan_message_ram *)
+			(FDCAN1_RAM_BASE + (2 * sizeof(struct fdcan_message_ram)));
+	}
+
+	return (void *) 0;
```
karlp, line 178: could this just be "return NULL" ?

lib/stm32/common/fdcan.c
karlp: nvm, I see now later that this is an internal method :)

lib/stm32/common/fdcan.c
```diff
@@ -0,0 +1,763 @@
...
+
+/** Put FDCAN block into INIT mode for setup
+ * Initialize the selected CAN peripheral block. This function will switch CAN block
+ * into initialization mode and set crucial configuration flags and timing registers.
+ * CAN block is then left in initialization mode in order to
+ *
+ * @param[in] canport CAN register base address @ref can_reg_base.
+ * @param [in] timeout Amount of empty busy loops, which routine should wait for FDCAN
+ *				confirming that it entered INIT mode. If set to 0, function will return
+ *				immediately.
```
karlp, line 237: what's the function that can be used to then check for it being ready? that should be linked here too.

include/libopencm3/stm32/fdcan.h
karlp: don't worry about this, I think it's fine as is...

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,793 @@
...
+#define FDCAN_RXF1S_F1FL_SHIFT			0
+#define FDCAN_RXF1S_F1FL_MASK			(0xF << FDCAN_RXF1S_F1FL_SHIFT)
+
+/* F1GI[1:0]: Get index of Rx FIFO 1 */
+#define FDCAN_RXF1S_F1GI_SHIFT			8
+#define FDCAN_RXF1S_F1GI_MASK			(0x3 << FDCAN_RXF1S_F1GI_SHIFT)
+
+/* F1PI[1:0]: Put index of Rx FIFO 1 */
+#define FDCAN_RXF1S_F1PI_SHIFT			16
+#define FDCAN_RXF1S_F1PI_MASK			(0x3 << FDCAN_RXF1S_F1PI_SHIFT)
```
karlp, line 419: it might be possible tosimpliy this series, as it's all, 0xf, 0x3@8, 0x3@16, for both fifos? probably no gain rightnow, but might make some stuff easier.

include/libopencm3/stm32/fdcan.h
```diff
@@ -0,0 +1,793 @@
...
+#define FDCAN_IE(can_base)				MMIO32(can_base + 0x0054)
+#define FDCAN_ILS(can_base)				MMIO32(can_base + 0x0058)
+#define FDCAN_ILE(can_base)				MMIO32(can_base + 0x005C)
+#define FDCAN_RXGFC(can_base)			MMIO32(can_base + 0x0080)
+#define FDCAN_XIDAM(can_base)			MMIO32(can_base + 0x0084)
+#define FDCAN_HPMS(can_base)			MMIO32(can_base + 0x0088)
+#define FDCAN_RXF0S(can_base)			MMIO32(can_base + 0x0090)
+#define FDCAN_RXF0A(can_base)			MMIO32(can_base + 0x0094)
+#define FDCAN_RXF1S(can_base)			MMIO32(can_base + 0x0098)
+#define FDCAN_RXF1A(can_base)			MMIO32(can_base + 0x009C)
```
karlp, line 85: this for instance could potentially beocme 
```
FDCAN_RXFIFOS(can, fifo)  MMIO(can_base + 0x0090 + (8*fifo))
FDCAN_RXFIFOA(can, fifo)  MMIO(can_base + 0x0094 + (8*fifo))
```
then in a few places later, where you have two if cases with ~identical code, you could simplify it a bit.  Just a thought though

lib/stm32/common/fdcan.c
```diff
@@ -0,0 +1,763 @@
...
+	} else if (fifo_id == 1) {
+		get_index = (FDCAN_RXF1S(canport) & FDCAN_RXF1S_F1GI_MASK)
+					>> FDCAN_RXF1S_F1GI_SHIFT;
+		pending_frames = (FDCAN_RXF1S(canport) & FDCAN_RXF1S_F1FL_MASK)
+			>> FDCAN_RXF1S_F1FL_SHIFT;
+
+		if (pending_frames) {
+			FDCAN_RXF1A(canport) |= (get_index & FDCAN_RXF1A_R1AI_MASK)
+				<< FDCAN_RXF1A_R1AI_SHIFT;
+		}
```
karlp, line 701: like this section could just become 
```
get_index = FDCAN_RXFIFOS(canport, fifoid) & ___fifo_GI_mask >> __fifo_GI_SHIFT
pending = same...
FDCAN_FIFOA(can, fifoid) |= get_index && mask << shift....
```
and then just return void, and drop teh fifo check?

lib/stm32/common/fdcan.c
```diff
@@ -0,0 +1,763 @@
...
+	}
+
+	if (release) {
+		if (fifo_id == 0) {
+			FDCAN_RXF0A(canport) |= (get_index & FDCAN_RXF0A_R0AI_MASK)
+				<< FDCAN_RXF0A_R0AI_SHIFT;
+		} else {
+			FDCAN_RXF1A(canport) |= (get_index & FDCAN_RXF1A_R1AI_MASK)
+				<< FDCAN_RXF1A_R1AI_SHIFT;
+		}
```
karlp, line 662: same here

include/libopencm3/stm32/fdcan.h
ventZl: I am not aware of any parts which would have both BXCAN and FDCAN. It probably doesn't make any sense as FDCAN is generally backwards compatible with CAN 2.0 and is more capable, than BXCAN.

include/libopencm3/stm32/fdcan.h
ventZl: I have removed use of __BITS() macro. I have probably took some older code with shifted masks as an example. Will adjust it as it definitely makes more sense.

include/libopencm3/stm32/fdcan.h
ventZl: I chose this deliberately complicated way of defining the size, because value of 16 is actually calculated. Spec just says, that there is space for 64 data bytes allocated. It doesn't explicitly state the size of FIFO / buffer element anywhere. IMO it provides more clue to put numbers which exist in specs in, than calculate arbitrary numbers which you have to pair on your own. Especially given the fact that 64-byte long buffer is weirdly defined as an array of 32-bit ints, because byte access doesn't work with Message RAM at all.

include/libopencm3/stm32/fdcan.h
ventZl: Those two modes can be combined. Actually, fdcan_set_can() is mandatory, because timings set here are used in arbitration phase even for FDCAN fast mode messages. Optional fdcan_set_fdcan() only configures FDCAN specific stuff which you later can, but don't have to use (it can be enabled on per-message basis). And yes, there's one stray tab.

include/libopencm3/stm32/fdcan.h
ventZl: That's actually a good idea! I felt, that the code below, which gets affected by such change was ugly and duplicated but couldn't come up with elegant way to deduplicate it.

include/libopencm3/stm32/fdcan.h
ventZl: I'd like to keep naming consistent, so if FDCAN_<register>_<parameter> is defined somewhere, it should also be defined for all other valid combinations of register and parameter even if those are duplications. But yes, actual values could be put into "register agnostic" macro and then just referenced here.

ventZl:
*v3* 

Masks are now unshifted, code adjusted accordingly and fixed a bunch of bugs related to shifting and masking. Remaining function names changed to follow fdcan_action_subject strategy. Some code simplified and deduplicated.

include/libopencm3/stm32/fdcan.h
karlp: ok, I'm merging this now, but how abotu I add ""64bytes memory, only word accessible" or something?

karlp:
merged via rebase in 458250dc thanks!

PR #1319, opened 2021-03-08
STM32 RTC: added better RTC periph API functions

bdebyl:
# Summary
- Adds missing functionality to RTC library shared for L1, F0, F2, and F4 STM32s
- Adds two new `enum`s to RTC:
   - `rtc_fmt`
   - `rtc_wdu`
- Revises `RST_BDCR` naming in `rcc.h` for consistency (_no other variables are named by full name like such_)
- Adds private non-user facing function `_rtc_dec_to_bcd(uint8_t dec)` in `rcc_common.c` for ease of conversion of time units

# Notes
- Unsure about doxygen commenting style, feedback appreciated
- Tested on STM32F0 and STM32F4, do not have L0, or F2 to test, but reference manuals check out
- Intentionally avoided adding BCD helpers as this is up to the user.

### DIFF
--- a/include/libopencm3/stm32/common/rtc_common_l1f024.h
+++ b/include/libopencm3/stm32/common/rtc_common_l1f024.h
@@ -426,6 +426,20 @@ specific memorymap.h header before including this header file.*/
 #define RTC_ALRMXSSR_SS_SHIFT     (0)
 #define RTC_ALARXSSR_SS_MASK      (0x7fff)
 
+enum rtc_fmt {
+	RTC_CR_FMT_24H = 0x0,
+	RTC_CR_FMT_12H,
+};
+
+enum rtc_wdu {
+	RTC_DR_WDU_MON = 0x01,
+	RTC_DR_WDU_TUE,
+	RTC_DR_WDU_WED,
+	RTC_DR_WDU_THU,
+	RTC_DR_WDU_FRI,
+	RTC_DR_WDU_SAT,
+	RTC_DR_WDU_SUN,
+};
 
 BEGIN_DECLS
 
@@ -435,6 +449,23 @@ void rtc_lock(void);
 void rtc_unlock(void);
 void rtc_set_wakeup_time(uint16_t wkup_time, uint8_t rtc_cr_wucksel);
 void rtc_clear_wakeup_flag(void);
+void rtc_set_init_flag(void);
+void rtc_clear_init_flag(void);
+bool rtc_init_flag_is_ready(void);
+void rtc_wait_for_init_ready(void);
+void rtc_set_bypass_shadow_register(void);
+void rtc_enable_bypass_shadow_register(void);
+void rtc_disable_bypass_shadow_register(void);
+void rtc_set_hour_format(enum rtc_fmt fmt);
+void rtc_calendar_set_year(uint8_t year_tens, uint8_t year_units);
+void rtc_calendar_set_weekday(enum rtc_wdu weekday);
+void rtc_calendar_set_month(uint8_t month_tens, uint8_t month_units);
+void rtc_calendar_set_day(uint8_t day_tens, uint8_t day_units);
+void rtc_time_set_am_notation(void);
+void rtc_time_set_pm_notation(void);
+void rtc_time_set_hour(uint8_t hour_tens, uint8_t hour_units);
+void rtc_time_set_minute(uint8_t min_tens, uint8_t min_units);
+void rtc_time_set_second(uint8_t sec_tens, uint8_t sec_units);
 
 END_DECLS
 /**@}*/
--- a/include/libopencm3/stm32/common/rtc_common_l1f024.h
+++ b/include/libopencm3/stm32/common/rtc_common_l1f024.h
@@ -426,12 +426,7 @@ specific memorymap.h header before including this header file.*/
 #define RTC_ALRMXSSR_SS_SHIFT     (0)
 #define RTC_ALARXSSR_SS_MASK      (0x7fff)
 
-enum rtc_fmt {
-	RTC_CR_FMT_24H = 0x0,
-	RTC_CR_FMT_12H,
-};
-
-enum rtc_wdu {
+enum rtc_weekday {
 	RTC_DR_WDU_MON = 0x01,
 	RTC_DR_WDU_TUE,
 	RTC_DR_WDU_WED,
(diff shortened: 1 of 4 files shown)


include/libopencm3/stm32/f4/rcc.h
```diff
@@ -1092,6 +1092,9 @@ enum rcc_periph_rst {
 	RST_SAI1RST	= _REG_BIT(0x24, 22),/* F42x, F43x */
 	RST_LTDC	= _REG_BIT(0x24, 26),/* F42x, F43x */
 	RST_DSI		= _REG_BIT(0x24, 27),/* F42x, F43x */
+
+	/* Advanced peripherals */
+	RST_BDCR	= _REG_BIT(0x70, 16),/* BDCR[16] */
```
bdebyl, line 1097: The RCC reset bit for the BDCR was never added here, only in the F0

include/libopencm3/stm32/common/rtc_common_l1f024.h
```diff
@@ -426,6 +426,20 @@ specific memorymap.h header before including this header file.*/
 #define RTC_ALRMXSSR_SS_SHIFT     (0)
 #define RTC_ALARXSSR_SS_MASK      (0x7fff)
 
+enum rtc_fmt {
+	RTC_CR_FMT_24H = 0x0,
+	RTC_CR_FMT_12H,
```
bdebyl, line 431: I'm not entirely sure a struct is necessary here, could change this to the same method as `rtc_time_set_am_notation` and `rtc_time_set_pm_notation` instead of passing an enum to `rtc_set_hour_format`

On a side note, I'm not entirely sure why the `RTC_CR` has a 12/24h format option **and** the `RTC_TR` time register itself has an AM(24h)/PM format option.

include/libopencm3/stm32/common/rtc_common_l1f024.h
```diff
@@ -426,6 +426,20 @@ specific memorymap.h header before including this header file.*/
...
+};
+
+enum rtc_wdu {
```
karlp, line 434: what's wrong with rtc_weekday?

include/libopencm3/stm32/f4/rcc.h
```diff
@@ -1092,6 +1092,9 @@ enum rcc_periph_rst {
 	RST_SAI1RST	= _REG_BIT(0x24, 22),/* F42x, F43x */
 	RST_LTDC	= _REG_BIT(0x24, 26),/* F42x, F43x */
 	RST_DSI		= _REG_BIT(0x24, 27),/* F42x, F43x */
+
+	/* Advanced peripherals */
```
karlp, line 1096: Where does this "advanced peripherals" come from? It's not something that seems to be in rm0090 anywhere?

lib/stm32/common/rtc_common_l1f024.c
```diff
@@ -122,4 +122,189 @@ void rtc_clear_wakeup_flag(void)
...
+/** @brief Sets the RTC BCD calendar year value (tens and ones)
+
+@details Requires unlocking the RTC write-protection (RTC_WPR)
+*/
+void rtc_calendar_set_year(uint8_t year_tens, uint8_t year_units)
+{
+	RTC_DR = ((year_tens & RTC_DR_YT_MASK) << RTC_DR_YT_SHIFT) |
+		(RTC_DR & ~(RTC_DR_YT_MASK << RTC_DR_YT_SHIFT));
+	RTC_DR = ((year_units & RTC_DR_YU_MASK) << RTC_DR_YU_SHIFT) |
+		(RTC_DR & ~(RTC_DR_YU_MASK << RTC_DR_YU_SHIFT));
```
karlp, line 212: can this just be
```
RTC_DR &= ~(YT_MASK << YT_SHIFT | YU_MASK << YU_SHIFT);
RTC_DR |= (yt << YT_SHIFT | yu << YU_SHIFT)
```
I don't really see why we need to modify each section separately?

Further, is this a useful user facing api? Do users _want_ to write their own div/mod/log functions to get the different parts of a year out?

lib/stm32/common/rtc_common_l1f024.c
```diff
@@ -122,4 +122,189 @@ void rtc_clear_wakeup_flag(void)
...
+}
+
+/*---------------------------------------------------------------------------*/
+/** @brief Sets RTC time to use AM or 24-hour format
+
+@details Requires unlocking the RTC write-protection (RTC_WPR)
+*/
+void rtc_time_set_am_notation(void)
+{
+	RTC_TR &= ~RTC_TR_PM;
```
karlp, line 259: I don't believe this does what you think.

As I understand it, the AM/PM bit in the TR register is for the _time_, so, say you want to set the time to 2pm, you initialise by setting hours to 2, and am/pm bit set.

that _loads_ the time, but if you have the CR bit set to 24 hour mode, you would re-read it as hours == 14.

For plain reading, this is just the display bit.  I think it's only writeable here for init time, not a mode change.  I'd probably just make the setting/clearing of this bit an extra boolean parameter on rtc_time_set_hour() ?

include/libopencm3/stm32/common/rtc_common_l1f024.h
karlp: Yeah, I don't think you need an enum for this.  Keep those two functions perhaps, btu make them set the RTC_CR flag appropriately, and move teh TR settings elsewherew as I suggested.

include/libopencm3/stm32/common/rtc_common_l1f024.h
bdebyl: Just naming the enum differently? I'm fine with this :+1:

include/libopencm3/stm32/f4/rcc.h
bdebyl: I saw it in the f0 section of the rcc so I just copy-pasted it

lib/stm32/common/rtc_common_l1f024.c
bdebyl: Ultimately it's difficult to say exactly what users want without taking a poll. I could try to come up with a more effective way of setting the year, and similar functionality.

lib/stm32/common/rtc_common_l1f024.c
bdebyl: Combining it to set-hour is a great idea!

include/libopencm3/stm32/f4/rcc.h
bdebyl: I've changed the name from "Advanced peripherals" to the rm0090s "Backup domain control"

lib/stm32/common/rtc_common_l1f024.c
bdebyl: I've added a `_rtc_dec_to_bcd()` static void function as part of rtc_common.c -- I don't expect this to be a user facing function and did not want to make this a define macro. Let me know if you want this changed up

lib/stm32/common/rtc_common_l1f024.c
```diff
@@ -28,6 +28,11 @@
 
 #include <libopencm3/stm32/rtc.h>
 
+static uint8_t _rtc_dec_to_bcd(uint8_t dec)
```
bdebyl, line 31: I didn't want to make this a `#define` unless this is desirable here. Additionally, this is not intended to be a user-facing function unless people believe users will want to use this.

include/libopencm3/stm32/common/rtc_common_l1f024.h
```diff
@@ -435,6 +444,22 @@ void rtc_lock(void);
...
 void rtc_clear_wakeup_flag(void);
+void rtc_set_init_flag(void);
+void rtc_clear_init_flag(void);
+bool rtc_init_flag_is_ready(void);
+void rtc_wait_for_init_ready(void);
+void rtc_set_bypass_shadow_register(void);
+void rtc_enable_bypass_shadow_register(void);
+void rtc_disable_bypass_shadow_register(void);
+void rtc_set_am_format(void);
+void rtc_set_pm_format(void);
```
bdebyl, line 455: Let me know if you think the naming of this is clear enough.

karlp:
merged via rebase after squishing reviews in 0d72e673

PR #1320, opened 2021-03-16
stm32: add timer_reset function

seymar:
I was missing a `timer_reset` function in the timer peripheral, I2C, SPI, CAN, already have it. May be useful for other peripherals as well.


karlp:
it was actually removed deliberately, and should be removed from the others.  there's no point beyond code bloat for this indirection, just use the rcc_ functions directly. see 034dbf20ff

PR #1325, opened 2021-04-09
STM32H7: Implement basic FDCAN support

ventZl:
Add stm32h7 support for FDCAN peripheral. Source level compatibility is
provided with stm32g4. Additional features of stm32h7 such as
configurable buffers are supported. Implementation offers feature parity
with stm32g4 implementation.


ventZl:
I'd like to hear comments on how I dealt with dissimilarity of G4 and H7 FDCAN at this stage before I proceed with implementing the rest.

include/libopencm3/stm32/fdcan.h
```diff
@@ -55,6 +40,7 @@ LGPL License Terms @ref lgpl_license
 #define FDCAN_FIFO1						1
 /**@}*/
 
+#define FDCAN_BLOCK_ID(can_base)		(((can_base) - CAN1)/(CAN2 - CAN1))
```
karlp, line 43: are you sure this is always a simple op? and not pulling in divide? (not as big a deal on "big" cpus as on small, but still, you don't want to be dividing a lot here?

include/libopencm3/stm32/fdcan.h
```diff
@@ -461,17 +414,14 @@ LGPL License Terms @ref lgpl_license
...
-#define FDCAN_TXFQS_TFGI_MASK			0x3
+#define FDCAN_TXFQS_TFGI_SHIFT			8
 
 /* TFQPI[1:0]: Tx FIFO put index */
-#define FDCAN_TXFQS_TFQPI_SHIFT			0
-#define FDCAN_TXFQS_TFQPI_MASK			0x3
+#define FDCAN_TXFQS_TFQPI_SHIFT			16
 
-#define FDCAN_TXFQS_TFQF				(1 << 0)
+#define FDCAN_TXFQS_TFQF				(1 << 21)
```
karlp, line 424: this looks like it was a standalone bug fix?

include/libopencm3/stm32/fdcan.h
```diff
@@ -792,6 +723,7 @@ struct fdcan_tx_buffer_element {
 #define FDCAN_FIFO_RXTS_SHIFT			0
 #define FDCAN_FIFO_RXTS_MASK			0xFFFF
 
+#if 0
```
karlp, line 726: you don't want to leave it like this.

include/libopencm3/stm32/h7/fdcan.h
```diff
@@ -0,0 +1,260 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this library.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef LIBOPENCM3_STM32G4_FDCAN_H
+#define LIBOPENCM3_STM32G4_FDCAN_H
```
karlp, line 30: to avoid this sort of junk, just use #pragma once instead.

karlp:
I mean, seems reasonably enough, but I have zero insight or testability on any of the can code, on either of these platforms :)

include/libopencm3/stm32/h7/fdcan.h
ventZl: Ok, I thought that pragmas are not used, because some ancient compiler which doesn't know them is still supported.

include/libopencm3/stm32/fdcan.h
ventZl: Yes, sorta. Somehow it never broke functionality of the old G4 code. Is it absolutely necessary to keep it in separate pull request?

include/libopencm3/stm32/fdcan.h
ventZl: Checked with godbolt, ARM32 GCC 10.2.1 none with -mcpu=cortex-m0 -mthumb -msoft-float -O0 yields only loads, stores and shifts/rotates. With -O2, it boils down to load, move, add and shift.

include/libopencm3/stm32/h7/fdcan.h
karlp: yeah, that wsa the belief, then I went and actually checked....  At least for new code," #pragma once" is all good to go.

include/libopencm3/stm32/fdcan.h
karlp: yay! I don't always have time to godbolt everything myself :) thanks for doing that.

karlp:
merged, I added some doxygen cleanup as well. that you may wish to review.

PR #1329, opened 2021-04-20
Flash get page size

jsphuebner:
I added a function that returns the page size for STM32F1 devices.
Tested with STM32F103RBT6, STM32F105RBT6 and STM32F107VCT6


jsphuebner:
Erm sorry, where did all these other commits come from?

karlp:
you need to rebase your branch on our master please :) that will make all the other ones go away.

lib/stm32/f1/flash.c
```diff
@@ -299,5 +299,27 @@ void flash_erase_all_pages(void)
...
+	
+	switch (idcode)
+	{
+		case 0x412:
+		case 0x410:
+			page_size = 1024;
+			break;
+		case 0x414:
+		case 0x418:
+		case 0x430:
```
karlp, line 317: at the very least, you want all these cases to be numerically ordered.

lib/stm32/f1/flash.c
```diff
@@ -299,5 +299,27 @@ void flash_erase_all_pages(void)
 	FLASH_CR2 &= ~FLASH_CR_MER;
 }
 
+/** @brief Get size in bytes of one flash page
+*/
+uint32_t flash_get_page_size(void)
+{
+	uint32_t idcode = MMIO32(DBGMCU_BASE) & 0x7FF;
```
karlp, line 306: just ```idcode == DBGMCU_IDCODE & 0x7ff``` is enough.

lib/stm32/f1/flash.c
karlp: Also, you're missing the f100 line, 0x420 and 0x428 (at least)

jsphuebner:
Alright, will make those changes.
That said, I tested on an older F103RBT7 today and the ID reads as 0
https://community.st.com/s/question/0D50X00009XkgRzSAJ/access-mcu-device-id-code-on-stm32f103
So not a reliable way to determine the page size on older silicon.
I have added a default branch now that returns 1024, but hmm. What to do?

lib/stm32/f1/flash.c
jsphuebner: These are the only IDs I found in the manual. Where did you find the others?

karlp:
this is why that sort of functionality hasn't been bothered with much before.  It's also rarely the sort of thing you need from an application library, normally something a debugger host reads.  I mean, I'm not _opposed_ to it, much, but it's missing from all the other families too, and f1 is the least interesting future proof part...

jsphuebner:
Well it becomes interesting for anything that writes to flash. Various openinverter firmware uses the last flash pages for saving application info and this helps to have it run on various devices without recompiling for each variant.
The current shortage in ST devices has forced us to relocate to other variants.

karlp:
yeah, I get it, but you can also see why it's never been anything that anyone's bothered implementing in the _library_ before.  (you definitely shouldn't try returning 1024 if it reads zero. you have no better information that it's correct, and far more rational to simply return zero.

lib/stm32/f1/flash.c
karlp: there's more than one manual :) if you're only looking at rm0008, you're not covering all of the f1s....  go back to st's website, click on "f1" and then click on docs, you can see the others :)

jsphuebner:
Yes I reckon we keep it out of the library then, gotta rethink this.

PR #1330, opened 2021-04-24
stm32: Define FLASH_BASE and PERIPH_BASE for all families

matthijskooijman:
FLASH_BASE was already defined for some and PERIPH_BASE for all but one,
but this makes these available for all families. Note that the value is
identical for all familes (I doublechecked the more exotic ones such
STM32H7), but it is still useful to have these defines to make code more
readable and so that libopencm3 users can write portable code without
having to check that these are identical on all STM32 families.

This fixes https://github.com/libopencm3/libopencm3-examples/issues/224 (which I accidentally reported in the wrong repository).


karlp:
sure, harmless enough :)

PR #1332, opened 2021-04-28
libopencmsis: typo stm32/l4 > stm32/g0

engineeror:


karlp:
fixed thanks, I've added some other entries missing there too.  I should note that the libopencmsis layers are unmaintained.

PR #1333, opened 2021-04-30
fix(lib/stm32/common/iwdg_common_all.c): Fix stm32-iwdg period set function bug

weycen:
Fix the bug that the iwdg counter is not refreshed after the configurationis complete, if this counter is not refreshed after the configuration is completed, the first iwdg counting period will be as long as 26 seconds.
We want the watchdog to take effect immediately after the period configuration is completed, so we need to refresh this counter immediately.

### DIFF
--- a/lib/stm32/common/iwdg_common_all.c
+++ b/lib/stm32/common/iwdg_common_all.c
@@ -108,6 +108,9 @@ void iwdg_set_period_ms(uint32_t period)
 	while (iwdg_reload_busy());
 	IWDG_KR = IWDG_KR_UNLOCK;
 	IWDG_RLR = count & COUNT_MASK;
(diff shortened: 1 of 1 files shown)


karlp:
Did you actually run into this?  What was your code sequence that led to this 26 second timeout?

I'm not... against it, I'm just looking at all my code that does a single iwdg_set_period() once, then does iwdg_start() and it "just works" ?  I've never really looked at reconfiguring it while it's running.  Presumably that's what you're doing, and as far as I can tell from teh docs it certainly doesn't hurt to call _reset() when it hasn't been started, so sure.... I guess?

karlp:
merged via rebase in 12da5bbd

PR #1335, opened 2021-05-11
Fix #1328: Make error return codes negative values

### LINKED ISSUE #1328: fdcan: error codes overlap with successful returns
In https://github.com/libopencm3/libopencm3/pull/1325, it was noticed that fdcan_transmit() (see https://github.com/libopencm3/libopencm3/blob/504dc95d9ba1c2505a30d575371accfe49a69fb9/lib/stm32/common/fdcan_common.c#L486) returns the positive mailbox id, _or_ the error code.  However the error codes are a simple enum, and are not returned negated, so there's no way to tell the difference between a success and an error.

I've not checked whether other methods have a similar issue.

ventZl:
Fix the bug where certain functions were returning meaningful return
and/or error code, where positive values of error codes were interfering
with meaningful return value. Error codes now have negative values as it
was originally intended but never implemented.

### DIFF
--- a/include/libopencm3/stm32/fdcan.h
+++ b/include/libopencm3/stm32/fdcan.h
@@ -723,28 +723,30 @@ struct fdcan_tx_buffer_element {
 #define FDCAN_FIFO_RXTS_MASK			0xFFFF
 
 
-/** FDCAN error return values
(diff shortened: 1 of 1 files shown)


karlp:
merged via rebase in 1033131e

PR #1336, opened 2021-05-12
Fix FDCAN incorrect FIFO acknowledgment

ventZl:
Fix incorrect way of acknowledging FIFO processing. Old code ORed old
value in register with index of FIFO buffer just processed, which
generated invalid value for acknowledge. This caused FIFO to repeatedly
return same content. Both fdcan_receive() and fdcan_release_fifo() were
affected.


karlp:
merged via rebase in 9478931d tahnks!

PR #1338, opened 2021-05-18
Fix STM32H7 FDCAN FIFO acknowledgment process

ventZl:
Fix FDCAN FIFO acknowledge register definition to make it correct for H7
MCUs. Previous definition contained hardcoded offset instead of using
MCU-specific macro.

Fix incorrect decoding of buffer element size. During decoding, value
returned was erratically set to 7 instead of setting 4th LSB. Buffer
element size was then always reported as 15 bytes.


karlp:
merged via rebase in e9c68ff9 (had to remove the duplicate retval stuff)

PR #1339, opened 2021-05-18
STM32H7: Implement support for H7 SPI peripheral

ventZl:
Provide actually working implementation of SPI for STM32 H7 family of
microcontrollers. Previous implementation assumed that SPI in H7 didn't
change, which is false. This implementation is derived from "v2" SPI
driver used for recent STM32 MCUs. There are API changes induced by
change in how SPI in H7 handles certain stuff, such as CRC.


lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+@endcode
+
+@todo need additional functions to aid ISRs in retrieving status
+
+*/
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2009 Uwe Hermann <uwe@hermann-uwe.de>
```
karlp, line 40: this doesn't match the earlier lines, and as this is a "v3" file, I'm not sure how anyone else could claim any copyright on it at all, or be granted.

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+@note The I2S protocol shares the SPI hardware so the two protocols cannot be
+used at the same time on the same peripheral.
+
+Example: Clk/4, positive clock polarity, leading edge trigger, 8-bit words,
+LSB first.
+@code
+	spi_init_master(SPI1, SPI_CR1_BAUDRATE_FPCLK_DIV_4, SPI_CR1_CPOL_CLK_TO_0_WHEN_IDLE,
+			SPI_CR1_CPHA_CLK_TRANSITION_1, SPI_CR1_DFF_8BIT, SPI_CR1_LSBFIRST);
+	spi_write(SPI1, 0x55);		// 8-bit write
+	spi_write(SPI1, 0xaa88);	// 16-bit write
```
karlp, line 28: what? this is C, how would that possibly happy?

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+/** @brief SPI Reset.
+
+The SPI peripheral and all its associated configuration registers are placed in
+the reset condition. The reset is effected via the RCC peripheral reset system.
+
+@param[in] spi_peripheral Unsigned int32. SPI peripheral identifier @ref
+spi_reg_base.
+*/
+
+void spi_reset(uint32_t spi_peripheral)
```
karlp, line 71: don't add this function.  We've dropped this double indirection in timers already, it has no business being added.

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+
+@todo Error handling?
+
+@param[in] spi Unsigned int32. SPI peripheral identifier @ref spi_reg_base.
+*/
+
+void spi_enable(uint32_t spi)
+{
+	SPI_CR1(spi) |= SPI_CR1_SPE; /* Enable SPI. */
+	SPI_CFG1(spi) |= SPI_CFG2_AFCNTR; /* Claim GPIO pins to be used by SPI */
```
karlp, line 122: should that be in the other order?  there's a few things related to NSS that happen when the SPE bit changing state don't they?

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+@param[in] spi Unsigned int32. SPI peripheral identifier @ref spi_reg_base.
+*/
+
+void spi_disable(uint32_t spi)
+{
+	uint32_t reg32;
+
+	reg32 = SPI_CR1(spi);
+	reg32 &= ~(SPI_CR1_SPE); /* Disable SPI. */
+	SPI_CR1(spi) = reg32;
```
karlp, line 140: just a SPI_CR1(spi) &= ~(SPI_CR1_SPE) is fine here.

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+	/* Wait to receive last data */
+	while (!(SPI_SR(spi) & SPI_SR_RXP));
+
+	uint16_t data = SPI_RXDR(spi);
+
+	/* Wait to transmit last data */
+	while (!(SPI_SR(spi) & SPI_SR_TXC));
+
+	/* Wait until not busy */
+	while ((SPI_SR(spi) & SPI_SR_EOT) == 0);
```
karlp, line 165: if you want to make some people happy (not really me, but others) you add in _blocking and is_() funcitons so that some of this stuff can be used more async...

karlp: (can be done later of course)

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+Data is written to the SPI interface.
+
+@param[in] spi Unsigned int32. SPI peripheral identifier @ref spi_reg_base.
+@param[in] data Unsigned int16. 8 or 16 bit data to be written.
+*/
+
+void spi_write(uint32_t spi, uint16_t data)
+{
+	/* Write data (8 or 16 bits, depending on DFF) into DR. */
+	SPI_TXDR(spi) = data;
```
karlp, line 184: so, what, -v3 goes back to v1 behaviour, where it's _only_ DFF? not the memory access size of the register?

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+(using a clock wire and a bidirectional data wire), and is placed in a receive
+state.
+
+@param[in] spi Unsigned int32. SPI peripheral identifier @ref spi_reg_base.
+*/
+
+void spi_set_bidirectional_receive_only_mode(uint32_t spi)
+{
+	SPI_CFG2(spi) = (SPI_CFG2(spi) & ~(SPI_CFG2_COMM_MASK << SPI_CFG2_COMM_SHIFT))
+		| (SPI_CFG2_COMM_HALF_DUPLEX << SPI_CFG2_COMM_SHIFT);
```
karlp, line 259: see this is where it _might_ hve been ok to use a split form :)

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+ * reception. If enabled, then length includes CRC. It is mandatory to set
+ * transfer length before CRC is enabled and data transfer is started.
+ *
+ * @param[in] spi SPI peripheral identifier @ref spi_reg_base.
+ * @param[in] size size of next frame. If given as 0xFFFF, then transfer is
+ * infinite. Cant be used with CRC enabled.
+ */
+void spi_set_transfer_size(uint32_t spi, uint16_t size)
+{
+	/* This code works and I don't have a clue why */
```
karlp, line 317: well, that's confidence inspiring :)

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,834 @@
...
+In slave mode, and only when software slave management is used, this replaces
+the NSS signal with a slave select disable signal.
+
+@param[in] spi Unsigned int32. SPI peripheral identifier @ref spi_reg_base.
+*/
+
+void spi_set_nss_low(uint32_t spi)
+{
+	SPI_CR1(spi) &= ~SPI_CR1_SSI;
+}
```
karlp, line 415: cmon, this is all in spi_common_all.  Why can't we extend that? Are there some "alL" functions that are not "v12"?  then we just do that.  We'v edone that before, we'll do it again..

slorquet:
Were there any progresses on these comments?

ventZl:
Internally, I have rewritten large portion of this patch in parallel with karlp's comments but did not manage to push / sync with comments here.

slorquet:
May I ask if you could do it, or is there any way I could help? Thanks.

ventZl:
In following days/weeks I have to unpack relevant HW so I will be able to test the code again. I'd also like to get this PR off my table sooner than later.

slorquet:
Hi,

I am now at the point where I'm trying to poke the STM32H7 SPI. My code is ready to get the jedec id from a serial flash.

Basic init/deinit seems to work but data transfer is of course broken, I just receive the byte I previously sent. I'm ready to work with anyone having started to develop this in a libopencm3 private branch in order to get STM32H7 SPI working and pushed upstream.

lib/stm32/h7/Makefile
```diff
@@ -46,7 +46,7 @@ OBJS += gpio_common_all.o gpio_common_f0234.o
 OBJS += pwr.o rcc.o
 OBJS += rcc_common_all.o
 OBJS += rng_common_v1.o
-OBJS += spi_common_all.o spi_common_v2.o
+OBJS += spi_common_v3.o
```
slorquet, line 49: spi_common_all.o should be retained so that spi_common_v3.o only implements v3 specific behaviour.

lib/stm32/common/spi_common_v3.c
slorquet: Handling this right now. The spiv3 file is heavily based on spiv2 with changes. Shall we discard all authors in the spiv3 file and keep only the original submitter of this pull request ? What is the usual method for that kind of thing? is is documented somewhere? Thanks.

lib/stm32/common/spi_common_v3.c
slorquet: Removed.

lib/stm32/common/spi_common_v3.c
slorquet: This is a comment documenting an example of use, that was copied from spiv3. Should we remove it from both spiv2 and spiv3 files? In a general rule, are there other files with a proper structure that we could use as template?

lib/stm32/common/spi_common_v3.c
slorquet: This function is already implemented in spi_common_all, except the SPI_CFG1 bit. How can we make that stm32h7 specific in the common file?

If this function is considered specific, how can we *replace* just the implementation of this function so we can benefit of the other common functions?

lib/stm32/common/spi_common_v3.c
slorquet: Done

lib/stm32/common/spi_common_v3.c
slorquet: No it's configurable, 4-16 or 4-32 (RM0433 page 2181, chapter 50.3)

lib/stm32/common/spi_common_v3.c
slorquet: "EOT is set by hardware as soon as a full transfer is complete"
"In master, EOT event terminates the data transaction and handles SS output optionally. In both master and slave, the event handles CRC transaction."
"EOT is cleared by software write 1 to EOTC bit at SPI2S_IFCR."

So if the previous transfer has completed, this bit is clear, probably to allow starting a new transfer? Not sure why this is required, this bit should be cleared before a transaction...

slorquet (commented):
I am continuing this work but I need input on how to render spi_common specific to a chip that does things differently. If this is not possible, then spi_common_all cannot be used at all.

PR #1345, opened 2021-06-29
Add STM32G051 support

stecman:
A couple of small tweaks were needed to work with the relatively new STM32G051 line:

- Add `devices.data` entries for STM32 G05 and G06 parts with 18K of RAM.

- Include the `dac_common_v1.h` API for STM32G0 parts. Tested and works correctly on the hardware. The registers defined in `stm32/common/dac_common_v1.h` match the STM32G0x1 reference manual.


karlp:
merged via rebase thanks! 6f9f40a7 and 3ab83bc07869dbb3291a721fdb9dc2b1c64a0ddf

PR #1349, opened 2021-07-17
STM32: Fixed ltdc_setup_windowing helper

bdebyl:
This PR adds some documentation, but primarily fixes the logic behind the incorrect function for setting up windowing. I've tested this on two different LCDs with the STM32F4. The fix adds two desired inputs to make it more clear to the user what `width` and `height` are, as opposed to incorrectly using and being given `active_width` and `active_height`


bdebyl:
Closed in favor of non-whitespace breaking: https://github.com/libopencm3/libopencm3/pull/1350

PR #1350, opened 2021-07-17
STM32: Fixed ltdc_setup_windowing helper

bdebyl:
This PR adds some documentation, but primarily fixes the logic behind the incorrect function for setting up windowing. I've tested this on two different LCDs with the STM32F4. The fix adds two desired inputs to make it more clear to the user what `width` and `height` are, as opposed to incorrectly using and being given `active_width` and `active_height`

### DIFF
--- a/lib/stm32/common/ltdc_common_f47.c
+++ b/lib/stm32/common/ltdc_common_f47.c
@@ -67,20 +67,25 @@ void ltdc_set_tft_sync_timings(uint16_t sync_width,    uint16_t sync_height,
 	LTDC_TWCR = (w << 16) | (h << 0);
 }
 
-void ltdc_setup_windowing(uint8_t  layer_number,
-			  uint16_t h_back_porch, uint16_t v_back_porch,
-			  uint16_t active_width, uint16_t active_height)
-{
-	active_width  += h_back_porch - 1;
-	active_height += v_back_porch - 1;
-	/*assert((h_back_porch & 0xfff == h_back_porch) &&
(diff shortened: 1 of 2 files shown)


lib/stm32/common/ltdc_common_f47.c
```diff
@@ -67,20 +67,25 @@ void ltdc_set_tft_sync_timings(uint16_t sync_width,    uint16_t sync_height,
...
-				     (h_back_porch << 0);
-	LTDC_LxWVPCR(layer_number) = (active_height << 16) |
-				     (v_back_porch << 0);
+/*---------------------------------------------------------------------------*/
+/** @brief LTDC Windowing Setup
+@param[in] layer_number unsigned int8. @ref ltdc_layer_num
+@param[in] h_back_porch unsigned int16. Horizontal Back Porch
+@param[in] v_back_porch unsigned int16. Vertical Back Porch
+@param[in] h_sync unsigned int16. Horizontal Sync
+@param[in] v_sync unsigned int16. Vertical Sync
```
karlp, line 76: personally, if the comments are just the variable name, I don't really see any gain.... (applies to porch/sync vars)  (also, in general, there's no reason for the type information, that's already in the decl)

bdebyl: I agree, although I followed the format in the dma documentation here even though it was vague it seems odd to omit some. Easy enough to remove these if you'd like.

karlp:
merged via rebase in 6763681c260cf280487d70ca0d1996a8b79fff30 thanks!

PR #1355, opened 2021-08-17
stm32: Extend the timer api 

Andreasdahlberg:
Add function to get the count direction bit. This is useful when the timer is in encoder interface mode and the DIR bit is modified by hardware.

### DIFF
--- a/include/libopencm3/stm32/common/timer_common_all.h
+++ b/include/libopencm3/stm32/common/timer_common_all.h
@@ -415,6 +415,7 @@ specific memorymap.h header before including this header file.*/
 @{*/
 #define TIM_CR1_DIR_UP			(0 << 4)
 #define TIM_CR1_DIR_DOWN		(1 << 4)
+#define TIM_CR1_DIR_MASK		(1 << 4)
 /**@}*/
 
 /* OPM: One pulse mode */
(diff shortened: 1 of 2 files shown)


include/libopencm3/stm32/common/timer_common_all.h
```diff
@@ -415,6 +415,7 @@ specific memorymap.h header before including this header file.*/
 @{*/
 #define TIM_CR1_DIR_UP			(0 << 4)
 #define TIM_CR1_DIR_DOWN		(1 << 4)
+#define TIM_CR1_DIR_MASK		(1 << 4)
```
karlp, line 418: Just drop this.

lib/stm32/common/timer_common_all.c
```diff
@@ -343,6 +343,21 @@ void timer_direction_down(uint32_t timer_peripheral)
...
+/** @brief Get the Timer count direction.
+
+Useful when the timer is in encoder interface mode and the DIR bit is modified
+by hardware.
+
+@param[in] timer_peripheral Unsigned int32. Timer register address base @ref
+tim_reg_base
+*/
+
+uint32_t timer_get_direction(uint32_t timer_peripheral)
```
karlp, line 356: just make it return a bool. Document the return parameters, so people have a clue of knowing which is which.

lib/stm32/common/timer_common_all.c
```diff
@@ -343,6 +343,21 @@ void timer_direction_down(uint32_t timer_peripheral)
...
+{
+	return TIM_CR1(timer_peripheral) & TIM_CR1_DIR_MASK;
```
karlp, line 358: and make this just be & TIM_CR1_DIR_DOWN

PR #1358, opened 2021-09-25
Introducing 96MHz SystemCoreClock for the F41x

cvrtky:


cvrtky:
A newer version of the so called blue pill / red pill minimal dev board using the F103 are now on sale on Aliexpress and other forums.
This new "black pill" is featuring the F411 which has a maximum system clock rate of 100 MHz.
The current standard clock setups has a 84MHz config for the F401 and a 168MHz for the F407 and the like, but no config for the F41x series. The F41x can of casue be run @ 84MHz, but why not utilize the higher clock if possible?
Running a F1x at 100MHz isn't optimal since the clock tree can't produce a PLLQ clock of 48MHz.
With a "systemCoreCock" of 96MHz this is possible hence this patch.

karlp:
I've merged this in f5813a54 but please follow existing style next time.  THat means whitespace, and comment style.

Issue #1343, opened 2021-05-31
stm32-l0 I2c blocks in 'i2c_transfer_complete'


matthijs040:
Hi. I'm working with the stm32-l041G6 in a DWM1004C module.

I am trying to port over I2c code I got working on my nucleo F303re from an old example project: 
https://github.com/libopencm3/libopencm3-examples/tree/master/examples/stm32/f3/stm32f3-discovery/i2c
I'm having issues with doing the same communication. It blocks at various stages in the 'i2c_transfer7' function.
(while(wait), while(i2c_nack) and currently while(!i2c_transfer_complete(i2c)) depending on at what frequency i initialize the peripheral)

```c
void i2c_transfer7(uint32_t i2c, uint8_t addr, const uint8_t *w, size_t wn, uint8_t *r, size_t rn)
{
	/*  waiting for busy is unnecessary. read the RM */
	if (wn) {
		i2c_set_7bit_address(i2c, addr);
		i2c_set_write_transfer_dir(i2c);
		i2c_set_bytes_to_transfer(i2c, wn);
		if (rn) {
			i2c_disable_autoend(i2c);
		} else {
			i2c_enable_autoend(i2c);
		}
		i2c_send_start(i2c);

		while (wn--) {
			bool wait = true;
			while (wait) {
				if (i2c_transmit_int_status(i2c)) {
					wait = false;
				}
				while (i2c_nack(i2c)); /* FIXME Some error */
			}
			i2c_send_data(i2c, *w++);
		}
		/* not entirely sure this is really necessary.
		 * RM implies it will stall until it can write out the later bits
		 */
		if (rn) {
			while (!i2c_transfer_complete(i2c));
		}
	}
        ...
}
```
I'm not sure what I'm doing wrong. I'm currently using the HSI16 clock for I2c and left all other settings untouched. (Which I think means that the MSI clock at 2MHz is the sysclk / peripheral clock for everything else?)
Before that I tried setting up the PLL clock as sysclk using the rcc_clock_setup_pll function and using sysclk to drive the peripheral at many frequencies. 

I also found the reference manual describing 'I2C clock requirements'. I am not sure what clock frequencies are valid after providing them to the i2c_set_speed function the API provides. I thought that that could cause an issue but don't know how to determine what valid frequencies are.

Here is my current  initialization code:
```c
	if( !rcc_is_osc_ready(RCC_HSI16) )
		rcc_osc_on(RCC_HSI16);

[...]

Issue #1026, opened 2019-02-20
Infinite loop in i2c_transfer7


kajusK:
Hi guys,
is there any specific reason why there's an infinite loop in i2c_transfer7 (file i2c_common_v2.c)?

while (i2c_nack(i2c)); /* FIXME Some error */

It's a really dumb idea to go into the endless loop when i2c device fails to respond. One dead i2c device will take the whole system down... Wouldn't it be better to just return false or something?

karlp:
lots of things will do infinite loop.  Feel free to suggest a patch for this if you have concrete changes.  Using a watchdog is also a great idea.

karlp:
if you're having a hard time understanding, try asking yourself what your proposal even is.  you even said yourself, "return false, or something"  So... what  _is_ the right thing to do here?  What should the API look like?

kajusK:
Please don't take the following text personal, but I would be glad if you read it whole and thing about that for a while.

Well, infinite loop in the library is nonsense, it's broken by desing. It must be up to a user to decide what to do when something goes wrong. Just imagine that the mcu controlls e.g. a drone and uses i2c to talk to something like accelerometer, gyroscope and pressure sensor. Accelerometer dies, therefore no ack arrives and the drone hits the earth/someone/someones property. Watchdog is pointless as the i2c code will be called after boot again, i2c stays in while loop, watchdog reboots the mcu, i2c stays in the while loop...

There are very few places where the infinite loop makes sense in the low level hardware abstraction library - like unhandled exception, stack overflow... The rest should be handled by user. I don't think the i2c nack is the best place for while loop as nack can happen for thousands of reasons (i2c device missing, busy processing the data,...) and user should expect and handle these. Killing the whole system/thread just for a single NACK is a horrible mistake that leas to dead device without a way to ressurect with some error msg or something. It should be up to a user to decide what to do.

Yes, nobody forces me to use that function, I could reimplement it myself, but does it make a sense to use the library at all if I have to rewrite function for multiple peripherals?

Well, what should the api look like? just change the function to return bool and if nack happens, return false. I can a pull request if you want, I'm just not sure what's the opinion of the core developers on api changes. But if this is not a rare exception, it should be brought into more public discussion and general change to error handling should be considered.

Thank you for your time.
Jakub

dzarda:
[...]

PR #1359, opened 2021-10-01
Reduce possibility of freezing in STM32 I2C v2

dzarda:
I2C transfer freezes on STM32 v2 probably whenever there's a NACK. This is unwanted behaviour. The freeze is triggered also when accessing a missing device (address invalid) as this also leads to a NACK.

I don't really know how to "properly" implement I2C and I'm not convinced this fixes all freezes in all cases. But should be beneficial.

The STOPF clearing seems to be needed as subsequent transactions seem to fail with the flag ON.

Possibly related:
- #1343
- #1026


slorquet (approved):
I2C on stm32 is tricky, if that change was tested to improve the global behaviour of I2C transactions I recommend applying it.

dzarda:
Since then I encountered another case when the transaction froze. But this definitely improved things.

Issue #409, opened 2015-02-21
USB MSC: Writes aren't acknowledged


Spritetm:
The USB MSC code has a bug: if a write request is sent, the MSC code will handle it but won't send an CSW back, leaving the host to wait forever. This is actually visible in the examples in the example repo: if you flash a msc example and try to write to the resulting drive, at least under Linux the device hangs.

Here's the fix I used to solve it. I'm not a USB expert by far, but as far as I can tell this fixes the issue:

``` diff
--- usb_msc.c   2015-01-28 17:38:50.275440725 +0100
+++ usb_msc_mine.c  2015-02-21 23:04:41.681417393 +0100
@@ -586,6 +588,19 @@ static void msc_data_rx_cb(usbd_device *
                trans->current_block++;
            }
        }
+
+       if (false == trans->csw_valid) {
+           scsi_command(ms, trans, EVENT_NEED_STATUS);
+           trans->csw_valid = true;
+       }
+       left = sizeof(struct usb_msc_csw) - trans->csw_sent;
+       if (0 < left) {
+           max_len = MIN(ms->ep_out_size, left);
+           p = &trans->csw.buf[trans->csw_sent];
+           len = usbd_ep_write_packet(usbd_dev, ms->ep_in, p, max_len);
+           trans->csw_sent += len;
+       }
+
    } else if (trans->byte_count < trans->bytes_to_write) {
        if (0 < trans->block_count) {
            if ((0 == trans->byte_count) && (NULL != ms->lock)) {
```

klenSA:
I confirm the problem under Linux, and the decision of the bug above
code
thanks Spritetm

Spritetm:
Whoah, that took a while. Thanks for taking care of this!

PR #1361, opened 2021-10-04
fix an occasional hangup with PR #409, and include PR #1294

stefaandesmet2003:
Hi,
I had occasional hangups with the mass storage class code, related to the changes made in PR #409. I'm using the usb_msc example, slightly modified to run on stm32f103. The hangup occurs occasionally after a write_10 by the host (can be triggered by mounting/unmounting several times, or adding a file to the storage device). After the write_10 the host is initiating an IN transaction to receive the CSW, but target hasn't prepared any data. 
I have noticed that just before the hangup, the code in https://github.com/libopencm3/libopencm3/compare/master...stefaandesmet2003:MSC_bugfix_PR409?expand=1#diff-dae3753d1fc8542f037be2cf8456abd0c8418f8b307f132c9f66db8ef679fcadL636-L651
 is trying to write to LBA=1, although the ongoing transaction was for a 1-sector write to LBA=0 (happens on other LBA too). This code is wrong and should be removed entirely. It is placed in an if-block that is supposed to handle transactions which don't require additional read/write, and only require a CSW response from the msc device. Furthermore lba + trans->block_count is the block too far to write, and thus corrupting the storage device.
the code from PR #409 copies the CSW 'all-good' response to packet memory on the first msc_data_rx_cb (ie. after the first 64 bytes of data to write have been read from the ep-out; it is expecting at least 512). My code postpones this action till all bytes from the transaction have been received and written to the storage device, and solves the occasional hangup entirely.
https://github.com/stefaandesmet2003/libopencm3/blob/MSC_bugfix_PR409/lib/usb/usb_msc.c#L790-L805

The other diffs are related to PR #1294 merged here, because this PR is useful and works correctly.
Thanks for considering this PR,
Cheers,
Stefaan

### DIFF
--- a/include/libopencm3/usb/msc.h
+++ b/include/libopencm3/usb/msc.h
@@ -78,6 +78,23 @@ typedef struct _usbd_mass_storage usbd_mass_storage;
 #define USB_MSC_REQ_BULK_ONLY_RESET	0xFF
 #define USB_MSC_REQ_GET_MAX_LUN		0xFE
 
+/* SBC-2, 5.1.22 START STOP UNIT, Table 52 */
+#define USB_MSC_SBC_2_POWER_CONDITION_START		0x10	// Additional code.
+#define USB_MSC_SBC_2_POWER_CONDITION_STOP		0x20	// Additional code.
+#define USB_MSC_SBC_2_POWER_CONDITION_ACTIVE		0x01
+#define USB_MSC_SBC_2_POWER_CONDITION_IDLE		0x02
+#define USB_MSC_SBC_2_POWER_CONDITION_STANDBY		0x03
+#define USB_MSC_SBC_2_POWER_CONDITION_SLEEP		0x05
+#define USB_MSC_SBC_2_POWER_CONDITION_DEVICE		0x07
+#define USB_MSC_SBC_2_POWER_CONDITION_DEVICE_IDLE	0x0A
+#define USB_MSC_SBC_2_POWER_CONDITION_DEVICE_STANDBY	0x0B
+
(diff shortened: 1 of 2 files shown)


karlp:
At the very least, this has style/whitespace issues that you should fix.

Secondly, this sort of change is _screaming_ for some test cases.  Can we get some scripted tests that can be used to verify that the MSC class works?  I'll accept anything, but ideally the same sort of python tests that are already used for the usb gadget zero.

You don't have to test everything, just even getting _some_ tests that cover what's broken, and what is then fixed, would make things massively easier to manage down the road

stefaandesmet2003:
Thanks for the quick reply.
my editor is using spaces for tabs; I guess that's where the styling issue comes from. I need to use tabs (\t), is that right?
Good proposal.
I will come up with a scripted test and firmware example that proves the current implementation is broken. And add the files under ./tests/usb_msc ?
The hangup is timing related. PC host sends OUT packet for the next scsi command before mcu has handled the IN callback following CSW, and usbd_poll handles the OUT first, which is then incorrectly interpreted as part of the previous transaction.

karlp:
This still needs the actual test code, not just the test target, but the code that runs the test.  It can be a script that returns 0/1, or ideally more of the same sort of python test runners that are in the usb gadget zero tree.

You're actual problem sounds like it might just be a usb core issue, and perhaps easier to trigger there with in/outs.

(minor, but you shouldnt' commit the generated linker script, and you need to fix the style of any test code too, and you should be able to share a rules.mk, instead of copying it again)

PR #1363, opened 2021-11-04
stm32l4: define CAN2 address and RCC bits

Niautanor:
Some STM32L4s have a second CAN instead of the USB FS peripheral. This merge request defines the address of the peripheral and the RCC bits to support it.

As far as I am aware, this will not have any negative effects on STM32L4s that do not have the second CAN (except a tiny increase in code size in can_reset).


karlp:
thanks!

PR #1364, opened 2021-11-09
devices.data: add samd09?13* and samd21?18*

chronopoulos:


karlp:
merged thanks! I added a few more from the same line to be more complete as well

PR #1365, opened 2021-11-10
Added missing bit definitions to f4/pwr.h

kapacuk:
Added definitions for `PWR_CR_ADCDC1`, `PWR_CR_MRLVDS`, and `PWR_CR_LPLVDS` to f4/pwr.h.


karlp:
merged via rebase in 3e480eecaeaa013bf329231543fa729d61d57484 thanks!  pity ST used different names for them in some earlier parts, but so be it, I've added compat aliases for both names.

PR #1368, opened 2021-11-28
Implement PLL helper as seen in other MCU families

vladlogyin:
The L4 header seemed to be missing the rcc_clock_setup_pll helper that can be found for other MCU families like the F1, F4, L1 etc. Only 2 HSI16 configs are included currently.

### DIFF
--- a/include/libopencm3/stm32/l4/rcc.h
+++ b/include/libopencm3/stm32/l4/rcc.h
@@ -33,22 +33,30 @@
  * along with this library.  If not, see <http://www.gnu.org/licenses/>.
  *
  */
-
 /**@{*/
 
 #ifndef LIBOPENCM3_RCC_H
 #define LIBOPENCM3_RCC_H
 
-/* --- RCC registers ------------------------------------------------------- */
+#include <libopencm3/stm32/pwr.h>
 
+/** @defgroup rcc_registers RCC Registers
+ * @brief Reset / Clock Control Registers
+ @{*/
+ /** Clock control register */
 #define RCC_CR				MMIO32(RCC_BASE + 0x00)
 #define RCC_ICSCR			MMIO32(RCC_BASE + 0x04)
+ /** Clock Configuration register */
 #define RCC_CFGR			MMIO32(RCC_BASE + 0x08)
+/** PLL Configuration register */
 #define RCC_PLLCFGR			MMIO32(RCC_BASE + 0x0c)
 #define RCC_PLLSAI1_CFGR		MMIO32(RCC_BASE + 0x10)
 #define RCC_PLLSAI2_CFGR		MMIO32(RCC_BASE + 0x14)
+/** Clock interrupt enable register */
 #define RCC_CIER			MMIO32(RCC_BASE + 0x18)
+/** Clock interrupt flag resiger */
 #define RCC_CIFR			MMIO32(RCC_BASE + 0x1c)
+/** Clock interrupt clear register */
 #define RCC_CICR			MMIO32(RCC_BASE + 0x20)
 #define RCC_AHB1RSTR_OFFSET		0x28
 #define RCC_AHB1RSTR			MMIO32(RCC_BASE + RCC_AHB1RSTR_OFFSET)
@@ -87,12 +95,17 @@
 #define RCC_APB2SMENR_OFFSET		0x80
 #define RCC_APB2SMENR			MMIO32(RCC_BASE + RCC_APB2SMENR_OFFSET)
 #define RCC_CCIPR			MMIO32(RCC_BASE + 0x88)
+/** Backup Domain control register */
 #define RCC_BDCR			MMIO32(RCC_BASE + 0x90)
+/** Clock control and status register */
 #define RCC_CSR				MMIO32(RCC_BASE + 0x94)
 #define RCC_CRRCR			MMIO32(RCC_BASE + 0x98)
+/** @}*/
 
-/* --- RCC_CR values ------------------------------------------------------- */
-
+/** @defgroup rcc_cr_values RCC_CR values
+ * @ingroup rcc_registers
+ * @brief Clock Control register values
+ @{*/
 #define RCC_CR_PLLSAI2RDY			(1 << 29)
 #define RCC_CR_PLLSAI2ON			(1 << 28)
 #define RCC_CR_PLLSAI1RDY			(1 << 27)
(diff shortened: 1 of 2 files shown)


lib/stm32/l4/rcc.c
```diff
@@ -38,12 +38,51 @@
...
+    .ppre1 = RCC_CFGR_PPRE_NODIV,
+    .ppre2 = RCC_CFGR_PPRE_NODIV,
+    .voltage_scale = PWR_SCALE1,
+    .flash_config = FLASH_ACR_DCEN | FLASH_ACR_ICEN |
+    FLASH_ACR_LATENCY_4WS,
+    .ahb_frequency  = 72000000,
+    .apb1_frequency = 72000000,
+    .apb2_frequency = 72000000,
+  },
+};
```
karlp, line 84: you have some whitespace issues here at least, possible elsewhere as well too.

lib/stm32/l4/rcc.c
```diff
@@ -341,6 +380,83 @@ uint32_t rcc_system_clock_source(void)
...
+  rcc_set_ppre1(clock->ppre1);
+  rcc_set_ppre2(clock->ppre2);
+  
+  /* Disable PLL oscillator before changing its configuration. */
+  rcc_osc_off(RCC_PLL);
+  
+  /* Configure the PLL oscillator. */
+  rcc_set_main_pll(clock->pll_source, clock->pllm, clock->plln,
+                         clock->pllp, clock->pllq, clock->pllr);
+
```
karlp, line 424: all whitespace here is off as well, you need to use the normal hard tab styles please

lib/stm32/l4/rcc.c
```diff
@@ -341,6 +380,83 @@ uint32_t rcc_system_clock_source(void)
...
+  }
+  flash_set_ws(clock->flash_config);
+  
+  /* Select PLL as SYSCLK source. */
+  rcc_set_sysclk_source(RCC_CFGR_SW_PLL);
+  
+  /* Wait for PLL clock to be selected. */
+  rcc_wait_for_sysclk_status(RCC_PLL);
+  
+  return;
```
karlp, line 448: wat?

karlp:
THis looks generally good

lib/stm32/l4/rcc.c
vladlogyin: Thank you for taking your time with this. I'm a real newbie when it comes to VCS and contributing to others' code. This function I got line for line from lib/stm32/f4/rcc.c:789. I definitely forgot to make my formatting consistent. Should I fix the formatting and add another commit or should I alter the history? (again, I'm not perfectly familiar with the etiquette)

karlp: more commits is fine if the end result is just one feature.  I can just squash it when I merge it.    Where I really ask for separate commits is where you have multiple separate functional changes.

include/libopencm3/stm32/l4/rcc.h
```diff
@@ -750,218 +788,218 @@ enum rcc_osc {
...
-	RCC_TSC = _REG_BIT(RCC_AHB1ENR_OFFSET, 16),
-	RCC_CRC = _REG_BIT(RCC_AHB1ENR_OFFSET, 12),
-	RCC_FLASH = _REG_BIT(RCC_AHB1ENR_OFFSET, 8),
-	RCC_DMA2 = _REG_BIT(RCC_AHB1ENR_OFFSET, 1),
-	RCC_DMA1 = _REG_BIT(RCC_AHB1ENR_OFFSET, 0),
+	RCC_TSC = _REG_BIT(0x48, 16),
+	RCC_CRC = _REG_BIT(0x48, 12),
+	RCC_FLASH = _REG_BIT(0x48, 8),
+	RCC_DMA2 = _REG_BIT(0x48, 1),
+	RCC_DMA1 = _REG_BIT(0x48, 0),
```
karlp, line 795: please restore these, as I commented on the earlier verson.

lib/stm32/l4/rcc.c
```diff
@@ -38,12 +38,51 @@
...
+		.ppre1 = RCC_CFGR_PPRE_NODIV,
+		.ppre2 = RCC_CFGR_PPRE_NODIV,
+		.voltage_scale = PWR_SCALE1,
+		.flash_config = FLASH_ACR_DCEN | FLASH_ACR_ICEN |
+		FLASH_ACR_LATENCY_4WS,
+		.ahb_frequency  = 80000000,
+		.apb1_frequency = 80000000,
+		.apb2_frequency = 80000000,
+	},
+	{ /* 72MHz PLL from HSI16 VR1 */
```
karlp, line 67: 80MHz is max speed, what is 72 for?  We generally try and avoid having too many of these defined, only max for each part really.  Is this intended for L476 and L475 that have the dwc_otg usb core, to be max core speed that allows a 48MHz USB clock from HSE?

lib/stm32/l4/rcc.c
```diff
@@ -341,6 +380,83 @@ uint32_t rcc_system_clock_source(void)
...
+	flash_set_ws(clock->flash_config);
+
+	/* Select PLL as SYSCLK source. */
+	rcc_set_sysclk_source(RCC_CFGR_SW_PLL);
+
+	/* Wait for PLL clock to be selected. */
+	rcc_wait_for_sysclk_status(RCC_PLL);
+
+	return;
+	/* Set the peripheral clock frequencies used. */
```
karlp, line 449: my "wat" here was you have code after the return statement.  this has no effect, so it's probably not doing what you think...

lib/stm32/l4/rcc.c
```diff
@@ -509,7 +626,8 @@ uint32_t rcc_get_i2c_clk_freq(uint32_t i2c)
 /** @brief Get the peripheral clock speed for the SPI device at base specified.
  * @param spi  Base address of SPI device to get clock frequency for (e.g. SPI1_BASE).
  */
-uint32_t rcc_get_spi_clk_freq(uint32_t spi) {
+uint32_t rcc_get_spi_clk_freq(uint32_t spi)
+{
```
karlp, line 630: You're right, this is wrongly formatted, but it's unrelated to your code, and so generally you should not include this.  It makes it easier to apply just your change, to only touch your code.  I can handle this for you though.

lib/stm32/l4/rcc.c
vladlogyin: That was my main logic behind it.

karlp:
merged in 3122c0b3 thanks!  I dropped the 72Mhz, you can't run USB from HSI anyway, according to the docs, it has dedicated plls for doing it frrom MSI, if you have an LSE crystal attached, or you need to use HSE.

PR #1369, opened 2021-11-29
README: added info about partial build, take 2

schodet:
See https://github.com/libopencm3/libopencm3/pull/1287

- split white space change in another commit,
- remove list of targets from README to avoid duplicated work and add a list-targets in Makefile,
- replace “footprint of the library” by “build time”, as someone could wrongly understand that the compiled size of the lib depends on the list of built targets.


karlp:
merged thanks!

PR #1370, opened 2021-12-01
stm32: fix typo for auto baud rate request bit in USART_RQR

whg:
Just a small change to fix a wrong character


karlp:
great thanks!

PR #1374, opened 2022-01-04
make sure dev->pm_top is even

j1rie:
Working on a hid device with rx and tx size 17 bytes, setting up tx first did not work, while rx first did.
Setting up rx first increases dev->pm_top by 18, so tx address starts at even, works.
Setting up tx first increases dev->pm_top by 17, so rx address starts at odd, doesn't work.
Packet memory is 16-bit aligned, so an address needs to be even. See the comment for Bit 0 of ADDRn_RX and ADDRn_TX in RM0091 and RM0008.
USB_SET_EP_TX_ADDR and USB_SET_EP_RX_ADDR rely on dev->pm_top being even, when st_usbfs_ep_setup() is called again.
So if the increment size is odd, increase dev->pm_top by the next even (18 instead of 17 in this example).
Double of https://github.com/libopencm3/libopencm3/issues/537 and part of https://github.com/libopencm3/libopencm3/pull/628/commits/7224576eeb20eed4f6811dd679c2ee591eedc96b
This bug and the fix are known but not accepted since 6 years, so if you need more information to accept this, I am willing to clarify.

### DIFF
--- a/lib/stm32/common/st_usbfs_core.c
+++ b/lib/stm32/common/st_usbfs_core.c
@@ -104,7 +104,13 @@ void st_usbfs_ep_setup(usbd_device *dev, uint8_t addr, uint8_t type,
 		}
 		USB_CLR_EP_TX_DTOG(addr);
 		USB_SET_EP_TX_STAT(addr, USB_EP_TX_STAT_NAK);
-		dev->pm_top += max_size;
+		/* Packet memory is 16-bit aligned, so an address needs to be even.
+		 * See the comment for Bit 0 of ADDRn_RX and ADDRn_TX in RM0091, RM0008.
+		 * USB_SET_EP_TX_ADDR and USB_SET_EP_RX_ADDR rely on dev->pm_top being even,
+		 * when st_usbfs_ep_setup() is called once more after this instance.
+		 * So if max_size is odd, increase dev->pm_top by (max_size + 1).
+		 */
+		dev->pm_top += ((max_size + 1) >> 1) << 1;
 	}
 
 	if (!dir) {


karlp:
If you want to make usb changes, you need to make tests for it.  See the gadget 0 tests for inspiration.

j1rie:
And all tests:
python test_gadget0.py
Running (user) tests for DUT:  stm32f103-generic
............read 5324800 bytes in 0:00:07.021150 for 740.6194141985288 kps
.wrote 5324800 bytes in 0:00:06.510841 for 798.6679447401649 kps
..................

Ran 31 tests in 14.190s

OK

karlp:
It's a good start! glad the docs were enough to let you get them running!  Now, I need you to _add_ tests to catch what you are saying is broken, because those tests passed before as well...

j1rie:
For bug demonstration I changed usb-gadget0.c (diff at end of post):
BULK_EP_MAXPACKET from 64 to 61
bMaxPacketSize from BULK_EP_MAXPACKET to 64 (actually no change)

Running python test_gadget0.py with my patch gives result:
Running (user) tests for DUT:  stm32f103-generic
............ss.................

Ran 31 tests in 0.637s

OK (skipped=2)

But running python test_gadget0.py without my patch gives result:
Running (user) tests for DUT:  stm32f103-generic
FF..........ss.................

FAIL: test_dual_loop (__main__.TestConfigLoopBack)
Testing that we don't mix our data up, just plain and simple

Traceback (most recent call last):
  File "test_gadget0.py", line 279, in test_dual_loop
    self._inner_basic(epo, epi, data)
  File "test_gadget0.py", line 261, in _inner_basic
    self.assertEqual(expected, read, "should have read back what we wrote")
AssertionError: array('B', [187, 187, 187, 187, 187, 187, 187, 187, 1[259 chars]187]) != array('B', [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,[137 chars], 0]) : should have read back what we wrote

======================================================================
FAIL: test_dual_loop_back_to_back (__main__.TestConfigLoopBack)

Traceback (most recent call last):
  File "test_gadget0.py", line 304, in test_dual_loop_back_to_back
    self.assertEqual(expected, r, "should have read back what we wrote")
AssertionError: array[249 chars]0, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170]) != array[249 chars]0, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 170, 0]) : should have read back what we wrote


Ran 31 tests in 0.638s

FAILED (failures=2, skipped=2)

Setting up endpoint 0x81 spoils dev->pm_top, after that setting up endpoint 0x82 starts at odd, creates problem.
You only see this, when you work with odd sizes, not when you use 64 for everything.

For clarity the change as diff:
```
--- usb-gadget0.c.B     2022-01-07 23:33:47.220172322 +0100
+++ usb-gadget0.c       2022-01-07 23:26:53.478771186 +0100
@@ -55,7 +55,7 @@
 #define GZ_CFG_SOURCESINK      2
 #define GZ_CFG_LOOPBACK                3

-#define BULK_EP_MAXPACKET      64
+#define BULK_EP_MAXPACKET      61 // JR 64

 static const struct usb_device_descriptor dev = {
        .bLength = USB_DT_DEVICE_SIZE,
@@ -64,7 +64,7 @@ static const struct usb_device_descripto
        .bDeviceClass = USB_CLASS_VENDOR,
        .bDeviceSubClass = 0,
        .bDeviceProtocol = 0,
-       .bMaxPacketSize0 = BULK_EP_MAXPACKET,
+       .bMaxPacketSize0 = 64, // JR BULK_EP_MAXPACKET,

        /* when we're compatible with gadget 0
         * #define DRIVER_VENDOR_NUM       0x0525
```

j1rie:
May I ask about your considerations how to proceed?

PR #1389, opened 2022-02-25
added additional configuration bits for STM32G4 series timer 1,8 and 20

qosch:
Trying to configure the TRGO2 signal present in the advanced timers of the STM32G4 series I could not find the corresponding register, the MMS2 register (neither in timer_common_all.h nor in timer_common_f24.h). This pull request adds it, along all other changes of this register compared to STM32F2 and STM32F4. (see STM32G4 reference manual p. 1170)
The MMS register has a 4th bit added that is separated from the other 3 bits. I am not sure I handled this the way it should be (see line 61 and 62), feel free to suggest a better solution.
I have also no idea whether doxygen will get along with what I did.

Open for any suggestions!

### DIFF
--- a/include/libopencm3/stm32/g4/timer.h
+++ b/include/libopencm3/stm32/g4/timer.h
@@ -29,3 +29,51 @@
 
 #pragma once
 #include <libopencm3/stm32/common/timer_common_f24.h>
+
+/* --- TIMx_CR2 values ----------------------------------------------------- */
+
+/****************************************************************************/
+/** @defgroup tim_x_cr2_ois TIMx_CR2_OIS: Force Output Idle State Control Values
+@{*/
+/* OIS4N:*//** Output idle state 4 (OC4N output) */
+#define TIM_CR2_OIS4N			(1 << 15)
+
+/* OIS5:*//** Output idle state 5 (OC5 output) */
+#define TIM_CR2_OIS4N			(1 << 16)
+
+/* OIS6:*//** Output idle state 6 (OC6 output) */
+#define TIM_CR2_OIS4N			(1 << 18)
+#define TIM_CR2_OIS_MASK		(0xfff << 8)
+/**@}*/
+
+/* MMS[3:0]: Master mode selection */
+/****************************************************************************/
+/** @defgroup tim_mastermode TIMx_CR2 MMS[25:4]: Master Mode Selection
+@{*/
+#define TIM_CR2_MMS_RESET		(0x0 << 4)
+#define TIM_CR2_MMS_ENABLE		(0x1 << 4)
+#define TIM_CR2_MMS_UPDATE		(0x2 << 4)
+#define TIM_CR2_MMS_COMPARE_PULSE	(0x3 << 4)
+#define TIM_CR2_MMS_COMPARE_OC1REF	(0x4 << 4)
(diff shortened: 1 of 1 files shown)


karlp:
I'm guessing these timers are also available on l4, u5, and friends too, so I'd rather add it as a newer peripheral commonfile, rather than into each family one by one.

qosch:
Alright.
That would mean a new file  include/libopencm3/stm32/common/timer_common_v2.h that would include timer_common_f24.h or timer_common_all.h from the same directory?

karlp:
yeah, sounds about right.

include/libopencm3/stm32/g4/timer.h
```diff
@@ -29,3 +29,51 @@
...
+#define TIM_CR2_MMS_RESET		(0x0 << 4)
+#define TIM_CR2_MMS_ENABLE		(0x1 << 4)
+#define TIM_CR2_MMS_UPDATE		(0x2 << 4)
+#define TIM_CR2_MMS_COMPARE_PULSE	(0x3 << 4)
+#define TIM_CR2_MMS_COMPARE_OC1REF	(0x4 << 4)
+#define TIM_CR2_MMS_COMPARE_OC2REF	(0x5 << 4)
+#define TIM_CR2_MMS_COMPARE_OC3REF	(0x6 << 4)
+#define TIM_CR2_MMS_COMPARE_OC4REF	(0x7 << 4)
+#define TIM_CR2_MMS_ENCODER_OUTPUT	(0x200000 << 4)
+#define TIM_CR2_MMS_MASK		(0x200007 << 4)
```
karlp, line 62: Standard style for the library is to have defines unshifted, and use a _SHIFT define where required.

lib/stm32/common/timer_common_u5l4g04.c
```diff
@@ -0,0 +1,42 @@
+/** @addtogroup timer_file
```
karlp, line 1: plesae use a name like "-v2" or "v3" depending on your interpretation of timer generations.  trying to use part names like this in filenames has been demonstrably unmaintainable for years now :)

lib/stm32/common/timer_common_u5l4g04.c
```diff
@@ -0,0 +1,42 @@
...
+
+*/
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2010 Edward Cheeseman <evbuilder@users.sourceforge.org>
+ * Copyright (C) 2011 Stephen Caudle <scaudle@doceme.com>
```
karlp, line 9: None of this file is copyright either of those people in those years.....

qosch:
Ah sorry, this wasnt meant for review yet. I just committed to continue work in a different device. Most of your comments are still valid, though, so thanks :)

qosch:
I had to modify the function timer_set_master_mode() as well as the defines TIM_CR2_OIS_MASK and TIM_CR2_MMS_MASK to add support for the new CR2 bits and have warnings because of these redefinitions now. How'd you me like to handle this?
Also, please have a look at the new timer_set_master_mode() function. The 4 bit MMS bits are split up in the register and I couldn't find an example somewhere else how this should be implemented.
I'm open for any suggestions :)

PR #1391, opened 2022-03-03
LM4F: Make SYSCTL enum definitions C++ compliant

### LINKED ISSUE #1385: LM4F: Doesn't compile in C++ due to non-constexpr enum values in systemcontrol.h
`lm4f_clken` in `systemcontrol.h` describes a number of enum values with the form:

```RCC_WD0 = ((uint32_t)&SYSCTL_RCGCWD - SYSCTL_BASE) << 5```

In C++, casting `(volatile uint32_t*)` (which is the type of SYSCTL_RCGCWD due to MMIO32) to `uint32_t` ends up in a call to `reinterpret_cast` which is not `constexpr`. Therefore, these values are not valid in the definition of an `enum`. As GCC explains:

```
/home/ktims/.platformio/packages/framework-libopencm3/include/libopencm3/lm4f/systemcontrol.h:493:23: error: 'reinterpret_cast<volatile uint32_t* {aka volatile long unsigned int*}>(1074783744)' is not a constant expression
  493 |  RCC_WD0 = ((uint32_t)&SYSCTL_RCGCWD - SYSCTL_BASE) << 5,
/home/ktims/.platformio/packages/framework-libopencm3/include/libopencm3/lm4f/systemcontrol.h:493:56: error: enumerator value for 'RCC_WD0' is not an integer constant
  493 |  RCC_WD0 = ((uint32_t)&SYSCTL_RCGCWD - SYSCTL_BASE) << 5,
```

I don't think there is a clean way to cast pointer to integer that *is* `constexpr`, so perhaps the cleanest approach is to provide integer-type defines as well as pointer-type defines that are based on them.

ktims:
Refactor definitions to remove pointer-to-int casts.

Fixes #1385


karlp:
thanks!

PR #1394, opened 2022-03-16
Some TI and STM stuff

stefanct:
See commit messages. The only one requiring review is the one for the stm32l4 IMHO but be aware that I am kind of new to libopencm3.

### DIFF
--- a/lib/stm32/l4/rcc.c
+++ b/lib/stm32/l4/rcc.c
@@ -435,9 +435,10 @@ void rcc_set_rtc_clock_source(enum rcc_osc clk)
 	}
 }
 
-/* Helper to calculate the frequency of a UART/I2C based on the apb and clksel value. */
-static uint32_t rcc_uart_i2c_clksel_freq_hz(uint32_t apb_clk, uint8_t shift) {
-	uint8_t clksel = (RCC_CCIPR >> shift) & RCC_CCIPR_USARTxSEL_MASK;
+/* Helper to calculate the frequency of a UART/I2C based on the apb and clksel value.
+ * For I2C, clock selection 0b11 is reserved while it specifies LSE for UARTs. */
+static uint32_t rcc_uart_i2c_clksel_freq_hz(uint32_t apb_clk, uint8_t shift, uint32_t clock_reg) {
+	uint8_t clksel = (clock_reg >> shift) & RCC_CCIPR_USARTxSEL_MASK;
 	uint8_t hpre = (RCC_CFGR >> RCC_CFGR_HPRE_SHIFT) & RCC_CFGR_HPRE_MASK;
 	switch (clksel) {
 		case RCC_CCIPR_USARTxSEL_APB:
@@ -446,6 +447,8 @@ static uint32_t rcc_uart_i2c_clksel_freq_hz(uint32_t apb_clk, uint8_t shift) {
 			return rcc_ahb_frequency * rcc_get_div_from_hpre(hpre);
 		case RCC_CCIPR_USARTxSEL_HSI16:
 			return 16000000U;
+		case RCC_CCIPR_USARTxSEL_LSE:
+			return 32768U;
 	}
 	cm3_assert_not_reached();
 }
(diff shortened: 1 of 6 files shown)


lib/stm32/l4/rcc.c
```diff
@@ -40,6 +40,8 @@
 #include <libopencm3/stm32/rcc.h>
 
 /* Set the default clock frequencies after reset. */
+/* FIXME: These are never updated.
+ * They depend on RCC_CFGR, RCC_PLLCFGR, RCC_CR's MSIRANGE, etc. */
```
karlp, line 44: drop this please, this is library wide

stefanct: I have not looked around too much, but on one of my other targets, the stm32**f**4, the global variables are updated according to the pll and mux settings of the library calls, cf. https://github.com/libopencm3/libopencm3/blob/e7cf9362efe7d7cb6d225cede08252b990df558a/lib/stm32/f4/rcc.c#L851

I am a bit puzzled. I can of course remove the fixme from this patch but can you please elaborate on the frequency handling: Is the stm32f4 code wrong? Does this fixme apply to most/many/almost all of the other targets? Is this tracked anywhere and is there an established way to add frequency getters?

lib/stm32/l4/rcc.c
```diff
@@ -479,7 +482,20 @@ uint32_t rcc_get_usart_clk_freq(uint32_t usart)
...
+		uint8_t clksel = (RCC_CCIPR >> shift) & RCC_CCIPR_LPTIMxSEL_MASK;
+		switch (clksel) {
+			case RCC_CCIPR_LPTIMxSEL_APB:
+				return rcc_apb1_frequency;
+			case RCC_CCIPR_LPTIMxSEL_LSI:
+				return 32000;
+			case RCC_CCIPR_LPTIMxSEL_HSI16:
+				return 16000000U;
+			case RCC_CCIPR_LPTIMxSEL_LSE:
+				return 32768;
```
karlp, line 496: why have you mixed explicitly unsigned with signed here?

lib/stm32/l4/rcc.c
```diff
@@ -458,17 +463,15 @@ uint32_t rcc_get_usart_clk_freq(uint32_t usart)
...
-		return rcc_uart_i2c_clksel_freq_hz(rcc_apb1_frequency, RCC_CCIPR_USART1SEL_SHIFT);
+		return rcc_uart_i2c_clksel_freq_hz(rcc_apb2_frequency, RCC_CCIPR_USART1SEL_SHIFT, RCC_CCIPR);
 	} else if (usart == USART2_BASE) {
-		return rcc_uart_i2c_clksel_freq_hz(rcc_apb1_frequency, RCC_CCIPR_USART2SEL_SHIFT);
+		return rcc_uart_i2c_clksel_freq_hz(rcc_apb1_frequency, RCC_CCIPR_USART2SEL_SHIFT, RCC_CCIPR);
 	} else if (usart == USART3_BASE) {
-		return rcc_uart_i2c_clksel_freq_hz(rcc_apb1_frequency, RCC_CCIPR_USART3SEL_SHIFT);
-	} else if (usart == UART4_BASE) {
-		return rcc_uart_i2c_clksel_freq_hz(rcc_apb1_frequency, RCC_CCIPR_UART4SEL_SHIFT);
-	} else {  /* USART5 */
```
karlp, line 470: please don't.  What makes you think uart5 doesn't exist?  See RM0351 for instance?

lib/stm32/l4/rcc.c
karlp: If there's a clock setter routine, it should update those variables, if not, that's an independent bug, and not something I want comments scattered around for.

stefanct:
Thanks for the review. I'll get back to you at the end of next week, got some deadline...

lib/stm32/l4/rcc.c
stefanct: OK, so I'll remove the fixme and report it as extra bug on github(?)

lib/stm32/l4/rcc.c
stefanct: It's copy and paste and I think I read it as UL because that would somewhat make sense. Just U does not really help here AFAICT. With UL it makes clear that even on 16-bit machines the value should not be cut (not sure what the rules would be when the literal is directly used in a return statement of type `uint32_t` though), something that cannot happen with 32768 at all. However, that's not really relevant to this library anyway...
Counter question: why are unsigned literals used in the library? :)
Do you prefer them all to be unsigned?

lib/stm32/l4/rcc.c
stefanct: Interesting, RM0394 does not mention a 5th one (that's for STM32L41xxx/42xxx/43xxx/44xxx/45xxx/46xxx). How should this be handled?

lib/stm32/l4/rcc.c
karlp: please do remember the target here.  this is a cortex-m library, there is _no concept_ of int being 16bit here, and we're not going "just be safe" for that sort of thing :)

They're unsigned in many places because we are often shifting things, and shifting ints has some UB that was triggering warnings.

lib/stm32/l4/rcc.c
karlp: only if there _is_ a clock setter.  At present, for L4, there _is not_ so this is _not a bug_ (yet)  L4 doesn't (yet) have a routine like: https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/f4/rcc.c#L789 that takes one of the structs.
_those_ routines update those global vars.

lib/stm32/l4/rcc.c
karlp: what do you mean? we support the superset, it's user error to attempt to use a peripheral that doesn't exist.  We have zero interest in having two libraries for different L4s.  (that might be "optimal" but it's unmaintainable in the current state)

lib/stm32/l4/rcc.c
karlp: (so just drop the comment as requested, it's irrelevant.  You can file a new PR that adds the clock helpers.... but we don't need a bug saying there's no clock helpers...)

lib/stm32/l4/rcc.c
stefanct: OK, so implementing the superset and leaving the responsibility to the user is your choice to handle it. That's all I wanted to know... no strong feelings about this - getting the superset correct and complete is hard enough :)

lib/stm32/l4/rcc.c
stefanct: I do remember, that's why I said it's not really relevant. :) There won't be a lot of shifting when returning a literal though, so I am a bit puzzled why the suffix is used so often there. I'll add it to make it homogeneous.

lib/stm32/l4/rcc.c
stefanct: Oh, I see, so the PLL setters are not supposed to fix these variables at all - only the setup functions. That's quite surprising and had cost me some time to figure out why my UART output is broken ;)
It would be great if there would be some place to document all of this stuff (for devs and users, respectively) - is there?

lib/stm32/l4/rcc.c
karlp: Yes, we've been doing superset only since the beginning, even F1 has massive differences in availability.  At least, in 99% of cases, ST makes it a proper superset, not a conflicting set :)

lib/stm32/l4/rcc.c
karlp: Personally, I don't add them unless I get warnings, all I was commenting on here was the inconsistency :)

lib/stm32/l4/rcc.c
karlp: well, for ~everyone else, all the methods come in at the same time, so there's never any users using the raw helpers.  I have no idea why L4 ended up half way.  Should _allllll_ the helpers update those variables? perhaps, but that's that not what is done in the rest of the library right now at least.

stefanct:
Have you looked at the device.data patch too or should I separate that into its own PR to get the other changes in?

karlp:
merged thanks

PR #1395, opened 2022-03-23
Avoid warning message about cast pointer from smaller integer type

zvova7890:
This is avoid warning messages about cast pointers to integer, including analyzers.

Screenshot_20220323_124246

### DIFF
--- a/include/libopencm3/cm3/common.h
+++ b/include/libopencm3/cm3/common.h
@@ -64,10 +64,10 @@
 #include <stdbool.h>
 
 /* Generic memory-mapped I/O accessor functions */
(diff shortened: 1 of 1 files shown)


zvova7890:
This is kdevelop semantic analyzer, as I'm understand based on clang. Maybe kdevelop/clang suggest that is can build as 64bit code, so produce this warn. Any way, I think it can be useful for analyzers.

uintptr sounds reasonable. I have changed

karlp:
hang on, you were using an analyser without setting the target?  if you get no warnings when you set the target properly, I think this should be just closed as "not an issue" ?

PR #1397, opened 2022-03-31
NRF5x: I2C EasyDMA support for NRF5x, extended API

ventZl:
This commit adds support for NRF52 TWI Master mode and slightly extends
existing I2C API. This is a breaking change, while mode selection needs to be
done during enabling I2C. There is one additional breaking change done because:
1) Unicore MX API design was PITA for writes
2) It is incompatible with EasyDMA

I strongly apologize to all two users who might be affected by this change.

### DIFF
--- a/include/libopencm3/nrf/common/i2c.h
+++ b/include/libopencm3/nrf/common/i2c.h
@@ -75,9 +75,16 @@
 /* Register Contents */
 
 /** @addtogroup i2c_shorts I2C event -> task shortcuts
+ * Shortcut, which is activated causes that if certain event
+ * occurrs, chosen task is automatically acitvated by the
+ * hardware.
  * @{
  */
+
+/** On byte boundary, activate suspend task. */
 #define I2C_SHORTS_BB_SUSPEND      (1 << 0)
+
+/** On byte boundary, activate stop task. */
 #define I2C_SHORTS_BB_STOP         (1 << 1)
 
 /**@}*/
@@ -100,12 +107,34 @@
 /** @addtogroup i2c_mode I2C peripheral mode 
  * @{
  */
+/** NRF51 legacy mode. 
+ * On NRF51, this is the only mode available.
+ * On NRF52, this mode does not support EasyDMA.
+ */
 #define I2C_MODE_LEGACY           (5)
 /**@}*/
 
+/** @addtogroup i2c_freq_const I2C frequency constants
+ * @{
+ */
+
+/** 100kHz */
 #define I2C_FREQUENCY_100K         (0x01980000)
+/** 250kHz */
 #define I2C_FREQUENCY_250K         (0x04000000)
+/** 390kHz
+ * @note: This value is not documented in datasheet. It provides
+ * ~390kHz clock with correct timing.
+ */
+#define I2C_FREQUENCY_390K         (0x06200000)
+/** 400kHz
+ * @note: According to datasheet, there is HW bug which prevents
+ * MCU from generating correct timings, therefore it might be
+ * unusable. Use @ref I2C_FREQUENCY_390K instead, if this affects
+ * you.
+ */
 #define I2C_FREQUENCY_400K         (0x06680000)
+/**@}*/
 
 #define I2C_PSEL_OFF               (0xffffffff)
 
--- a/include/libopencm3/nrf/common/i2c.h
+++ b/include/libopencm3/nrf/common/i2c.h
@@ -75,9 +75,9 @@
 /* Register Contents */
 
 /** @addtogroup i2c_shorts I2C event -> task shortcuts
- * Shortcut, which is activated causes that if certain event
- * occurrs, chosen task is automatically acitvated by the
- * hardware.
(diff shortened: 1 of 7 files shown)


lib/nrf/common/i2c_common.c
```diff
@@ -52,14 +55,15 @@ void i2c_disable(uint32_t i2c)
...
  *
  * @param[in] i2c uint32_t i2c peripheral base.
- * @param[in] data uint8_t the first byte to send.
  */
-void i2c_start_tx(uint32_t i2c, uint8_t data)
+void i2c_start_tx(uint32_t i2c)
 {
-	PERIPH_TRIGGER_TASK(I2C_TASK_STARTTX(i2c));
-	I2C_TXD(i2c) = data;
+	I2C_TASK_STARTTX(i2c) = 1;
```
karlp, line 66: what was the "PERIPH_TRIGER_TASK() " wrapper doing?

karlp: never mind, I see it was a stupid dumb macro that should have been burnt in fire.

lib/nrf/common/i2c_common.c
```diff
@@ -180,5 +186,14 @@ void i2c_resume(uint32_t i2c)
 	PERIPH_TRIGGER_TASK(I2C_TASK_RESUME(i2c));
 }
 
-
+/** Configure event -> task shortcuts
+ *
+ * Sets new shortcut configuration for I2C peripheral.
+ * @param[in] i2c i2c peripheral base.
+ * @param[in] shorts shortcut activation configuration
```
karlp, line 193: does this have a @ref it could refer to?

ventZl: Sure, done

include/libopencm3/nrf/common/i2c.h
```diff
@@ -75,9 +75,16 @@
 /* Register Contents */
 
 /** @addtogroup i2c_shorts I2C event -> task shortcuts
+ * Shortcut, which is activated causes that if certain event
+ * occurrs, chosen task is automatically acitvated by the
+ * hardware.
```
karlp, line 80: this has typos and reads strangely?  does this make sense automatically to familiar NRF users?

include/libopencm3/nrf/common/i2c.h
```diff
@@ -97,21 +104,47 @@
...
 #define I2C_FREQUENCY_250K         (0x04000000)
+/** 390kHz
+ * @note: This value is not documented in datasheet. It provides
+ * ~390kHz clock with correct timing.
+ */
+#define I2C_FREQUENCY_390K         (0x06200000)
+/** 400kHz
+ * @note: According to datasheet, there is HW bug which prevents
+ * MCU from generating correct timings, therefore it might be
+ * unusable. Use @ref I2C_FREQUENCY_390K instead, if this affects
```
karlp, line 133: lol :)

karlp: Related to this, a) does this apply to nrf51 as well as nrf52? (it's in the common defines here) and b) should "set fast mode" just do a "set 390" behind the scenes on the affected targets?

include/libopencm3/nrf/common/i2c.h
ventZl: I should not write documentation before I drink morning coffee

include/libopencm3/nrf/common/i2c.h
ventZl: According to this (https://infocenter.nordicsemi.com/topic/errata_nRF52832_Rev3/ERR/nRF52832/Rev3/latest/anomaly_832_219.html), this bug is only present on NRF52 *and* only if I2C_MODE_MASTER is used. While the difference in timing is some 50ns , certain devices even might work, thus automatic setting of 390kHz instead of 400kHz is probably not what we want to do here.

The whole device is riddled with so many bugs, that I am not even sure what to do. Some of them can have workarounds implemented at level LOCM3 represents, but some workarounds are too heavy and complex to be implemented in LOCM3.

karlp: sounds reasonable, leave it as you have it then, it all looks good other than that oddly worded doc just above :)

include/libopencm3/nrf/common/i2c.h
```diff
@@ -75,9 +75,9 @@
 /* Register Contents */
 
 /** @addtogroup i2c_shorts I2C event -> task shortcuts
- * Shortcut, which is activated causes that if certain event
- * occurrs, chosen task is automatically acitvated by the
- * hardware.
+ * The effect of activated shortcut is, that upon I2C event
```
ventZl, line 78: The term "shortcut" itself is used in reference manual, so this wording should be familiar to the NRF users.

karlp: could we just use "shortcut" then instead of "shorts" ?

karlp:
merged via rebase in 66bf499e thanks!

PR #1399, opened 2022-04-01
doc: cm3: mpu: document defines of individual register fields

stefanct:
I have used the doxygen's post-syntax (`/**<`) as it clutters the code less IMHO.


karlp:
merged in 757a0a14 thanks!

PR #816, opened 2017-08-21
usbfs_v1: double buffering for isochronous endpoints

WGH-:
As specified in the manual (see RM0008, "23.4.4 Isochronous transfers"), isochronous endpoints on usbfs_v1 hardware are always implicitly in double-buffered mode.

Current libopencm3 code doesn't support this, so isochronous endpoints don't work at all.

This is a work-in-progress pull request. I've implemented double-buffered for IN endpoints only. I've tested it on my STM32F103 board, and it works smooth in simple test-cases.

Before implementing the same for OUT endpoints, I wish to collect some feedback. Am I even going in the right direction?

Somewhat related to #679, though that is issue is about usbfs_v2, and probably about _optional_ double buffered mode for bulk transfers.

PR #1401, opened 2022-04-06
st_usbfs: Double buffering for isochronous endpoints

tormodvolden:
Based on #816


lib/stm32/common/st_usbfs_core.c
```diff
@@ -37,44 +37,56 @@ void st_usbfs_set_address(usbd_device *dev, uint8_t addr)
...
+		*real_size = (encoded + 1) << 5;
 		/* Set BL_SIZE bit (no macro for this) */
-		size |= (1<<5);
+		encoded |= (1<<5);
 	} else {
 		/* round up and div by 2 */
-		size = (size + 1) >> 1;
-		realsize = size << 1;
+		encoded = (size + 1) >> 1;
+		*real_size = encoded << 1;
```
karlp, line 69: please drop all the whitespace changes.  All you're doing here is adding the "output" parameter right?  Given that the "output" is just the existing return value <<1 I'm kinda curious what the value is at all?

lib/stm32/common/st_usbfs_core.c
```diff
@@ -37,44 +37,56 @@ void st_usbfs_set_address(usbd_device *dev, uint8_t addr)
...
+uint16_t st_usbfs_set_ep_rx_bufsize(usbd_device *dev, uint8_t ep, uint32_t size)
+{
+	uint16_t real_size;
+
+	(void)dev;
 	/* write to the BL_SIZE and NUM_BLOCK fields */
-	USB_SET_EP_RX_COUNT(ep, size << 10);
-	return realsize;
+	USB_SET_EP_RX_COUNT(ep, convert_bufsize(size, &real_size) << 10);
+	return real_size;
```
karlp, line 89: What's actually even chantged here then?  Please use the commit message to explain why this has to change as well, it appears to be unrelated to "add double buffering for isochronous" and the whitespace changes above make it very hard to actually follow

lib/stm32/common/st_usbfs_core.c
```diff
@@ -97,27 +109,51 @@ void st_usbfs_ep_setup(usbd_device *dev, uint8_t addr, uint8_t type,
...
+		dev->pm_top += max_size;
 		if (callback) {
 			dev->user_callback_ctr[addr][USB_TRANSACTION_IN] =
 			    (void *)callback;
 		}
 		USB_CLR_EP_TX_DTOG(addr);
-		USB_SET_EP_TX_STAT(addr, USB_EP_TX_STAT_NAK);
-		dev->pm_top += max_size;
+		if (type == USB_ENDPOINT_ATTR_ISOCHRONOUS)
+			USB_SET_EP_TX_STAT(addr, USB_EP_TX_STAT_VALID); // or DISABLED?
```
karlp, line 132: what's the "or disabled" for? still unsure? please be sure? :)

karlp: Further, always, always, always use properly braced {} statements.

lib/stm32/common/st_usbfs_core.c
```diff
@@ -97,27 +109,51 @@ void st_usbfs_ep_setup(usbd_device *dev, uint8_t addr, uint8_t type,
...
-		USB_SET_EP_RX_ADDR(addr, dev->pm_top);
-		realsize = st_usbfs_set_ep_rx_bufsize(dev, addr, max_size);
+		uint16_t real_size;
+		if (type == USB_ENDPOINT_ATTR_ISOCHRONOUS) {
+			USB_SET_EP_TX_STAT(addr, USB_EP_TX_STAT_DISABLED);
+			USB_SET_EP_RX_0_ADDR(addr, dev->pm_top);
+			USB_SET_EP_RX_0_COUNT(addr, convert_bufsize(max_size, &real_size) << 10);
+			dev->pm_top += real_size;
+			USB_SET_EP_RX_1_ADDR(addr, dev->pm_top);
+			USB_SET_EP_RX_1_COUNT(addr, convert_bufsize(max_size, &real_size) << 10);
```
karlp, line 145: I find the idea that you use this _COUNT() macro in _completely different_ manners depending on the direction to be extremely confusing.  I realise this is "behind the scenes" code that isn't user facing, but why drop the _set_ep_rx_bufsize() usage?  if you've dropped it, is it still useful at all? if it's no longer useful, why still keep it?

lib/stm32/common/st_usbfs_core.c
```diff
@@ -200,15 +236,33 @@ void st_usbfs_ep_nak_set(usbd_device *dev, uint8_t addr, uint8_t nak)
...
 	(void)dev;
 	addr &= 0x7F;
 
-	if ((*USB_EP_REG(addr) & USB_EP_TX_STAT) == USB_EP_TX_STAT_VALID) {
-		return 0;
+	if ((GET_REG(USB_EP_REG(addr)) & USB_EP_TYPE) == USB_EP_TYPE_ISO) {
+		/* Isochronous endpoints always use double buffering.
+		 * Additionally, TX_STAT check doesn't make sense for them. */
+		if (GET_REG(USB_EP_REG(addr)) & USB_EP_TX_DTOG) {
+			// buffer 1 is in use by the peripheral
```
karlp, line 248: only C style comments.  Run "make stylecheck"

lib/stm32/common/st_usbfs_core.c
tormodvolden: Hi, thanks for reviewing. What whitespace changes? I changed the variable name "size" to "encoded" for readability (and not modify the input variable).

The "output" is just the encoded (returned) << 1 in the "else" clause but not in the "then" clause. The "output" may differ from the input "size" since it can be rounded up to next even number.

lib/stm32/common/st_usbfs_core.c
tormodvolden: This part was part of _set_ep_rx_bufsize but the calculation was factored out into convert_bufsize(). Maybe the diff output makes this unclear. If you look at the new code it should be simple.

tormodvolden: But yes, we can mention in the commit message that the RX buffer size was factored out. Maybe there's a better name too. encode_rx_bufsize() ?

lib/stm32/common/st_usbfs_core.c
tormodvolden: Oops forgot to remove that. It works like it is, with my testing.

lib/stm32/common/st_usbfs_core.c
tormodvolden: It is the register layout that makes this confusing. The RX count is encoded and sits at bits 10 and up, whereas the TX count is plain, written at bits 0 and up. The question is if the bit shifting <<10 should be part of the _SET_EP_RX macros.

tormodvolden: But then if we "hide" the shifting should we hide the encoding too?

lib/stm32/common/st_usbfs_core.c
tormodvolden: Sure. Remnants of the original PR :)

tormodvolden:
Yes, regression tests would be nice. My whole motivation for this is to make test devices that we can use for regression testing in libusb (and pyusb). I will look at the gadget-zero stuff.

Would you prefer fixup commits or repushing?

lib/stm32/common/st_usbfs_core.c
tormodvolden: > why still keep it

I thought it was exposed for applications, since it wasn't "static". Would be good to get rid of otherwise. It would make this function more difficult to read if we would use _set_ep_rx_bufsize in one case and the direct code in the other case. The symmetry makes it easier to read.

lib/stm32/common/st_usbfs_core.c
tormodvolden: If you like I can do the table layout in a separate commit.

denis-shienkov:
Guys, any progress in?

tormodvolden:
I'll probably not get to this for a while. My plan is to rebase the commits based on feedback, and I am also awaiting answers on my questions above. The major remaining work is to construct the tests, if you or anyone can help on that it would speed up things.

karlp:
fixup or squishes make no difference to me right now, we can handle that when we're happy with it.  I'm not sure what else this is waiting for me on, it _must_ have tests before I'll merge it, as I've tried saying, and will say again and again and again I believe :)

denis-shienkov:
To be honest, I have not ideas how to test it at all.

F.e. I faced with the interesting situation (using the DWC OTG FS && STM32F4x) when the `libopencm3` does not support the isochronous stream, but the stream is transferred somehow.

I have created the FW which implements the fake USB microphone which produces a sine wave.  

When I record the audio stream using any external software (e.g. Audacity or any other), then I got a correct data (a pure sine wave). 

But, when I try to listen the microphone stream directly (e.g. to forward the stream to the speaker, that is possible in Windows as "out in box"), then I hear the sound distortion (pulsation).

So, I meant, that it is impossible to detect that a stream is corrupted using auto-tests, because it receives correctly (hello Audacty). ;)

karlp:
You should absoltuely be able to use something akin to the gadget zero source/sink endpoints to have it programmatically generate expected sequences, and use pyusb to programmatically read them.    If you can hear distortion, you are most definitely not getting the expected data!  

I _cannot_ do regression tests on things that require listening for distortion in audio capture, that's simply untenable.  I completely believe that this fixes things for you, no doubt, my concern is next year....

PR #1402, opened 2022-04-07
Linker script: bug fix + additional symbols in all regions

stefanct:
Some enhancements for the linker script template


karlp:
sure, merged thanks.  kinda wish we had xxxx_start and xxx_end, but this is at least consistent with the gross historical mess :)

PR #1405, opened 2022-04-23
Libopencm3: Provide support for CMake projects

ventZl:
This adds CMake module, which essentially duplicates functionality of
genlink-*.mk modules for bare make. If used via find_package(), then it
attempts to identify MCU, generate linker script, fill in some variables
as per find_package() command interface. It also defines imported target
one can link against.

### DIFF
--- a/cmake/FindLibopencm3.cmake
+++ b/cmake/FindLibopencm3.cmake
@@ -1,0 +1,177 @@
+include(FindPackageHandleStandardArgs)
+
+#[=====================================[.rst
+FindLibopencm3
+--------------
+
+Finds the Libopencm3 library suitable for target device specified in variable
+DEVICE. You have to set this variable to full name of your MCU before calling
+find_package(Libopencm3) otherwise search will fail.
+
+Imported targets
+^^^^^^^^^^^^^^^^
+
+``Libopencm3::Libopencm3``
+  Libopencm3 HAL suitable for your MCU. This library automatically configures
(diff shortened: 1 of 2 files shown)


cmake/FindLibopencm3.cmake
```diff
@@ -0,0 +1,177 @@
...
+		if (NOT "${CPP_RESULT}" EQUAL "0")
+			message(FATAL_ERROR "Unable to generate linker script for device ${DEVICE}")
+		endif()
+
+		# Provide exported variables
+		set(Libopencm3_LIBRARY ${LOCM3_CANDIDATE})
+		set(Libopencm3_LIBRARY_DIRS ${Libopencm3_ROOT_DIR}/lib)
+		set(Libopencm3_INCLUDE_DIRS ${Libopencm3_ROOT_DIR}/include)
+		set(Libopencm3_DEFINITIONS ${Libopencm3_DEFINITIONS} ${ARCH_FLAGS})
+		set(Libopencm3_LINK_OPTIONS -static -nostartfiles ${ARCH_FLAGS} -T${Libopencm3_LINKER_SCRIPT})
```
ventZl, line 144: Here I deliberately add -static and -nostartfiles. Genlink-config.mk is not doing this, but I am not quite sure if there are any valid use cases, where you might not want these two switches. Having them here will prevent from nasty ugly messages from compiler, if used without additional flags in CMakeLists.txt.

PR #1408, opened 2022-04-27
stm32: Add an API to set external clock mode 2

slorquet:
Hello, here is my PR for external clock mode 2. I can update it to match your requirements. thank you.


karlp:
merged in d8aa2f17 thanks!

PR #1410, opened 2022-05-09
add I2C for STM32G4 

qosch:
The STM32G4 series has the exact same I2C peripheral as STM32G0 (that is already supported). This PR makes it possible to use I2C on the STM32G4 series as well.
The history has lots of unnecessary commits in it but only three files were added/changed. If you want this cleaned up before merge just let me know.


karlp:
I absolutely need this cleaned up before it can be merged, yes. :)

lib/stm32/g4/Makefile
```diff
@@ -49,6 +49,7 @@ OBJS += rcc.o rcc_common_all.o
 OBJS += spi_common_all.o spi_common_v2.o
 OBJS += timer_common_all.o timer_common_f0234.o
 OBJS += quadspi_common_v1.o
+OBJS += i2c_common_v2.o
```
karlp, line 52: this is a sorted list.

include/libopencm3/stm32/g4/i2c.h
```diff
@@ -0,0 +1,34 @@
+/** @defgroup i2c_defines I2C Defines
+ *
+ * @ingroup STM32G4xx_defines
+ *
+ * @brief <b>Defined Constants and Types for the STM32G0xx I2C</b>
```
karlp, line 5: this is wrong.

qosch:
I'll create a new pull request to get a clean history

Issue #1406, opened 2022-04-26
Atmel SAMD21 support


doxygen-spammer:
Hi! I have some evaluation boards with SAMD21 and SAMR21, which I would like to program with a free framework (i. e. libopencm3).

I understand that support for SAM is still incomplete, and for SAMD21 in particular there is only PORT.

There is currently a directory for SAMD, containing the interrupt table for SAMD09/10/11, and the library for PORT.
The interrupt tables of SAMD20 and SAMD21 are incompatible, each, so they would need either `#if` stuff or separate files.

I would like to implement more support for SAMD21.
This is what I would like to do first:

* Interrupt vector table (need to figure out where irq.json is processed)
* Library for PM (Power Manager, turns on/off peripherals like TC, SERCOM), can be ported from SAM4L, probably by making most code “common”.
* Library for EIC (interrupts from pins)
* Library for TC (timer), some register definitions from SAM3X available, do not apply here.

If I have still fun after that, I plan these:

* SERCOM: some library for USART and some register definitions for UART available, but they do not apply here. Apparently they were relevant for SAM3X once.
* ADC: some library for ADCIFE available for SAM4L, will figure out if that is the same.

Please let me know what you think about this.

I have read HACKING_COMMON_DOC.
Are there other guidelines or quality standards for adding support for new families, which I should read?

karlp:
knock yourself out!

Use separate trees where required, rather than ifdefs generally, especially when you're already talking about different families.  Just use common dirs where possible to share peripheral support code, as they are normally shared across families within the same vendor.  only thing that people sometimes miss is running "make doc" and "make stylecheck" to check that the doxygen is building nicely.

doxygen-spammer:
I decided to make a new family in the directory sam/d21.
SAMD is a similar family like STM32F, where e. g. SAMD10 and SAMD21 or STM32F0 
and STM32F1 are to wide extents compatible, but differ in details like memory 
addresses.
So I adapt the logic from STM32 to SAM.

The directory sam/d can stay for now, because D09, D10, and D11 are equivalent 
as far as already covered (interrupt vectors and PORT).

---

But I am not sure about the doxygen groups, which would be affected by a new 
family.

If I go to doc/samd/html/modules.html, I can see the groups “Peripheral APIs” 
and “SAMD Defines”, which are defined in doc-samd.h as `peripheral_apis` and 
`SAMD_defines`.
Currently there is the library for PORT, which appears as “IO Port API” and 
[...]

PR #1411, opened 2022-05-10
WIP: SAM3A: Include GPIO in doxygen

doxygen-spammer:
This is how I am going to enable doxygen modules for SAM family libraries.

I mentioned this effort in #1406, and am opening this pull request already now so you can check it already now, if you want. ;)


karlp:
:+1:

karlp:
This is still marked as a draft, but this looks fine for merge, is there any reason I should not merge this for you?

include/libopencm3/sam/common/pmc_common_all.h
```diff
@@ -27,160 +34,223 @@
 
 /* --- Power Management Controller (PMC) registers ----------------------- */
 
-/* System Clock Enable Register */
+/** System Clock Enable Register */
 #define PMC_SCER			MMIO32(PMC_BASE + 0x0000)
```
doxygen-spammer, line 38: Things like these are not problematic.

include/libopencm3/sam/common/pmc_common_all.h
```diff
@@ -27,160 +34,223 @@
...
 /* --- Register contents --------------------------------------------------- */
 
+/**
+ * @addtogroup pmc_scer PMC System Clock Enable Register (PMC_SCER)
+ * @{
+ */
 
-/* --- PMC System Clock Enable Register (PMC_SCER) ------------------------- */
-
-/* Programmable Clock Output Enable */
```
doxygen-spammer, line 93: Here I decided to make groups for contents of registers.
A similar method is used here (but not necessarily grouped by register, I don’t know): https://github.com/libopencm3/libopencm3/blob/467522778329d6f41781a6c951b77d6ff6744de6/include/libopencm3/nrf/common/i2c.h#L75-L82

So far, this method of grouping makes sense here.
Register contents for the same register may be defined in multiple `pmc_*_common.h` files, and with this grouping they end up in one place.
(Without this method of grouping, they would also end up in one place, in the main page of the PMC module (which is a large place).)

include/libopencm3/sam/common/pmc_common_all.h
```diff
@@ -27,160 +34,223 @@
...
+/**
+ * Programmable Clock Output Enable
+ * @{
+ */
 #define PMC_SCER_PCK0			(0x01 << 8)
 #define PMC_SCER_PCK1			(0x01 << 9)
 #define PMC_SCER_PCK2			(0x01 << 10)
+/** @} */
```
doxygen-spammer, line 109: This makes problems right now.
The method with `@{` and `@}` works in general, but not very nice for defines.
With this code, there is no grouping, only the first define gets the comment.
Its not very bad, the other two defines appear alphabetically immediately below, so one doesn’t need to search far away.

I am not sure what to do exactly here.
My referenced example (nrf/common/i2c.h) does not have any comments on this low level, the defines just appear in alphabetical order on the documentation page.
Do you think I should just be happy with the alphabetical order?

doxygen-spammer (commented):
See, I have continued this patch. :D

I am facing a minor problem now.
The module PMC has more “detailed” documentation than GPIO, because register contents are not just 32 pins, but heterogenous stuff, so there are “Register Contents” sections.
I tried to use the approach from nrf/common/i2c.h, but it doesn’t work out exactly the same.

I would be happy if you could give me some advice here. :)

doxygen-spammer:
> This is still marked as a draft, but this looks fine for merge, is there any reason I should not merge this for you?

Hmm, I replied to this via email, but to the wrong recipient.
You are probably confused now.
I wrote:

> Well. You can merge it now, I’d say.
> 
> I didn’t forget it.
> Actually I am planning to continue this, although I am currently not very 
> fast. ;)

You are probably still confused, because it says “SAM3A: Include GPIO in doxygen”, i. e. not including PIO and PMC.
I didn’t write the pull request title with much thought.

My idea was initially to enable doxygen for SAMxx at once, but now I limit to SAM3A.
I want to include SMC, and then it should be done for SAM3A.

include/libopencm3/sam/common/pmc_common_all.h
karlp: you need to eithter @defgroup or @addtogroup, if you want groups.  the @{ and @} won't work well by themselves.  the warning logs are normally useful.  Look at places like stm32, rather than nrf, it's had a lot more work on it over the years.

At this level, we would normally have something like : https://github.com/libopencm3/libopencm3/blob/master/include/libopencm3/stm32/common/adc_common_v2.h#L66-L83  with a group for all the values of a register, and that register in a group for the peripheral.  But you can also just go in steps, it doesn't have to do everything in one go.

doxygen-spammer: > with a group for all the values of a register, and that register in a group for the peripheral.

Yeah, that example looks good. I want to try that today. :)

doxygen-spammer: Done.

It turns out that NRF and STM32 do not differ much in this point, and I was pretty close to this:

> with a group for all the values of a register, and that register in a group for the peripheral.

It is just a bit more verbose now:
the precise answer for the following is to wrap related defines in @defgroup.

> The method with `@{` and `@}` works in general, but not very nice for defines.

---

I changed several lines to @defgroup, where I used @addtogroup before.
NRF seems to have been a bit less strict, but the decision for @defgroup and @addtogroup does not turn out to be difficult.
It just follows the place where the register address is defined: there the register is introduced, therefore the register contents group is defined with @defgroup.

include/libopencm3/sam/common/pmc_common_all.h
doxygen-spammer: (Answered as of https://github.com/libopencm3/libopencm3/pull/1411#discussion_r917442391)

include/libopencm3/sam/common/pmc_common_all.h
doxygen-spammer: (Answered as of https://github.com/libopencm3/libopencm3/pull/1411#discussion_r917442391)

doxygen-spammer:
I am happy with the current state/result of the patch, so the pull request is no longer a draft.
I also wrote a suggested squash commit message in the pull request description. :)

PR #1250, opened 2020-09-11
Stm32g4 common

Igorbunow:
Add stm32g4 peripheral support: EXTI, RTC, UART, SPI, I2C, RCC, SYSCFG + PWM, FLASH, DMAMUX, ADC, OPAMP, USB, CRC, DAC

PR #1416, opened 2022-05-27
add I2C for STM32G4

qosch:
The STM32G4 series has the exact same I2C peripheral as STM32G0 (that is already supported). This PR makes it possible to use I2C on the STM32G4 series as well.
(used to be #1410)


karlp:
overlaps with https://github.com/libopencm3/libopencm3/pull/1250 but this is simpler and easier to pull in

omarandlorraine:
Splendid, this is exactly what I need. What needs to happen for it to get merged in?

karlp:
merged in 1e71f18b thanks!

PR #1419, opened 2022-06-01
add package.json so this repository can directly be used in PlatformIO

qosch:
with this file present this repository can be used in PlatformIO using the command

`platform_packages = framework-libopencm3@https://github.com/libopencm3/libopencm3.git`

in the platformio.ini file of any project. 
This is helpful for anyone using this platform because the default package of libopencm3 that is provided by PlatformIO is 2 years old.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "framework-libopencm3",
+  "version": "1.10000.220601",
+  "description": "Open source ARM Cortex-M microcontroller library",
+  "keywords": [
+    "framework",
+    "arm"
+  ],
+  "homepage": "http://www.libopencm3.org/",
+  "license": "GPL-3.0-or-later",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/libopencm3/libopencm3"
+  }
+}


package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "framework-libopencm3",
+  "version": "1.10000.220601",
```
karlp, line 3: this sort of thing without having any updates is always going to be misleading I'm afraid.  I don't think there's any easiliy obvious way of doing this, as pointing to master like this is always a rolling target?

qosch: You mean the fixed version string? I don't know a nice solution to that but I am not sure there are consequences if it is not updated. 
@maxgerhardt maybe you have an idea?

maxgerhardt: I wouldn't add a datecode to the version (`220601` -> `0`), and it also doesn't seem like version 1.0.0 is official out in this repo looks at tags [and the README](https://github.com/libopencm3/libopencm3#status-and-api). Soo I guess the current version status is something in between 0.8 and 1.0?

Other frameworks like [Arduino-ESP32](https://github.com/espressif/arduino-esp32/blob/master/package.json) use the next projected release version, [Arduino-Pico](https://github.com/earlephilhower/arduino-pico/blob/master/package.json) uses the latest released version, [Arduino-ESP8266](https://github.com/esp8266/Arduino/blob/master/package.json) uses a `-dev` suffix. Arduino-Pico also auto-generates the `package.json` through [a script](https://github.com/earlephilhower/arduino-pico/blob/master/tools/makever.py), so maybe that can be automated as a git hook. But I don't seem to be able to find any file here declaring the framework version :/

So, I'd recommend something like `0.8.0` or `0.9.0-dev` and upon release just increase it to `1.0.0`(`-dev`). You can also leave the version as unspecific `0.0.0`. If `platform_packages` is used to refer to the package, PlatformIO will add the short commit hash like `+sha.7d971fa` anways, so it's look like `0.8.0+sha7d971fa` (if the base version is 0.8.0).

maxgerhardt: Maybe @ivankravets has some thoughts?

ivankravets: We recommend using https://semver.org/.
I see this repository already has a 0.8.0 tag. The version field in development branch can use build prefixes. For example, `"version": "0.8.1-beta.1"`

PR #1420, opened 2022-06-02
add RNG support for STM32G4 and correct typo in memorymap

qosch:
The STM32G4 series has the same random number generator as the STM32G0 that is already supported in libopencm3, so adding support only requires small changes. During testing on hardware I noticed that the peripheral is not responding. After a bit of debugging I found the issue: there are typos in the memorymap for this device (AES_BASE and RNG_BASE are both missing a zero). Correcting it I could get the peripheral working.


karlp:
merged in 46752277 thanks!

PR #1422, opened 2022-06-04
usart: use rcc helpers for peripheral clock frequency

via:
These rcc clock helpers seem to be implemented for the whole family, and H7 specifically has only its static clock structure and not the rcc_apb*_frequency variables, so this seemed like a sane switch to allow the H7 USART to work.

Tested with USART1 on an stm32h743.


ALTracer:
Likely resolved by 4d7a2ca (after e643f66), as note in #1456.

PR #1424, opened 2022-06-17
GD32F1x0: added USART, SPI, DMA and TIM support

elephantenlein:


lib/gd32/f1x0/Makefile
```diff
@@ -33,10 +33,14 @@ TGT_CFLAGS      += $(DEBUG_FLAGS)
...
+OBJS += dma_common_f1.o
 OBJS += flash.o flash_common_all.o flash_common_f.o flash_common_f01.o
 OBJS += gpio_common_all.o gpio_common_f0234.o
 OBJS += rcc.o rcc_common_all.o
+OBJS += spi_common_all.o spi_common_v2.o
+OBJS += timer_common_all.o
+OBJS += usart_common_all.o usart_common_v2.o
 
-VPATH += ../:../../cm3:../common:../../stm32/common
+VPATH += ../:../../cm3:../common:../../gd32/common
```
karlp, line 44: absolutely not,  you're completely breaking the whole purpose here, which was to share code, not copy/paste it whole sale.

karlp:
Absolutely not in any form like this.  It's wildly unmaintainable to copypaste entire trees like this.  The existing gd32 code already was sharing code, and you've removed that and copypasted a whole new slew of code.    You should have tried to keep that.

lib/gd32/f1x0/Makefile
elephantenlein: I would argue.
1.  GD32 has different bias for peripherals numbering. TIM1 in STM32 is like TIM0 in GD32, let's stick to datasheets;
2. Same registers etc. between STM32 and GD32 have different names. Only counted ones are completely equal;
3. Some register bits differ between STM32 and GD32. Missing here, but present there;
These three justify copy-paste and modifications, IMHO.

PR #1425, opened 2022-07-04
C11 usability improvements

dragonmux:
Made use of `__asm__` in the scb code which automatically implies volatile and allows use with C11

This has been done in prep to allow the Black Magic Debug project to switch to C11 ahead of some changes that make use of C11 atomics.

As noted in the GCC manual page, `asm` is a vendor extension, which became an illegal word in C11 (all vendor extensions being required to be decorated with `__`). Technically the use of `volatile` with the standards conforming version of `asm` is extra verbiage for nothing, however after a conversation with @esden we agreed that keeping it in this change would be good for readers not familiar with the difference in semantics.

For further reading please refer to the [GCC page on Basic Asm](https://gcc.gnu.org/onlinedocs/gcc/Basic-Asm.html#Basic-Asm), the [C reference on identifiers](https://en.cppreference.com/w/c/language/identifier#Reserved_identifiers) and [GCC page on Alternate Keywords](https://gcc.gnu.org/onlinedocs/gcc/Alternate-Keywords.html#Alternate-Keywords) for more on extension naming.

It should be noted that `asm` can cause problems on non-GCC compilers which is a further reason to avoid using it, while `__asm__` is fairly consistently supported.

### DIFF
--- a/include/libopencm3/cm3/scb.h
+++ b/include/libopencm3/cm3/scb.h
@@ -549,7 +549,7 @@ struct scb_exception_stack_frame {
 
 #define SCB_GET_EXCEPTION_STACK_FRAME(f)				\
 	do {								\
-		asm volatile ("mov %[frameptr], sp"			\
+		__asm__ __volatile__("mov %[frameptr], sp"			\
 			      : [frameptr]"=r" (f));			\
 	} while (0)
 


karlp:
I'm more than happy to make the __asm__ change, but the volatile seems separate.  This is the only raw undecorated asm call left, and should clearly be a __asm__ but there's about 50/50 volatile and __volatile__ (with asm) and I'd like to see those cleaned up or reviewed together.  This statement is extended asm not basic asm, so the "volatile has no effect" statement doesn't matter _here_ and if the volatile vs __volatile__ is also causing you c11 problems, then the _other_ places will also cause problems, you just might not have hit them yet.  (other locations are in cortex.h)

In general I'd rather see this function dropped and replaced with a more standard simple function more like in cortex.h, but that's a different matter :)

dragonmux:
We had a conversation about it with Esden last night and agreed (in the same vein as this) that trying to eliminate the need for `__asm__` would be the more correct patch (we just went with the obvious one first).

To that end we will rework this PR to try to replace the use of `__asm__` here and do a review of the other instances of volatile.

karlp:
I can just fix asm ->  `__asm__` myself if that works for you?  we don't really need updated PRs on this unless you really want it like that.

dragonmux:
Sure, that works - was offering to do the review of `volatile` vs `__volatile__`, but yeah, that'd be just fine.

karlp:
thanks for reporting this!

PR #1431, opened 2022-08-13
STM32F4 clock tree fixes

beppone:
fix RCC_CFGR_HPRE* symbols
harmonize PLLM value (freq after division = 1MHz)


karlp:
This should be two separate commits.  I don't understand teh "fix" comment on the hpre?  if you're standardizing against a value defined multiple times, you might be wanting to deprecate the others, or remove.  this just... changes them... with no further information on why.  LIkewise, why do the pllm values need harmonization?

PR #1432, opened 2022-08-17
Feature: BOS and Microsoft OS specific descriptors support

dragonmux:
This PR implements support to locm3's USB stack for Binary Object Store (BOS) descriptors, and the Microsoft OS 2.0 Descriptor specification (which requires BOS support).

The motivation for doing this it to provide support in Black Magic Probe for auto-binding drivers to the Trace and DFU interfaces via descriptors in the firmware.

Additionally this PR implements tests for all the new descriptor requests added via the gadget0 framework and implements a gadget0 platform to allow the native BMP hardware to be used as a testing platform (this is the hardware we had available)


dragonmux:
We're aware the build is currently failing.. we'll fix the reason why in the morning. It's a question of platforms missed in the addition rather than a real bug.

petrutoader:
Is this of interest to the project? We currently need BOS descriptors and need to decide if we should create a fork or not.

karlp:
It is of interest to the project, and the submitter here has done a great job of adding the requested tests.  Regardless, testing on windows is not exactly something that fills me with joy.  It's pretty high on my list though.  As you can see of course, it's not exactly a minor "the header has the wrong value copied from the reference manual" so it's not something that happens overnight.

petrutoader:
Thanks for the clarification, totally understandable.

ALTracer:
Reporting that this is a working and useful feature. 
https://developer.chrome.com/articles/build-for-webusb/#microsoft-os-compatibility-descriptors
Both my Win10 and Win11 boxes recognized the WINUSB in WCID/BOS descriptors and attached WinUSB.sys drivers to Blackmagic Debug Probe-compatible board interfaces for DFU and TraceSWO (the other two interfaces are CDC-ACM and "just work" with usbser.sys already). This is after https://github.com/blackmagic-debug/blackmagic/pull/1509 got merged.
It should improve user experience even on old&unsupported WinXP & Win7 after a one-time .inf installation. https://github.com/pbatard/libwdi/wiki/WCID-Devices#wcid-advantages

esden (approved):
Considering that we are using this code for a while now in Black Magic Debug it is pretty well tested code. I think we can consider it tested and we can go ahead and merge it.

PR #1433, opened 2022-08-23
Feature: u(s)art configuration retrieval

dragonmux:
As part of addressing some defects in BMP, we found that the USB CDC support in locm3 didn't support the class specific request `GET_LINE_CODING`.

This PR addresses that by adding the missing define and then providing support for STM32 and LM4F (TM4C) for retrieving the current UART configuration to answer this USB request.


esden (approved):
LGTM, I have checked the datasheets for correctness and it all matches up. This is just adding extra get functions so it is a low risk PR.

PR #1436, opened 2022-09-08
[stm32] Fix DMAMUX RG channel reset

hepoun:
There was reset of DMAMUX request multiplexer channel instead of the DMAMUX request generator channel.


karlp:
merged in c82c7406 thanks!

PR #1438, opened 2022-09-16
stm32:dmamux: fix individual DMAMUX register aliases DMAMUXn_*

neoxic:
Argument `dmamux_base` is set by an alias and hence is always unused in the macros.


karlp:
merged thanks!

PR #1445, opened 2022-11-09
Arterytek's  AT32F4 support

beefdeadbeef:
Hi. I'm currently amusing myself with at32f403a based blackpill board --
you probably heard of those, stm32f10x frankenclone with m4f core, richer peripheral set,
flash shadowed by SRAM (no wait states!) and overally more of it.
Problem is, it's frankenclone in denial -- they renamed any and every register and bit name 
in their docs, although maintaining very close compatibility at register level.
So, i'm curious, if is this approach with hijacking stm32f1 code base is acceptable at all (thus RFC) ?
Currently i have only  basic things like rcc & flash,  gpio and timers, enough for blinky with systick and timer (at 240Mhz wow).


include/libopencm3/at32/f4/memorymap.h
```diff
@@ -0,0 +1,131 @@
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2009 Uwe Hermann <uwe@hermann-uwe.de>
```
karlp, line 4: if you want a more favourable beginning, don't do things like this...

include/libopencm3/stm32/f1/rcc.h
```diff
@@ -57,6 +58,10 @@
 #define RCC_CSR					MMIO32(RCC_BASE + 0x24)
 #define RCC_AHBRSTR				MMIO32(RCC_BASE + 0x28) /*(**)*/
 #define RCC_CFGR2				MMIO32(RCC_BASE + 0x2c) /*(**)*/
+#define RCC_CFGR3				MMIO32(RCC_BASE + 0x30) /*(^^)*/
+#define RCC_CFGR4				MMIO32(RCC_BASE + 0x50) /*(^^)*/
+#define RCC_CFGR5				MMIO32(RCC_BASE + 0x54) /*(^^)*/
+#define RCC_CFGR6				MMIO32(RCC_BASE + 0x5c) /*(^^)*/
```
karlp, line 64: yeah, so, I'm not going personally merge anything that adds f1 registers for a different target and a different family.  There's got to be a better way.

beefdeadbeef: ifdef'ing those counts for (slightly) better way ?

karlp: I'm not sure honestly, the two approaches people have tried to take so far have largely been "copy paste the entire f1 tree" which is a maintenance nightmare, and "ifdef f1 code to be a trashfire supporting everything in the world"   I imagine breaking it all into much smaller pieces so it can be composed more freely is likely the path forwards, but I've not tried (and it's not remotely on my own personal roadmap)

beefdeadbeef:
So, another attempt. This time i've decided to let stm32f1 code alone, let's see how this helps me.

lib/at32/f40x/gpio.c
```diff
@@ -0,0 +1,179 @@
...
+@brief <b>libopencm3 AT32F40x General Purpose I/O</b>
+
+@version 1.0.0
+
+@author @htmlonly &copy; @endhtmlonly 2009
+Uwe Hermann <uwe@hermann-uwe.de>
+@author @htmlonly &copy; @endhtmlonly 2012
+Ken Sarkies <ksarkies@internode.on.net>
+@author @htmlonly &copy; @endhtmlonly 2022
+Sergey Bolshakov <beefdeadbeef@gmail.com>
```
karlp, line 12: copyrights here don't match what you've got lower down, and isn't this all your own work? pretty sure there's nothing left of the 2009 code.

karlp:
I think this style looks a lot better personally!

lib/at32/f40x/gpio.c
beefdeadbeef: there's slightly modified (from f1 variant) gpio_set_mode(), unmodified gpio_set_eventout() and new gpio_set_mux().
git blame shows few lines exactly from 2009 by Uwe. But honestly all this copyright stuff isn't my idea of fun,
i'll gladly agree with what you see appropriate here.

lib/at32/f40x/rcc.c
```diff
@@ -0,0 +1,764 @@
...
+		RCC_BDCR |= (1 << 9);
+		break;
+	case RCC_HSE:
+		/* Turn the HSE on and wait while it stabilises. */
+		RCC_CR |= RCC_CR_HSEON;
+		while ((reg32 = (RCC_CR & RCC_CR_HSERDY)) == 0);
+
+		/* Choose HSE as the RTC clock source. */
+		RCC_BDCR &= ~((1 << 8) | (1 << 9));
+		RCC_BDCR |= (1 << 9) | (1 << 8);
```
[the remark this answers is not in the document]
beefdeadbeef: it survived that way from original submission in 2014, it seems. another wtf moment

beefdeadbeef:
> I think this style looks a lot better personally!

thanks. next in my todo list  are usb with clock recovery, maybe flash bits. 
oh, and could you please pick [db9f4ed] alone ? i2c code here depends on it.

karlp:
Ify ou want standalone commits reviewed, make them a standalone PR please.

include/libopencm3/at32/f40x/memorymap.h
```diff
@@ -93,13 +93,13 @@
 #define TIM9_BASE			(PERIPH_BASE_APB2 + 0x4c00)
 #define TIM10_BASE			(PERIPH_BASE_APB2 + 0x5000)
 #define TIM11_BASE			(PERIPH_BASE_APB2 + 0x5400)
-#define ACC_BASE            (PERIPH_BASE_APB2 + 0x5800)
-#define I2C3_BASE           (PERIPH_BASE_APB2 + 0x5c00)
-#define USART6_BASE         (PERIPH_BASE_APB2 + 0x6000)
-#define USART7_BASE         (PERIPH_BASE_APB2 + 0x6400)
-#define USART8_BASE         (PERIPH_BASE_APB2 + 0x6800)
-#define I2S2_EXT_BASE       (PERIPH_BASE_APB2 + 0x6c00)
-#define I2S3_EXT_BASE       (PERIPH_BASE_APB2 + 0x7000)
```
karlp: it would be _really_ nice not to have fixup commits in the same stack please.  It makes splitting and bisection on a long set like this awful.

include/libopencm3/at32/f40x/rcc.h
```diff
@@ -59,7 +59,7 @@
 #define RCC_CFGR3		MMIO32(RCC_BASE + 0x30)		/* CRM_MISC1 */
 #define RCC_CFGR4		MMIO32(RCC_BASE + 0x50)		/* CRM_MISC2 */
 #define RCC_CFGR5		MMIO32(RCC_BASE + 0x54)		/* CRM_MISC3 */
-#define RCC_INTMAP		MMIO32(RCC_BASE + 0x5c)		/* CRM_INTMAP */
+#define RCC_CFGR6		MMIO32(RCC_BASE + 0x5c)		/* CRM_INTMAP */
```
karlp, line 62: Is this defines to be compatible with stm32, but with comments (that no-one will ever see) showing what the actual vendor names are?  I'm inclined to say you want them both defined in that case honestly.

include/libopencm3/at32/f40x/memorymap.h
beefdeadbeef: i'm planning to squash this commit with first in series anyway -- if that's okay.

include/libopencm3/at32/f40x/rcc.h
beefdeadbeef: well, that's about two dosens registers, which do not exist in stm32f1 --  making them just as they are in at32 docs
feels sometimes really meh.  This particular diff just fixes difference in register naming with register value  define for this register below. 
And by the way comments like these _are_  wisible in doxygen, no ?

lib/at32/f40x/rcc.c
```diff
@@ -577,15 +577,15 @@ void rcc_set_hpre(uint32_t hpre)
...
 /*---------------------------------------------------------------------------*/
-/** @brief RCC HSI/6 divider bypass.
+/** @brief Set RCC HSI divider.
 
-@param[in] nodiv Bypass HSI/6 divider @ref rcc_cfgr3_hsi_nodiv
+@param[in] hsidiv HSI divider @ref rcc_cfgr3_hsidiv
 */
 
-void rcc_set_hsi_bypass(bool bypass)
+void rcc_set_hsi_div(uint32_t hsidiv)
```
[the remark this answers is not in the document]
beefdeadbeef: no,  that's was newly added. later i've decided that function naming doesn't follow existing pattern.
as i said, i'd like to squash this with initial commit later.

PR #1449, opened 2022-12-21
STM32L412 intial support

fk0815:
Add STM32L412 memory definition.


karlp:
You missed the l412xb vs l412x8 distinction, but thank you for filing this to draw attention to the missing parts.  I've fixed this via f58ffd4b thanks!

PR #1451, opened 2023-01-05
Allow building for stm32h7 targets

slorquet:
Hi,

STM32H7 build was broken because of the FPU flags and memory range addresses.

I fixed that and verified that a basic template project could be built.

Thanks for the integration.

If I need to fix anything please dont hesitate to ask, tbh I did these fix quite quickly.


karlp:
Hrm, interesting.  STM32L4 and h7 (after your fix) use RAM, RAM2 (first ram is ram1) and SAM3 and LPC use RAM and RAM1.  (First ram is ram zero)  Same for ROM.  The h7 code at the end was using the zero based style.  It's gross having them be different, but I think I like 1 based better, so let's go with it, for stm32s.

karlp:
merged via rebase in 84a59d92 thanks!

slorquet:
Thank you!

PR #1452, opened 2023-01-10
Enable required files to ensure successful build of a "stm32h7" only lib

slorquet:
Hi, to accelerate lib build time, I enabled just stm32h7 and noticed the final link failed some files were missing in the stm32h7 makefile.

Here is a patch to fix that.

Also, the uart source needs an update to use get_uart_clk instead of directly accessing the global vars.

Thanks!

### DIFF
--- a/lib/stm32/common/usart_common_all.c
+++ b/lib/stm32/common/usart_common_all.c
@@ -52,17 +52,7 @@ usart_reg_base
 
 void usart_set_baudrate(uint32_t usart, uint32_t baud)
 {
-	uint32_t clock = rcc_apb1_frequency;
-
-#if defined USART1
-	if ((usart == USART1)
-#if defined USART6
-		|| (usart == USART6)
-#endif
-		) {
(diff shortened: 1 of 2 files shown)


slorquet:
Hi,

even if my other issue with data section is questionable, the changes in this PR are required for a successful library usage.

If you can merge it, it will be highly appreciated.

karlp:
I've merged the earlier PR for the uart clock helper, I'm guessing the missing object is still required?  I've got a stm32h723 nucleo board on the way, it was one target I didn't have a board for already, so I'll test things more coming up.

slorquet:
Yes. Let me send a new cleaner pull request, this one had two fixes in one.

slorquet:
Well, checked again and the answer is no !? I need to clarify this.

slorquet:
final answer is yes, needed ^^
See https://github.com/libopencm3/libopencm3/pull/1464
Closing this now.

Issue #1453, opened 2023-01-10
Avoid usage of initialized data ?


slorquet:
Hi,

I want to use libopencm3 for a bootloader project.
After the bootloader is built it will be merged with the main code (nuttx rtos based) using some objcopy magic.

This method would work best if I could avoid initialized data in libopencm3, which is the only place that define contents in the .data section.

I'm using a stm32, and the only libopencm3 code that uses initialized data is rcc.c that stores the clock tree (eg, list of clock rates).
This clock tree is initialized in the file via a data initializer, which uses the data section.

This could be avoided if the initializer was replaced by a function (eg rcc_setup_default, or rcc_init) that populates the clock tree default value, instead of initialized data.

I see that pre_main defined here could be the place to automatically call rcc_init.
https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/h7/vector_chipset.c

What do you think about this? Could this be done inconditionnally?

I'm willing to do the changes for all stm32 platforms. Would you accept such a change?

Thanks for your attention.

slorquet:
Hi, I am talking of:
https://github.com/libopencm3/libopencm3/blob/master/lib/stm32/h7/rcc.c#L35

This is the only place in the full libopencm3 basic chip support routines that use initialized vars.

objcopy itself is not the main reason for a change here.

Writing a bootloader means I would like something as self contained as possible.

slorquet:
screw it. I need to use littlefs in this project, and even with the correct defines to avoid them, this thing uses assert and malloc, which in turn brings the newlib impure stuff into the link and I cant get around it, so it means I have to deal with initialized data after all. Thats a shame.

It's not going to be easy.

slorquet:
I will not be able to what I want. I managed to add the bootloader code to my application ELF file because the ELF file already has an empty memory reservation for where the bootloader resides.
However, massaging the ELF file to add the initialized data section does not work for several reasons
-there is no segment associated to this new section: This means that the loader (openocd or whatever) will not load this segment into memory when programming
-the --change-section-lma option of objcopy has no effect, it is not possible to add a section to an ELF file having different load and execution addresses.

So i'm back to square one: I cannot use newlib functions (malloc, etc) in my bootloader, which are required for littlefs. So, I can ditch littlefs entirely. I'll use a raw flash zone instead.

It also means that I am indeed interested in avoiding initialized data in rcc.c. I'm sending an pull request.

slorquet:
[...]

PR #1454, opened 2023-01-10
this pull request is a demonstration showing how to avoid initialized data in libopencm3

slorquet:
here is my attempt at avoiding the only initialized data in the libopencm3 project for stm32
I just initialized the required fields from a boot hook.

This PR is not ready for inclusion yet. it's a demo for the issue discussed at https://github.com/libopencm3/libopencm3/issues/1453

### DIFF
--- a/include/libopencm3/stm32/h7/rcc.h
+++ b/include/libopencm3/stm32/h7/rcc.h
@@ -708,6 +708,8 @@ enum rcc_periph_rst {
 
 BEGIN_DECLS
 
+void rcc_init(void);
+
 /**
  * Setup the base PLLs and clock domains for the STM32H7. This function will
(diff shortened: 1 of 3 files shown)


karlp:
I'm still curious why _this_ causes you problems, but other static global initialized variables are ok?  rcc_ahb_frequency, rcc_apb1_frequency and rcc_apb2_frequency variables are ~equivalent to this structure?

slorquet:
It would have been the same problem, but the h7 uses a structure instead of individual vars.

The problem was my inability to copy the bootloader initialized data section in my final ELF file.

But using the binary image of the bootloader, generated from the ELF file, works very fine, since it already embeds all the required loadable sections.

So this request and the issue are not relevant anymore. I close them.

PR #1456, opened 2023-01-12
stm32: When setting baudrate of a USART, first get the clock frequency for that USART.

harrygeyer:
Hi,
A quick commit to get the clock frequency for the USART before calculating the value for the desired baudrate.
Thanks.

### DIFF
--- a/lib/stm32/common/usart_common_all.c
+++ b/lib/stm32/common/usart_common_all.c
@@ -52,7 +52,7 @@ usart_reg_base
 
 void usart_set_baudrate(uint32_t usart, uint32_t baud)
 {
-	uint32_t clock = rcc_apb1_frequency;
+	uint32_t clock = rcc_get_usart_clk_freq(usart);
 
 #if defined USART1
 	if ((usart == USART1)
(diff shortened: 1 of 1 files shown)


lib/stm32/common/usart_common_all.c
```diff
@@ -52,7 +52,7 @@ usart_reg_base
 
 void usart_set_baudrate(uint32_t usart, uint32_t baud)
 {
-	uint32_t clock = rcc_apb1_frequency;
+	uint32_t clock = rcc_get_usart_clk_freq(usart);
```
karlp, line 55: you need to delete the lines after this too.... seeing as you are now handling clock selection elsewhere.

I'm guessing you weren't using usart1/6? :)

slorquet:
Hi,
This pull request has the correct fix. Can you push it or tell me what prevents you from doing so?
https://github.com/libopencm3/libopencm3/pull/1452

Thanks

slorquet (approved):
Thanks for the fix.

karlp:
This has merged in 4d7a2ca2.  It required e643f663423f6 first, or g4 would never have linked properly.  Thanks for pushing this forwards!

slorquet:
Thank you for merging!

PR #1458, opened 2023-01-13
STM32F0/G0/G4/L0: ADC: add missing macros, fix mask/shift style

neoxic:
Hello!

I extensively use libopencm3 in my projects for bare-metal programming. In particular, here's the link to a piece of BLDC motor control firmware that runs on the F0 and G0 chips: https://github.com/neoxic/ESCape32

Recently, I added support for ADC for current/temperature/voltage measurements to that project, but I'm unable to publish it until the following PR is accepted and merged. So hopefully, you'll find these fixes consistent and useful. Besides being successfully tested and used in the above project, they have also been checked against libopencm3-examples (no conflicts found). I made explicit effort to maintain your code style (tab size 8, etc.) and made a couple of corrections here and there where it was necessary.

The PR consists of three parts:

1. G0: add ADC compatibility macros that already exist across other families.
2. F0/G0/L0: add `ADC_CFGR1_EXTSEL` values.
3. G0/G4: fix `*_SHIFT/*_MASK` macros that didn't match the general approach (breaking change).

PR #1468, opened 2023-02-06
SPI v3 implementation for stm32h7

slorquet:
Including reorganization of existing stm32 spi cores.

Rationale:

The old SPI code hierarchy was:
spi_common_v1 -> spi_common_all
spi_common_v2 -> spi_common_all

But the spiv3 changes a lot of small details, so most of the spi_common_all file cannot be reused without conditional compilation (which is not welcome if I understand correctly) : Many functions work with the same logic, but use same (or different) bits in different registers.

So only the really common code was kept in spi_common_all, and everything that is common to spiv1 and spiv2 was moved to a new file spi_v12, resulting in the new hierarchy:

spi_common_v1 -> spi_common_v12 -> spi_common_all
spi_common_v2 -> spi_common_v12 -> spi_common_all
spi_common_v3 -> spi_common_all

I tried to avoid this but it cannot be avoided because most of the bits that were in common regs in v1 and v2 are now in new registers.

I know that the copyright of the files is not correct and I have no idea what else to put there

I did my best not to add dumb mistakes but I am ready to change anything required to bring this upstream.

Thank you for your review.


include/libopencm3/stm32/common/spi_common_all.h
```diff
@@ -51,342 +51,11 @@ specific memorymap.h header before including this header file.*/
...
-void spi_enable_error_interrupt(uint32_t spi);
-void spi_disable_error_interrupt(uint32_t spi);
-void spi_enable_ss_output(uint32_t spi);
-void spi_disable_ss_output(uint32_t spi);
-void spi_enable_tx_dma(uint32_t spi);
-void spi_disable_tx_dma(uint32_t spi);
-void spi_enable_rx_dma(uint32_t spi);
-void spi_disable_rx_dma(uint32_t spi);
-void spi_set_standard_mode(uint32_t spi, uint8_t mode);
+void spi_reset(uint32_t spi_peripheral);
```
slorquet, line 58: This is the only common function for all three spi !

include/libopencm3/stm32/common/spi_common_v12.h
```diff
@@ -0,0 +1,390 @@
+/** @addtogroup spi_defines
+ *
+ * @author @htmlonly &copy; @endhtmlonly 2009 Uwe Hermann <uwe@hermann-uwe.de>
```
slorquet, line 3: Did not change this because I copied the file, but the contents is not written by me.

lib/stm32/common/spi_common_v12.c
```diff
@@ -0,0 +1,647 @@
...
+@author @htmlonly &copy; @endhtmlonly 2023
+Sebastien Lorquet <sebastien@lorquet.fr>
+
+Functions common to spiv1 and spiv2, but not usable in spiv3.
+*/
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2023 Sebastien Lorquet <sebastien@lorquet.fr>
```
slorquet, line 13: This is not fully correct. We need to include other contributors who almost got this correct.

lib/stm32/common/spi_common_v12.c
```diff
@@ -0,0 +1,647 @@
...
+The SPI peripheral is enabled.
+
+@todo Error handling?
+
+@param[in] spi Unsigned int32. SPI peripheral identifier @ref spi_reg_base.
+*/
+
+void spi_enable(uint32_t spi)
+{
+	SPI_CR1(spi) |= SPI_CR1_SPE; /* Enable SPI. */
```
slorquet, line 44: Hit me on space versus tabs ! I'm ready!

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,963 @@
...
+@endcode
+
+@todo need additional functions to aid ISRs in retrieving status
+
+*/
+
+/*
+ * This file is part of the libopencm3 project.
+ *
+ * Copyright (C) 2009 Uwe Hermann <uwe@hermann-uwe.de>
```
slorquet, line 40: Also not correct...

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,963 @@
...
+ * @param[in] spi SPI peripheral identifier @ref spi_reg_base.
+ * @param[in] size size of next frame. If given as 0xFFFF, then transfer is
+ * infinite. Cant be used with CRC enabled.
+ */
+void spi_set_transfer_size(uint32_t spi, uint16_t size)
+{
+    uint32_t reg32;
+
+	reg32 = SPI_CR2(spi);
+    reg32 &= ~(SPI_CR2_TSIZE_MASK << SPI_CR2_TSIZE_SHIFT);
```
slorquet, line 469: ooops, fixes incoming.

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,963 @@
...
+@param[in] clkdiv Unsigned int8. Baudrate clock divider 0 (div/2) - 7(div/128).
+@param[in] cpol bool. Clock polarity, false for 0, true for 1.
+@param[in] cpha bool. Clock Phase, false for 0, true for 1.
+@param[in] lsbfirst bool. Frame format lsb/msb first.
+@param[in] keepgpio bool. True to keep GPIO controlled by SPI block.
+@returns int. Error code.
+*/
+
+int spi_init_master(uint32_t spi, uint8_t clkdiv, bool cpol, bool cpha,
+		    bool lsbfirst, bool keepgpio)
```
slorquet, line 849: at least do we agree on this definition? should keepgpio be predefined here and modifiable by a pair of dedicated functions?

lib/stm32/common/spi_common_v3.c
```diff
@@ -0,0 +1,963 @@
...
+		SPI_CFG2(spi) |= SPI_CFG2_CPHA;
+	} else {
+		SPI_CFG2(spi) &= ~SPI_CFG2_CPHA;
+	}
+
+	if (lsbfirst) {
+		SPI_CFG2(spi) |= SPI_CFG2_LSBFRST;
+	} else {
+		SPI_CFG2(spi) &= ~SPI_CFG2_LSBFRST;
+	}
```
slorquet, line 874: BTW why are these spi_init_master functions in all spi versions not calling the individual access routines?

include/libopencm3/stm32/common/spi_common_v3.h
```diff
@@ -0,0 +1,415 @@
...
+#define SPI2_DR8		SPI_DR8(SPI2_BASE)
+#define SPI3_DR8		SPI_DR8(SPI3_BASE)
+
+/* CRCL: CRC Length */
+/****************************************************************************/
+/** @defgroup spi_crcl SPI crc length
+ * @ingroup spi_defines
+ *
+ * @{*/
+
```
karlp, line 269: all the definitions below here looks to have been badly copied from somewhere else.

slorquet: Thanks, yes it seems so. I based my code on someone else's pull request.

I will remove the definitions for specific peripherals.

slorquet: I see definitions are shifted. I'll have a look at your review of the adc code to understand your exact requirements.

slorquet: Well I'm loosing my mind, what is your recommendation for register bits definitions? Do you have an example of code you like so I can copy it?

slorquet: It would help a lot if your comments were more actionable
What does your comment mean?
-is it better to remove these copied lines ?
-is it okay to keep them but make sure they match the fact that stm32h7 has 6 possible SPIs?
At the moment I cant tell.

Sorry for the verbosity.

slorquet:
I have fixed tab and space issues.

lib/stm32/common/spi_common_v3.c
slorquet: took care of spaces/tabs

slorquet:
Please review again.

slorquet:
Wow, this pull request has become so cold it will start to freeze.

josuah:
What about a first shot only introducing `spi_common_v3`?

Before:
spi_common_v1 (+1/-1) -> spi_common_v12 (+123/-0) -> spi_common_all (+123/-123) 
spi_common_v2 (+1/-1) -> spi_common_v12 (+123/-0) -> spi_common_all (+123/-123) 
spi_common_v3 (+1/-1) -> spi_common_all (+123/-123) 

After:
spi_common_v1 (+0/-0) -> spi_common_all (+0/-0)
spi_common_v2 (+0/-0) -> spi_common_all (+0/-0)
spi_common_v3 (+123/-0)

In practice, given as you said,there is only `spi_reset` that is common to all, this would not be so hard to handle as an edge case.

The the patch would be very small: just addition of the extra code, maintaining the v1/v2 unchanged, easier to review, and then merging (the difficult part) could be done separately?

Hope that this helps in deciphering why these changes would get blocked from merging in silence.

include/libopencm3/stm32/common/spi_common_v3.h
josuah: It is using a tab stop of 4px wide, so it looks misaligned and as you saw, pull request with just space adjustment take ages to get merged in, so would it be possible to adjust that straight in?

As seen in #1458 the way ahead seems to be embedding the shift in the code rather than in the register definitions.

Just caching @karlp's previous answers, not sure if I am 100% accurate.

josuah:
If only I had the time to...

1. fork
2. write a coding style covering the content of each port: how the code and API (both #defines and C functions) should be structed
3. rewrite all the code to match that coding style
4. build something like the redcuttle project to test libopencm3 on silicon for all ports
5. offer a lightweight patch to everyone using libopencm3 to move over to that fork
6. once done, offer that fork to be the new version of libopencm3.

josuah:
This would be all about "breaking everything" without breaking the projects relying on libopencm3, much alive themself with commits every weeks.

PR #1476, opened 2023-03-30
Small makefile change to clean up debugging path resolutions.

adamcbraun:
Changed the compilation command to utilize absolute paths for source files. This helps resolve file locations during debugging more cleanly.

This change avoids confusion in debuggers with trying to resolve relative paths to track down the source files.

### DIFF
--- a/lib/Makefile.include
+++ b/lib/Makefile.include
@@ -38,7 +38,7 @@ $(SRCLIBDIR)/$(LIBNAME).a: $(OBJS)
 
 %.o: %.c
 	@printf "  CC      $(<F)\n"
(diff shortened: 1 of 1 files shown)


f4grx:
Hi,

As a simple contributor I can't see any problem with this change, so I approve it, but @karlp always have interesting and accurate remarks about the structure of this project, and I cant speak for them.

Did you make sure that no other place in the build infrastructure would benefit from a similar change? Thats the only detail I cant think about.

Thank you.

esden:
I am bit afraid that this will cause the maximum command line length to cause issues. Especially when people put the project in some annoyingly deep directory. This does happen more often than not.

Should we make this some kind of switch? This seems to be just a tool to debug the makefile execution and not needed on the daily basis.

PR #1478, opened 2023-04-24
Add stm32f7 HS USB driver

jamesturton:
This PR aims to upstream the stm32f7 high speed USB driver which was developed for the [Blackmagic Debug](https://github.com/blackmagic-debug/blackmagic) project.

This PR consists of two commits. The first is just adding more register definitions which I really hope won't cause too many objections to merging and will hopefully be easy to review.

The second commit adds the usb_f732 driver and doesn't modify any of the other code. I hope this way it is more acceptable to be merged as not to change any existing usb implementations, but just add the USB HS support for this chip.


lib/usb/usb_f723.c
```diff
@@ -0,0 +1,485 @@
...
+
+	OTG_HS_DOEPMSK |= OTG_DOEPMSK_STUPM | OTG_DOEPMSK_XFRCM;
+	OTG_HS_DIEPMSK |= OTG_DIEPMSK_XFRCM;
+
+	OTG_HS_DCTL &= ~OTG_DCTL_SDIS;
+
+	return &_usbd_dev;
+}
+
+/* Note: the original 'usb_dwc_common.c' source code
```
stoyan-shopov, line 111: This comment should belong to the `stashed_packets` array variable, at lines 36-41

Issue #1482, opened 2023-06-05
Incorrect interpretation of disconnected parameter in st_usbfs_v2_disconnect


dhoepfl:
According to the [documentation of `usbd_disconnect`](https://github.com/libopencm3/libopencm3/blob/10cfbd765269748697d14b983c94dcb57848b146/include/libopencm3/usb/usbd.h#L180), the `disconnected` parameter is to be set to “true to request a disconnect”.

The [implementation for V2 drivers](https://github.com/libopencm3/libopencm3/blob/10cfbd765269748697d14b983c94dcb57848b146/lib/stm32/st_usbfs_v2.c#L88) (as used e.g. for STM32F0xx) is as follows:

```
static void st_usbfs_v2_disconnect(usbd_device *usbd_dev, bool disconnected)
{
        (void)usbd_dev;
        uint16_t reg = GET_REG(USB_BCDR_REG);
        if (disconnected) {
                SET_REG(USB_BCDR_REG, reg | USB_BCDR_DPPU);
        } else {
                SET_REG(USB_BCDR_REG, reg & ~USB_BCDR_DPPU);
        }
}
```

The [reference manual for STM32F0x1/x2/x8](file:///C:/Users/DaHoe/Desktop/rm0091-stm32f0x1stm32f0x2stm32f0x8-advanced-armbased-32bit-mcus-stmicroelectronics.pdf) says:

> Bit 15 DPPU: DP pull-up control
This bit is set by software to enable the embedded pull-up on the DP line. Clearing it to ‘0’
can be used to signalize disconnect to the host when needed by the user software.

Thus, if `disconnected` is true, the bit should be *cleared* but it is set.

The two branches of the `if` should be swapped.

(The [LM4F implementation](https://github.com/libopencm3/libopencm3/blob/10cfbd765269748697d14b983c94dcb57848b146/lib/usb/usb_lm4f.c#L572) seems to be correct.)

karlp:
oh, thanks for closing, normally the "Fixes #blah" tag does the right thing, but it missed it here, oh well.

PR #1483, opened 2023-06-05
Fixes issue #1482: Incorrect interpretation of disconnected parameter in st_usbfs_v2_disconnect

dhoepfl:
The `disconnected` parameter was interpreted incorrectly: When set to true it connected but should disconnect.


karlp:
merged via rebase in eaea8dc1 thanks!

PR #1500, opened 2023-09-27
Feature: individual STM32 timer oc mode calls

dragonmux:
When implementing a new feature in Black Magic Debug to soft-start the target power output, we wound up having too interact with the timers and set up the output compare block mode for a channel. This pulls in an extra ~400 bytes of code due to the complex unoptimisable nature of the switch-case-in-switch-case setup of the STM32 `timer_set_oc_mode()` function.

This PR addresses some of this by splitting out the case logic of the outer switch-case into per-channel functions (still exposed by this unified mode set function), saving around 332 bytes when only a single channel is used. It otherwise maintains the exact same logic and call structure for existing code.


esden (approved):
LGTM

PR #1502, opened 2023-10-03
STM32G4 swapped FDCAN interrupt isr FDCAN1_IT0 <> FDCAN1_IT1

TomasPech:
The swapped lines in irq.json generated the wrong nvic.h, then jumping into the wrong ISR.

See data sheet RM0440 (442, table 97).
image


karlp:
Thanks for catching this!  I've merged this via rebase in 1acf0c1f but do look at 1f3abd43 as well, as I've unified the interrupt naming to fdcanX_itY to be consistent with other parts, as well as the reference manual

PR #1510, opened 2023-11-06
Update Makefile fix issue

PortiESP:
# Fixed issue from Stack Overflow

https://stackoverflow.com/questions/36365752/make-error-127-when-running-trying-to-compile-code

## Issue description

The command `make` runs into an error (shown in the image below) were the compilation of the `libopencm3/` directory fails to build several files
image

It looks like in the makefile at `libopencm3/makefile` is missing the variable where the default compiler should be defined, since this variable (`MAKE`) is used in the rest of the code, the execution of `make` command fails because (_in the makefile_) the `MAKE` variable is evaluated as _undefined_. This explains why we face the **error code 127**, because there is no command installed known as `undefined`

## Solution

The solution is simple, just add the missing variable in the makefile so the `MAKE` variable is evaluated as it should be (_the make command_)

```
MAKE := make
```

> This solition worked for me since I encounterd the same issue as the stack overflow question

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,9 @@ Q := @
 MAKEFLAGS += --no-print-directory
 endif
 
+# Define the make command
+MAKE := make
+
 # Avoid the use of shell find, for windows compatibility
 IRQ_DEFN_FILES  := $(foreach TARGET,$(TARGETS),$(wildcard include/libopencm3/$(TARGET)/irq.json))
 NVIC_H := $(IRQ_DEFN_FILES:%/irq.json=%/nvic.h)


karlp:
You're going to have to explain what that's meant to be trying to fix, and how it's fixing things for you first.  A link to a stack overflow post is not a commit message.

PortiESP:
> You're going to have to explain what that's meant to be trying to fix, and how it's fixing things for you first. A link to a stack overflow post is not a commit message.

Updated the PR comment

karlp:
yeah, no way.  There's something else going on, or _no-one_ would be able to build.  We've not run into this... _ever_ and I've never seen a makefile with "make=make" before.  make=gmake, yes, but make=make? no.   Something else is going on.

What system are you on, what version of make, from a clean checkout, record exactly what commands you run and what you get.  Linking to an 8 year old unrelated stack overflow description for "error 127" and saying "this fixes it" isn't good enough for me at least.  The original stack overflow post isn't even _using_ libopencm3

PR #1518, opened 2024-01-02
Update Makefile - stm32G4

TomasPech:
Probably missing line with "desig" support?

### DIFF
--- a/lib/stm32/g4/Makefile
+++ b/lib/stm32/g4/Makefile
@@ -40,6 +40,7 @@ OBJS += cordic_common_v1.o
 OBJS += crs_common_all.o
 OBJS += crc_common_all.o crc_v2.o
 OBJS += dac_common_all.o dac_common_v2.o
+OBJS += desig_common_all.o desig_common_v1.o
 OBJS += dma_common_l1f013.o
 OBJS += dmamux.o
 OBJS += fdcan.o fdcan_common.o


karlp:
You need to fill me with confidence that you've tested this :)  a "patch-2" editing inline on github doesn't do that for me.

TomasPech:
OK. So I tested and it works :o).

Only one function `void desig_get_unique_id(uint32_t *result)` from both files **desig_common_v1.c** and **desig_common_all.c** is tight to the HW by addresses defined in _include/libopencm3/stm32/g4/**memorymap.h**_ as: `DESIG_FLASH_SIZE_BASE , DESIG_UNIQUE_ID2, DESIG_UNIQUE_ID1, DESIG_UNIQUE_ID0`

Then it works. Have you gained enough trust?

TomasPech:
> You need to fill me with confidence that ...

Is it possible to be more specific how to do this? I would like to have it into "master" because in future projects with STM32G4 and their updates I have to remember it and add the line manualy. 
Maybe I did pull RQ incorrectly. If so, please let me know what I did wrong.

thx

karlp:
so, as a heuristic, the "patch-n" form indicates someone editing patches within github, which normally means it hasn't been ever compile tested.  Starting your only commentary with "probably" and a question mark just doesn't really inspire confidence :)   It just puts things off on the wrong foot.  But yes, many of these peripherals are common, they just need someone to verify and add them.  I've merged this now, via rebase, after editing the commit message. Thanks for reporting it works!

PR #1521, opened 2024-01-05
stm32g[01]: Fix device.data file 

PocketPi:
Update information in device.data file to actually match what is written by STM.

Also convert letters from upper case (BC) to lower case (bc) to make regex parsing pass.


karlp:
merged via rebase in 3272a7a1 thanks!

Issue #988, opened 2018-11-06
ADC support missing for STM32 F2


sgoll:
There currently is no ADC definition for the STM32 F2 series. It seems it was lost when the ADC headers were unified in commit 3eaeaf693c02eba7aee0119d37c1dff3b16b4765 back in 2014:

```
This pulls out all the common header definitions for the F1, L1, F4 and F37x
parts.  It's verified against the datasheet for F2 as well, but we don't have
any good F2 test boards or any support for that yet.  (The F2 header would be
_exactly_ the same as the F4 header, so it's a target for a future round of
unification, not this one)
```

Would it be possible to add F2 support back in? I have a project that runs on F2 and I'd be willing to help with that where possible.

sgoll:
Possibly related pull requests: #231, #233

karlp:
As far as I can tell, there has simply never been adc support for F2, it wasn't removed in that commit, that commit just noted that it was probably fairly straightfoward to add it.

I'd have no problem with you doing this, you probably just want to ...
* edit include/stm32/adc.h to include f2.
* add include/stm32/f2/adc.h with the doxygen stub, and have it include the common_v1 header.
* profit?

If you want lots of definitions that are in the f4 file, then you need to look at extracting a common file there, rather than just copying it over.

karlp:
(you would probably end up creating an _v1_multi.h file, that pulls up all the shared definitions of L1/F2/F4/F37x. which are same, but different to f1)

PR #1525, opened 2024-02-04
unified STM32 F2 and F4 ADC peripheral defines and added F2 include

benfarkas98:
Based on suggestion in Issue #988

### DIFF
--- a/include/libopencm3/stm32/adc.h
+++ b/include/libopencm3/stm32/adc.h
@@ -24,6 +24,8 @@
 #       include <libopencm3/stm32/f0/adc.h>
 #elif defined(STM32F1)
 #       include <libopencm3/stm32/f1/adc.h>
+#elif defined(STM32F2)
+#       include <libopencm3/stm32/f2/adc.h>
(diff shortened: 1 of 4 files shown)


include/libopencm3/stm32/f4/adc.h
```diff
@@ -1,200 +1,6 @@
...
-
-
-BEGIN_DECLS
-
-void adc_set_multi_mode(uint32_t mode);
-void adc_enable_vbat_sensor(void);
-void adc_disable_vbat_sensor(void);
-
-END_DECLS
+#include <libopencm3/stm32/common/dma_common_f24.h>
```
karlp, line 4: sure about this?

benfarkas98: d'oh! I only built my project with F2 support, so I missed this erroneous copy-paste.

karlp: That's one of the reasons I was against supporting people building only part of the project :)

PR #1530, opened 2024-03-03
stm32g4: fix ST_TSENSE_CAL2_110C -> ST_TSENSE_CAL2_130C

neoxic:
image


karlp:
thanks, merged in 0ee2477d
I'm going to ignore g0 being a mess of no calibration, single point and dual point calibration for now...

PR #1531, opened 2024-03-05
stm32g4: fix NVIC "tim1_trg_tim17" -> "tim1_trg_com_tim17"

neoxic:
image

### DIFF
--- a/include/libopencm3/stm32/g4/irq.json
+++ b/include/libopencm3/stm32/g4/irq.json
@@ -26,7 +26,7 @@
         "exti9_5",
         "tim1_brk_tim15",
         "tim1_up_tim16",
-        "tim1_trg_tim17",
+        "tim1_trg_com_tim17",
         "tim1_cc",
         "tim2",
         "tim3",


karlp:
Hrm, no, nack on this. It's already close enough, it's a breaking change, (worse, it's a breaking change that won't cause a compile failure, people's existing interrupt handlers will simply stop working)  and your proposed name is simply a _different_ abbreviation/simplification of the name.  I don't see this as an error, nor as any sort of "fix" that must be done.

neoxic:
https://github.com/libopencm3/libopencm3/commit/708fe1516c33d77b24d49926a11bc87dfe88d0fd
https://github.com/libopencm3/libopencm3/commit/5796f41caaa7352cdd7c171d2876d230ef2cd25d

So these were okay for some reason, but the above is not? Mind you, they were also "close enough" and breaking changes, and people (me included) had to change their code.

neoxic:
> worse, it's a breaking change that won't cause a compile failure, people's existing interrupt handlers will simply stop working

Those interrupt handlers first need to be enabled, and that will generate a compilation failure. For example:
```
nvic_set_priority(NVIC_TIM34_IRQ, 0x40);
nvic_enable_irq(NVIC_TIM34_IRQ);
```

PR #1540, opened 2024-04-12
stm32g4/fdcan: Fix wrong ILE value computation

iqyx:
As reported by zar@libera.chat, the computation doesn't work as intended.


royh99 (commented):
This is not correct as the value of irq is either FDCAN_ILE_INT0 or FDCAN_ILE_INT1, i.e. 1 or 2

PR #1490, opened 2023-07-23
stm32g4: rcc: Add support for 24MHz external clocks

bene42:
ST decided to use 24 MHz crystals on the Nucleo 64 boards NUCLEO-G431RB, NUCLEO-G474RE and NUCLEO-G491RE.
Unfortunately there is no support for 24 MHz clocks in libopencm3.
Add support for 24 MHz clocks to ease development on these Nucleo-64 boards.

PR #1548, opened 2024-06-28
stm32g4:  rcc: added hse config for 24mhz

jojo7682345:


esden:
This was added with #1490 already, which is now merged. Thank you for the contribution.

PR #1557, opened 2024-08-11
Feature: STM32H72/H73 support

dragonmux:
This PR works through correcting the STM32H7 support and adding support for the H72/H73 parts to allow them to be used with USB, DMA and several other components that were missing.

NB: Because of some issues in assumptions libopencm3 makes about QuadSPI controllers, this does introduce a breaking change into the QuadSPI support to allow there to be more than one controller - the H72/H73 parts have two.


esden (dismissed):
LGTM

That is quite a big hunk of support added. Thank you.

The backward compatibility break is unfortunate but unavoidable.

esden (approved):
LGTM

dragonmux:
Should be all fixed now :)

PR #1565, opened 2024-10-22
Adds functions to enter all of the L4 power modes

andrewmcg1:
Added the following 7 functions which can be used to enter all of the power modes available on the L4. 
- `scb_set_deepsleep(void)`
  - sets the deepsleep bit in the m3 registers. This bit controls whether or not the cpu enter sleep mode or one of the deeper sleeps
- `scb_clear_deepsleep(void)`
  - clears the deepsleep bit
- `scb_set_sleeponexit(void)`
  - sets the sleeponexit bit in the appropriate register. Setting this bit will make the cpu enter into a low power mode when exiting an interrupt
- `scb_clear_sleeponexit(void)`
  - clears the sleeponexit bit
- `pwr_set_low_power_run(void)`
  - sets the low-power run bit. This bit controls whether the L4 will enters into low-power run or low-power sleep
- `pwr_clear_low_power_run(void)`
  - clears the low-power run bit
- `pwr_set_low_power_mode_selection(lpms)`
  - Sets the lpms bits in the pwr_cr1 register. Setting these bits is how the cpu reaches the stop modes and standby and shutdown modes. This function uses existing defines in pwr.h as parameters to enter stop modes and standby and shutdown modes

### DIFF
--- a/include/libopencm3/cm3/scb.h
+++ b/include/libopencm3/cm3/scb.h
@@ -554,6 +554,10 @@ struct scb_exception_stack_frame {
 	} while (0)
 
 void scb_reset_system(void) __attribute__((noreturn));
+void scb_set_sleepdeep(void);
+void scb_clear_sleepdeep(void);
+void scb_set_sleeponexit(void);
+void scb_clear_sleeponexit(void);
 
 /* Those defined only on ARMv7 and above */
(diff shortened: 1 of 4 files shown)


esden (approved):
LGTM, thanks for the contribution

PR #1566, opened 2024-11-01
Add support for overdrive for STM32F4

vdanjean:
Some STM32F4 boards such as stm32f446re are able to enable overdrive.

I added the constants and functions (mainly copying from the stm32f5 corresponding files).
As overdrive should generally be set while setting clocks, I renamed `rcc_clock_setup_pll` into `rcc_clock_setup_pll_x`
so that I can add a boolean parameter (to enable overdrive or not)
and I provided a `rcc_clock_setup_pll` function for backward compatibility.
Indeed, calling `pwr_enable_overdrive` before or after the previous `rcc_clock_setup_pll`  does not work.

I'm sure the design can be improved but, being new to libopencm3, I just share this code. If you tell me which design should be used, I can rewrite this patch.

Regards,
  Vincent

### DIFF
--- a/lib/stm32/f4/rcc.c
+++ b/lib/stm32/f4/rcc.c
@@ -785,8 +785,9 @@ uint32_t rcc_system_clock_source(void)
  * needed to establish a system clock.
  *
  * @param clock clock information structure.
+ * @param overdrive boolean to request OverDrive.
  */
-void rcc_clock_setup_pll(const struct rcc_clock_scale *clock)
+void rcc_clock_setup_pll_x(const struct rcc_clock_scale *clock, int overdrive)
 {
 	/* Enable internal high-speed oscillator (HSI). */
 	rcc_osc_on(RCC_HSI);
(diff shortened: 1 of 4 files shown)


lib/stm32/f4/rcc.c
```diff
@@ -785,8 +785,9 @@ uint32_t rcc_system_clock_source(void)
  * needed to establish a system clock.
  *
  * @param clock clock information structure.
+ * @param overdrive boolean to request OverDrive.
  */
-void rcc_clock_setup_pll(const struct rcc_clock_scale *clock)
+void rcc_clock_setup_pll_x(const struct rcc_clock_scale *clock, int overdrive)
```
esden, line 790: I think it would make more sense to call this function `rcc_clock_setup_pll_od` for OverDrive, it is more descriptive why there is a suffix to the function.

esden (changes requested):
Otherwise this PR looks good to me and we should be able to merge it after you make the naming change.

PR #1568, opened 2025-01-13
STM32L4: Add missing and remove non-existent peripherals according to RM

neoxic:
This PR contains fixes to RCC registers and memory map in order to match RM0394.

1) Non-existent peripherals removed: FMC3, TIM4, TIM5, TIM8, TIM17, UART5, CAN2, SAI2, GPIOF, GPIOG

2) Missing bits for peripherals added: I2C4, CRS, WWDG, RTCAPB

3) Temperature sensor calibration value changed from 110˚C to 130˚C according to the datasheet:

pic1


karlp:
Please remember that there are _other_ L4's than just those covered by RM0394 :)  
See also 
image

At least for the bit defns, the only option is really to go for a superset, that's been the normal approach.  As for the temp cal data, no idea, those cal bits have always been messy and inconsistent, they've not gotten any better, either way, removing stuff and hardswitching isn't going help much long term.

neoxic:
> Please remember that there are other L4's than just those covered by RM0394

Ahh, my bad. I was actually after adding a couple missing bits before going full bore. I'll try again, thanks!

> As for the temp cal data, no idea, those cal bits have always been messy and inconsistent

Hmm... Probably, a better idea would be to just call them `ST_TSENSE_CAL1` and `ST_TSENSE_CAL2` without any reference to temperature and forget about that.

neoxic:
As a follow-up: https://github.com/libopencm3/libopencm3/pull/1569

PR #1550, opened 2024-06-28
stm32g4: exti: added exti support

jojo7682345:

PR #1489, opened 2023-07-23
stm32g4: Add EXTI support

bene42:
The STM32G4xx series contain two regular EXTI v1 blocks; one at offset 0x00, the other at offset 0x20.
Add support for EXTI.

This pull request partly overlaps with pull request #1250.

PR #1570, opened 2025-01-29
stm32g4: exti: add exti support

elionne:
Continuing #1550.

exti common is compatible with stm32g4. It also use exti v1 not v2 api.


esden:
Hi @elionne,

Thank you for improving this PR. But I see commits related to SPI4 addition that don't belong here. Can you move them out of this PR so I can merge it?

Cheers,
Piotr

esden:
Decided to merge the original PR as it adds the correct pointers too. #1489

PR #1578, opened 2025-04-01
Add LPUART1 definition to usart.h for stm32g4x

blutack:
Add LPUART1 definition for the STM32G4xx LPUART as per the L4 family.
Without the LPUART1 definition, usart_set_baudrate will calculate BRR incorrectly and the UART will not work due to an ifdef. Tested working @ 115200 on a NUCLEO-G474RE.


esden (approved):
Thank you for the contribution. Merging.

PR #1591, opened 2025-05-26
Add support for LPUART1 for STM32G4, fix compile time baud rate formula selection 

n0stalgic:
Tested on NUCLEO-G474RE eval board.

Added support for LPUART1 on STM32G4 and fixed the ifdef to select baud rate in usart_set_baudrate(). LPUART1 gets defined after the check for LPUART1, resulting in incorrect baud rate when using the function on LPUART1


esden:
This patch was previously submitted in #1578. So I am closing this in favor for the previous version.

PR #1594, opened 2025-06-26
Correct shebang lines to get python3

manuwarfare:
Change shebang line to use python3 instead of previous versions of python


karlp:
This isn't even _trivially_ correct.  You _clearly_ haven't updated these scripts, let alone tested them.

PR #1597, opened 2025-07-12
Fix/chat and CI References

esden:
The chat moved as gitter was absorbed by matrix. And travis is pretty bitrotten by now.


dragonmux (approved):
LGTM - feel free to merge this when ready. We should add a badge for CI back in once an alternative such as GHA can be established.

PR #1598, opened 2025-07-12
Feature: Meson build support

dragonmux:
In this PR we add support for using the project with the Meson build system. While not a 100% complete port of the Makefile build system, this provides support for all the platforms BMD uses and a few extra we've had people like @northernpaws contribute to the fork project.

This is all tested working on the upstream project and should in theory simplify the CI rebuild process due to how Meson cross files work and operate.

Importantly, this does **not** replace the existing Makefile build system but is instead implemented in tandem with it to provide options for consumers of the project.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,9 @@ include/libopencmsis/stm32/
 include/libopencmsis/swm050/
 include/libopencmsis/vf6xx/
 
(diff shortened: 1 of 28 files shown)


esden (approved):
Great! Thanks for sending this over. :)

PR #1599, opened 2025-07-12
Feature: compute effective USART baudrate

ALTracer:
This adds a function, `usart_get_baudrate()`, to all STM32 USART IP, which allows confirming effective baudrate computed from usart_ker_ck and BRR. It may differ from previously requested `usart_set_baudrate()` because of integer prescalers.

Tested for 1 year https://github.com/blackmagic-debug/blackmagic/pull/1864 and merged in downstream https://github.com/blackmagic-debug/libopencm3/pull/5 and https://github.com/blackmagic-debug/libopencm3/pull/6, on STM32F103 and STM32F411, which additionally supports OVER8. Used to report UART-encoded SWO baudrate expected by Black Magic Probe (has to match DUT-transmitted SWO to about <2% precision).
LPUART tested on G071, not tested on G4/H7.
An implementation for TM4C123 UART (Arm PrimeCell PL011) is also provided to satisfy [downstream] CI, but I didn't test it myself, because I don't have a `launchpad-icdi` board. Other vendor MCU not considered.

Function requires 32-bit division, which incurs a libgcc `__udivsi3` library call on Cortex-M0 (like many clock-computing functions) i.e. F0/G0/L0/C0/U0. Function eliminated by linker when unused.

### DIFF
--- a/include/libopencm3/lm4f/uart.h
+++ b/include/libopencm3/lm4f/uart.h
@@ -443,6 +443,7 @@ enum uart_fifo_tx_trigger_level {
 BEGIN_DECLS
 
 void uart_set_baudrate(uint32_t uart, uint32_t baud);
+uint32_t uart_get_baudrate(uint32_t uart);
 void uart_set_databits(uint32_t uart, uint8_t databits);
 uint8_t uart_get_databits(uint32_t uart);
 void uart_set_stopbits(uint32_t uart, uint8_t stopbits);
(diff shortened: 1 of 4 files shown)


dragonmux (approved):
LGTM, merging. This nicely upstreams the already merged and working versions blackmagic-debug/libopencm3#5 and blackmagic-debug/libopencm3#6 so we have every confidence this is correct.

PR #1601, opened 2025-07-12
gadget0: Add WeActStudio.MiniSTM32F4x1 board

ALTracer:
This adds another F4-based board to gadget zero test suite. Also known as `blackpill-f411ce` platform in Black Magic Debug sourcetree.

I believe the board is very suitable for HITL setups due to being cheaper, smaller than 32F407VG Discovery (MB997) and 32F429ZI Discovery (MB1075), because of UFQFN48 package and no onboard debugger/display/peripheral ICs. Hopefully this helps weed out remaining bugs in "USBOTG_FS" driver.
A 25MHz HSE into 96MHz Hclk PLL config is used, but be aware that 84MHz F401 chips could work with one line edited, too. The new 8MHz HSE board was not tested with this and would require a different line. Unsure if this works on Nucleo-F411RE, as it lacks an HSE and USB port (maybe its STLINK/V2-1 provides 8MHz via MCO?).

Trace is provided via in-tree ITM_Write8() to SWO pin PB3, a debugger is expected to configure it.
I'm not adding an `openocd.stm32f411ce.cfg` file because I don't use it, but if HITL requires it then I'll copy from something.

Tested in June of 2024. Copyright string might be wrong, inherited from f103-generic.

### DIFF
--- a/tests/gadget-zero/Makefile.stm32f411-generic
+++ b/tests/gadget-zero/Makefile.stm32f411-generic
@@ -1,0 +1,40 @@
+##
+## This file is part of the libopencm3 project.
+##
+## This library is free software: you can redistribute it and/or modify
+## it under the terms of the GNU Lesser General Public License as published by
+## the Free Software Foundation, either version 3 of the License, or
+## (at your option) any later version.
+##
+## This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


esden (approved):
LGTM

PR #1602, opened 2025-07-14
This adds the STM32U5 Family to devices.data

ChuckM:
Taken from RM0456, Rev 6. This adds STM32U535, STM32U545, STM32U575, STM32U585, STM32U595, STM32U5A, STM32U5F, and STM32U5G.


esden (approved):
LGTM, compared to the reference manual and it seems all correct as far as I can tell.

PR #1610, opened 2025-09-21
Add ARMv8-M.Main support (Cortex-M33)

ALTracer:
* libopencm3_stm32u5 leads to FTBFS in gadget-zero test (pending).
* This PR updates the macro guards for ITM to avoid a preprocessor error "Instrumentation Trace Macrocell not available in CM0" when compiling for STM32U5 which is Cortex-M33 and implements ARMv8-M.Main profile. There are DWT, ITM, and TPIU in it.
* There are lots more macro guards in /cm3/ for most PPB blocks, which have to do with ARMv6-M not implementing them. This does not block gadget0.
* Likewise, it updates similar macro guards in DWC2 driver so that the OTG_FS peripheral works faster (in endpoint read/write packet FIFO).

### DIFF
--- a/include/libopencm3/cm3/itm.h
+++ b/include/libopencm3/cm3/itm.h
@@ -27,8 +27,8 @@
  */
 
 /* Those defined only on ARMv7 and above */
-#if !defined(__ARM_ARCH_7M__) && !defined(__ARM_ARCH_7EM__)
-#error "Instrumentation Trace Macrocell not available in CM0"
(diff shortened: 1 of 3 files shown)


dragonmux (approved):
This LGTM, so merging. Thank you for the contribution! It would be useful to please defined what "FTBFS" means here as you're in initialism central there.. however, not going to hold this back for that.

PR #1611, opened 2025-09-21
Add support for USBOTG_FS on STM32U575

ALTracer:
* libopencm3_stm32u5 won't enable power or clocks to USBOTG_FS peripheral on STM32U585.
* Adjust a few macros, add entire usb/dwc2 stack to objects.
* Add RCC_PWR pclk-enable macro (to use in application code), ~~do not add~~ and SVMCR/SVMSR registers (which remove Power Isolation to Vddusb domain). (*upd: added as last commit*)
* Add entire CRS by hooking to common so that HSI48 is trimmed.
* Reuse the DOEPINT STUP reg poke so that on STM32U585 the Setup packet Set Address works. The rest STM32H7-related snippets are not mandatory (I've tested that).

Based on #1610, but does not exactly depend on it.


ALTracer:
GH webinterface still thinks there are 7 commits, despite PR1610 being forward-merged. There's nothing for me to rebase. It's ready for review. I created a stub stm32/u5/pwr.c driver with a single function which enables power between Vddusb and internal FS PHY. The rest (like voltage scale, EPOD/BOOST) will only be required for PLL, which gadget0 does not require.

dragonmux (dismissed):
Only got the one note, which is a suggestion - with that taken care of, we'll be happy to merge this. LGTM.

dragonmux (approved):

PR #1612, opened 2025-09-21
Add USB Gadget Zero test for generic STM32U575

ALTracer:
* STM32U575/U585 contains a USBOTG_FS peripheral with internal FS PHY, but it is different enough that it does not work out-of-the-box with existing drivers. U59x/U5Ax and U5Fx/U5Gx boast HS, that's not handled. U535/U545 contain a different USB FS host/device peripheral (like STM32G0B1 probably), that's not handled.
* Add a minimal gadget0 main.c and Makefile to be able to test driver changes (license/copyright is probably wrong again).
* Incorporate required PWR->SVMCR and OTG_FS->GCCFG, OTG_FS->GOTGCTL reg pokes in it, so that USB PHY power is enabled, and Vbus detection for B-session valid is disabled/forced valid (no resistive divider between PA9 and Vbus).

TIM6 microsecond delay works, ITM/SWO likely works, HSI48/CRS works, Hclk is default HSI16 (no PLL support yet).
I can refactor some of the code once you tell me what form it should conform to. Note there's more endpoints and FIFO space, too, but test_gadget0.py passes already. Performance is not representative of max capabilities because of delay_us(100), slow memset from newlib-nano, slow Hclk (HSI16) with disabled ICACHE, and ITM-related function calls everywhere. 386/225 kps otherwise becomes 749/511 kps.
Tested on https://github.com/WeActStudio/WeActStudio.STM32U585Cx_CoreBoard
Depends on #1610, #1611.

```
Running (user) tests for DUT:  stm32u575-generic
...............ss....................
----------------------------------------------------------------------
Ran 37 tests in 0.463s

OK (skipped=2)
Running (user) tests for DUT:  stm32u575-generic
...............read 5324800 bytes in 0:00:13.468740 for 386.0791729590147 kps
.wrote 5324800 bytes in 0:00:23.054369 for 225.55377681341005 kps
.....................
----------------------------------------------------------------------
Ran 37 tests in 36.921s

OK
```

### DIFF
--- a/tests/gadget-zero/main-stm32u575-generic.c
+++ b/tests/gadget-zero/main-stm32u575-generic.c
@@ -34,6 +34,18 @@
 	do { } while (0)
 #endif
 
+#define OTG_GOTGCTL_BVALOVAL		(1U << 7U)
+#define OTG_GOTGCTL_BVALOEN		(1U << 6U)
+
+// #include <libopencm3/stm32/u5/pwr.h>
+#define PWR_SVMCR		MMIO32(PWR_BASE + 0x10)
+#define PWR_SVMSR		MMIO32(PWR_BASE + 0x3C)
+
+#define PWR_SVMCR_USV		(1U << 28U)
+#define PWR_SVMCR_UVMEN		(1U << 24U)
+
+#define PWR_SVMSR_VDDUSBRDY		(1U << 24U)
+
 int main(void)
 {
 	rcc_clock_setup_hsi(&rcc_hsi16mhz_configs);
@@ -41,7 +53,7 @@ int main(void)
 	while (!rcc_is_osc_ready(RCC_HSI48));
 	crs_autotrim_usb_enable();
 	rcc_periph_clock_enable(RCC_GPIOA);
-	rcc_periph_clock_enable(RCC_OTGFS);
+	//rcc_periph_clock_enable(RCC_OTGFS);
 
 	gpio_mode_setup(GPIOA, GPIO_MODE_AF, GPIO_PUPD_NONE, GPIO11 | GPIO12);
 	gpio_set_af(GPIOA, GPIO_AF10, GPIO11 | GPIO12);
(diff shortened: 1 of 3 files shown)


tests/gadget-zero/main-stm32u575-generic.c
```diff
@@ -0,0 +1,96 @@
...
+	gpio_set_af(GPIOA, GPIO_AF10, GPIO11 | GPIO12);
+
+	/* LED to indicate boot process */
+	rcc_periph_clock_enable(RCC_GPIOC);
+	gpio_mode_setup(GPIOC, GPIO_MODE_OUTPUT, GPIO_PUPD_NONE, GPIO13);
+	gpio_set(GPIOC, GPIO13);
+
+	/* Remove Vddusb power isolation */
+	rcc_periph_clock_enable(RCC_PWR);
+#if 0
```
dragonmux: Why the `#if 0` and manual poking? surely using the `pwr_` functions would be better and more maintainable..

dragonmux (changes requested):
Only got the one review note and with it taken care of we'll get this merged.

tests/gadget-zero/main-stm32u575-generic.c
ALTracer: Because this code predates my additions of `pwr_` functions; I wrote them later, as "extract method" and by looking at H7 API.
Removed with a new commit.

ALTracer:
This no longer compiles.
```c
delay.c: In function 'delay_setup':
delay.c:39:35: error: 'rcc_apb1_frequency' undeclared (first use in this function)
   39 |         timer_set_prescaler(TIM6, rcc_apb1_frequency / 1000000 - 1);
```
I could add a temporary in delay.c since the STM32U5 RCC API changed significantly during these 10 months.
```c
#if defined(STM32U5)
	const uint32_t rcc_apb1_frequency = rcc_get_bus_clk_freq(RCC_APB1CLK);
#endif
```

tests/gadget-zero/main-stm32u575-generic.c
```diff
@@ -0,0 +1,79 @@
...
+#endif
+
+#define OTG_GOTGCTL_BVALOVAL		(1U << 7U)
+#define OTG_GOTGCTL_BVALOEN		(1U << 6U)
+
+int main(void)
+{
+	rcc_clock_setup_hsi(&rcc_hsi16mhz_configs);
+	rcc_osc_on(RCC_HSI48);
+	while (!rcc_is_osc_ready(RCC_HSI48));
```
ALTracer: This can now be `rcc_clock_setup_hsi48()`

dragonmux: Then please do so and let us know when you have this building again.

ALTracer: Replaced two functions, 128 and 136 bytes large, with one that is 28 bytes.

ALTracer:
This compiles, runs, and passes `test_gadget0.py` now.

dragonmux (approved):

PR #1624, opened 2026-01-14
Clock functions for STM32 U5

Phylex:
Added code that configures the PLLs of the STM32U5 series chip. Also added some defines in the cm3 part of the library as the cortex m33 is slightly different that the v7m, v7em and so on.

### DIFF
--- a/include/libopencm3/stm32/u5/rcc.h
+++ b/include/libopencm3/stm32/u5/rcc.h
@@ -44,6 +44,22 @@
 /* Module definitions                                                        */
 /*****************************************************************************/
 
+
+struct rcc_pll_config {
+	uint8_t pll_clock_source; ///< specifies what clock is fed to the PLL
+	uint8_t pll_input_range; ///< pll input range specification, this is needed for pll stability
+	uint8_t pll_m; ///< input clock divider (the input clock needs to be in the range 4 - 16 MHz
+	uint8_t pll_mboost; ///< Clock divider for the power distribution booster output clock (only on pll1)
+	uint16_t pll_n; ///< pll output clock multiplier
+	uint16_t pll_frac_n; ///< pll output clock fractional multiplier
+	uint8_t pll_p; ///< output p pll clock divider
+	uint8_t pll_p_en;
+	uint8_t pll_q; ///< output q pll clock divider
+	uint8_t pll_q_en;
+	uint8_t pll_r; ///< output r pll clock divider
+	uint8_t pll_r_en;
+};
+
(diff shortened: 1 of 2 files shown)


Phylex:
Well I did not know of this fork, If I had I'd probably used that. I know that there is more involved to getting the whole clocking system working. I am currently playing around with an STM32u5g9 on the DK1 (the one with the round display). I stepped through the stm example and got my clocks configured the same way (I mean I was not as thorough with my defines, but I ended up with a working clock for the DSI and Display stuff.
Any plans merging this with 'normal' libopencm3?

Phylex:
The Idea behind this that you'd build the whole clock tree from different smaller configs. The PLL sounded like a sensible 'unit' of configuration to start with. but yea, I'll just go ahed and use that fork then.

dragonmux:
> Any plans merging this with 'normal' libopencm3?

Yes, we're still working some details to do with the GPDMA out but once we have the target hardware working properly as part of the Black Magic Debug project, that branch'll be PR'd here ahead of the PR that'll be opened on BMD for the platform that's being built for the new hardware.

Phylex:
want some help with the gpdma?

Phylex:
Do you have code that tests this? I am assuming the code is referencing the 'dma.c' file in the u5 folder of the fork that you mentioned?

Phylex:
Is there a more appropriate place to discuss the gpdma things or would this just continue here?

dragonmux:
Here or in an issue ticket is fine.. unless you wish to hop onto the 1BitSquared Discord to talk more in depth there

Regarding code for testing this, yes, dma.c is the locm3 implementation for the GPDMA; our test setup is the aux serial logic in BMD - if you wish to lift the general idea for that out into your own test firmware that uses a ring buffer and the two interrupts to move one half of that buffer from a UART to someplace else at a time, toggling back and forth between the upper and lower halfs, that should be sufficient to contrast and compare.

Phylex:
I'll have a look at the black magic debug probe usart code then.

dragonmux:
👍🏼 you'll want https://github.com/blackmagic-debug/blackmagic/blob/feature/bmpv3-support/src/platforms/common/aux_serial.c and https://github.com/blackmagic-debug/blackmagic/blob/d7097a6842facfad0a914a9770c07162c53a3f6b/src/platforms/common/usb_serial.c#L128-L180 in particular to see how BMD configures the UART and GPDMA. You can ignore the secondary UART for now as only the primary is plumbed up

PR #1632, opened 2026-03-21
Fix: st_usbfs_v2 data copy from packet memory

dragonmux:
In this PR we address the bug identified in #1631 caused by a buffer mishandling mistake in the code used to unload packets from the ST USB controller's packet memory.

What's surprising is that this hadn't bitten BMD yet, despite having a high probability of firing.

Fixes #1631

PR #1633, opened 2026-03-21
Feature: ST USBFS driver cleanup

dragonmux:
In this PR we apply some code quality improvements and general cleanup of the packet memory management code for both ST USB controller drivers to bring them both up to the same code quality and pointer arithmetic cleanliness standard.

Tested on BMPv2 for the v1 driver. The v2 driver has no functional changes and thus should remain the same as in PR #1632.


esden (approved):
LGTM

PR #1634, opened 2026-03-21
Fix: USB DWC2 on ARMv6

dragonmux:
This PR fixes a bug identified by ALTracer where the transmit endpoint packet enqueue function would only work for EP0 on ARMv6 devices due to a copypasta mistake in the data queueing function.


esden (approved):
LGTM

PR #1636, opened 2026-04-21
compilation fixes USB for stm32 H7

greenp12:
An initial set of changes to be able to compile libopencm3 with USB support on STM32 Nucleo H753ZI (for BlackMagic).

### DIFF
--- a/lib/stm32/h7/rcc.c
+++ b/lib/stm32/h7/rcc.c
@@ -15,6 +15,9 @@
 #define HZ_PER_MHZ		 	1000000UL
 #define HZ_PER_KHZ			1000UL
 
+
+uint32_t rcc_ahb_frequency = 96000000;
+
 /* Local private copy of the clock configuration for providing user with clock tree data. */
 static struct {
 	uint32_t sysclk;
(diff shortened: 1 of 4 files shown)


lib/stm32/h7/rcc.c
```diff
@@ -15,6 +15,9 @@
 #define HZ_PER_MHZ		 	1000000UL
 #define HZ_PER_KHZ			1000UL
 
+
+uint32_t rcc_ahb_frequency = 96000000;
```
dragonmux: This value should not be required as you can get it from rcc_clock_tree.hclk - which the RCC code is already calculating and tracking w/ `rcc_get_bus_clk_freq(RCC_AHBCLK)` able to retrieve. If there is code depending on `rcc_ahb_frequency` it should be fixed instead.

lib/stm32/h7/rcc.c
```diff
@@ -182,6 +185,18 @@ static void rcc_clock_setup_domain3(const struct rcc_pll_config *config)
 		rcc_prediv_3bit_log_div(rcc_clock_tree.hclk, config->ppre4);
 }
 
+static void rcc_clock_setup_cfgr(uint8_t sysclock_src, uint8_t sysclock_sws)
+{
+	uint32_t cfgr_sws = ((RCC_CFGR >> RCC_CFGR_SWS_SHIFT) & RCC_CFGR_SWS_MASK);
+	if (cfgr_sws != sysclock_sws) {
+		/* Domains dividers are all configured, now we can switchover to PLL. */
```
dragonmux: You probably want to update the comment as it doesn't apply in this function - only the original call site that could only deal with PLLs as the target clock for configuration. We suggest something like `Switch to the requested system clock and monitor for the switch-over to complete`

dragonmux (changes requested):
Couple of small tweaks and bits of feedback, and with them taken care of we'd be happy to merge this.

lib/stm32/h7/rcc.c
```diff
@@ -223,11 +235,19 @@ void rcc_clock_setup_pll(const struct rcc_pll_config *config)
...
-	while (cfgr_sws != RCC_CFGR_SWS_PLL1) {
-		cfgr_sws = ((RCC_CFGR >> RCC_CFGR_SWS_SHIFT) & RCC_CFGR_SWS_MASK);
+	switch (config->sysclock_source) {
+	case RCC_PLL:
+		rcc_clock_setup_cfgr(RCC_CFGR_SW_PLL1, RCC_CFGR_SWS_PLL1);
+		break;
+	case RCC_HSE:
+		rcc_clock_setup_cfgr(RCC_CFGR_SW_HSE, RCC_CFGR_SWS_HSE);
+		break;
+	case RCC_HSI:
```
greenp12, line 245: Note that I also added a missing case for the HSI clock here.

dragonmux (approved):
LGTM, merging. Thank you for the contribution! Really appreciate you building on our own work getting the H7 more sorted and building things out for cases we weren't able to handle at the time.

PR #1640, opened 2026-05-10
fix bug in Makefile

benzjimny:
The compilation fails when I compile the code using gnu make on windows.
The root cause is in line 65 and 69 of the Makefile, the python script executes as binary.
It is better to use python to execute the script.
I fixed this bug and tested it. The code can be compiled on windows now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,7 @@ LIB_DIRS:=$(wildcard $(addprefix lib/,$(TARGETS)))
 $(LIB_DIRS): $(IRQ_GENERATED_FILES)
 	$(Q)$(RM) .stamp_failure_$(subst /,_,$@)
 	@printf "  BUILD   $@\n";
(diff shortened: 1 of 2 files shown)


karlp:
You've made substantially more changes than what is described...

benzjimny:
hi,
What I want to make a pull request is commit 8dd861c61e9e16e3ac1773d42ef6515b17af2a2c.

After I made the pull request, I continued to updating the makefile and push to my repo on the same branch. I didn't know it would be also sync to the pull request. The "more changes" is inside the commit 7538adb24459e2991f56d74b0f2f2787e43e67ac.

In the commit **7538adb244 - move build artifacts into build folder**, I moved all the artifacts of the build process into the folder **build** which is under the project root. Please see the screenshot below.

图片

build command is

图片

The **make clean** command also take effect.

PR #1643, opened 2026-05-22
STM32H7xx SPI Port

fredericotuchtenhag-spec:
### STM32H7xx family SPI Port
Creating pull request as i've finished the main functionality of the SPI peripheral on the H7 port. 
Remapped all register values to an independent header file which doesnt inherit anything from common_v2 or common_all, chose this path as to minimize confusion, as the register mapping in the SPI peripheral on the H7 are drastically different.


include/libopencm3/stm32/h7/rcc.h
```diff
@@ -409,34 +409,36 @@ LGPL License Terms @ref lgpl_license
...
 #define RCC_D2CCIP2R_LPTIM1SEL_SHIFT        28U
 #define RCC_D2CCIP2R_CECSEL_SHIFT           22U
-#define RCC_D2CCIP2R_USBSEL_MASK            0x3U
 #define RCC_D2CCIP2R_USBSEL_SHIFT           20U
 #define RCC_D2CCIP2R_I2C123SEL_SHIFT        12U
+#define RCC_D2CCIP2R_I2C123SEL_MASK         0x7U
 #define RCC_D2CCIP2R_RNGSEL_MASK            0x3U
 #define RCC_D2CCIP2R_RNGSEL_SHIFT           8U
-#define RCC_D2CCIP2R_USART16910SEL_SHIFT    3U
+#define RCC_D2CCIP2R_USART16SEL_SHIFT       3U
```
dragonmux, line 418: Not sure why you're removing the 9 and 10 parts of this constant, but this value does control those

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -409,34 +409,36 @@ LGPL License Terms @ref lgpl_license
 
 #define RCC_D2CCIP2R_LPTIM1SEL_SHIFT        28U
 #define RCC_D2CCIP2R_CECSEL_SHIFT           22U
-#define RCC_D2CCIP2R_USBSEL_MASK            0x3U
```
dragonmux, line 412: Why are you removing this constant? It's correct, please don't do that - it breaks USB support.

include/libopencm3/stm32/h7/rcc.h
```diff
@@ -409,34 +409,36 @@ LGPL License Terms @ref lgpl_license
...
-#define RCC_D2CCIP2R_RNGSEL_LSI             3U
-#define RCC_D2CCIP2R_USART16SEL_PCLK2       0U
-#define RCC_D2CCIP2R_USART234578SEL_PCLK1   0U
-#define RCC_D2CCIP2R_USARTSEL_PCLK          0U
-#define RCC_D2CCIP2R_USARTSEL_PLL2Q         1U
-#define RCC_D2CCIP2R_USARTSEL_PLL3Q         2U
-#define RCC_D2CCIP2R_USARTSEL_HSI           3U
-#define RCC_D2CCIP2R_USARTSEL_CSI           4U
-#define RCC_D2CCIP2R_USARTSEL_LSE           5U
+#define RCC_D2CCIP2R_RNGSEL_HSI48           0
```
dragonmux: Please do not de-suffix these constants, they have the `U` suffix on them to make them unsigned so their usage doesn't cause signed-unsigned conversion errors and shift UB.

lib/stm32/common/i2c_common_v2.c
```diff
@@ -379,7 +379,7 @@ void i2c_transfer7(uint32_t i2c, uint8_t addr, const uint8_t *w, size_t wn, uint
 				if (i2c_transmit_int_status(i2c)) {
 					wait = false;
 				}
-				while (i2c_nack(i2c)); /* FIXME Some error */
+				while (i2c_nack(i2c));	/* FIXME Some error */
```
dragonmux, line 382: Not sure why this line is changing - it shouldn't.. the space between the `;` and `/` was correct.

lib/stm32/h7/Makefile
```diff
@@ -54,6 +54,7 @@ OBJS += spi_common_all.o spi_common_v2.o
 OBJS += timer_common_all.o
 OBJS += usart_common_all.o usart_common_v2.o usart_common_fifos.o
 OBJS += quadspi_common_v1.o
+OBJS += i2c_common_v2.o
```
dragonmux, line 57: Please could you also add this to the meson.build so the two are kept in lockstep.

lib/stm32/f1/Makefile
```diff
@@ -43,7 +43,7 @@ OBJS += dma_common_l1f013.o
 OBJS += exti_common_all.o
 OBJS += flash.o flash_common_all.o flash_common_f.o flash_common_f01.o
 OBJS += gpio.o gpio_common_all.o
-OBJS += i2c_common_v1.o
+OBJS += i2c_common_v1.o i2c_common_v2.o
```
dragonmux, line 46: Unsure why you're changing STM32F1 things like this.. and whether this is even correct in the first place? This has nothing to do with STM32H7 support so PR separately if correct.

lib/stm32/h7/spi.c
```diff
@@ -0,0 +1,192 @@
...
+ *                    can be configured two ways depending on the SSOE bit on SPI_CFG2;
+ * @param [in] ssom_mode Specifies the Slave Select Output Management mode, 1 or 0.3
+ * @param [in] ssoe Slave Select Output Enable, this bit should only be set when Master, if the device is the only master in the Bus,
+ *                  it should always be kept high. If Master & SSOE=0, the bus is configured for multimaster mode. if the SS pin is 
+ *                  pulled active in this mode, the SPI enters slave mode automatically.
+ * @param [in] mssi Master SS Idleness, this bits set the initial delay between the SS Active signal and the Clock Start. 
+ * @param [in] midi Master Inter-Data Idleness, this bits set the delay between two consecutive data frames in master mode, if SSOM=1, MIDI must be >1.
+ * @param [in] ssiop Slave Select polarity, 1 for active high, 0 for active low.
+ */                   
+void spi_config_ss(uint32_t spi, uint8_t ss_mgt, uint8_t ssoe, uint8_t ssom_mode, uint8_t ssiop, uint8_t mssi, uint8_t midi) {
```
dragonmux, line 60: Per the way it's done everywhere else in the code base, the opening `{` for a function goes on its own line. Same for every other function in this file.

include/libopencm3/stm32/common/spi_common_h7.h
```diff
@@ -0,0 +1,681 @@
...
+#define SPI2				SPI2_BASE
+#define SPI3				SPI3_BASE
+#define SPI4				SPI4_BASE
+#define SPI5				SPI5_BASE
+#define SPI6				SPI6_BASE
+/**@}*/
+
+/*H7 Specific SPI Registers, if it applies to any other family, feel free to copy*/
+/*Control Register 1*/
+#define SPI_CR1(spi_base)   MMIO32((spi_base) + 0x00)
```
dragonmux, line 43: Please always suffix all constants used in pointer arithmetic and mask-and-shift operations with `U` - this makes them unsigned so you don't do a UB at any point. For example, it is an error (by whichever means) to cause the sign bit of a signed integer (the default without this suffix!) to change, thus `(1 << 31)` **is UB**.

We've been trying to solve this and the way it can cause miscompilation that can even time-travel by applying the suffixing - `(1U << 31U)` is completely well defined and does exactly what was intended, compiling correctly and producing correct answers.

Similarly for pointer values, they want to be expressed as `uintptr_t`'s for the same basic reasons, hence suffixing maths used to make pointer values such as this.

dragonmux (changes requested):
Spotted a few different things that are a bit off and need some correction before this can be merged, but otherwise this is looking good and we'd be happy to approve and merge once they're corrected.

include/libopencm3/stm32/h7/rcc.h
fredericotuchtenhag-spec: not sure either why this has changed, i probably typo'd while editing the rcc file, as i was tinkering with it trying to get the i2c to work earlier, and had to insert bit masks for the i2c123 periphs, thanks for noticing

include/libopencm3/stm32/h7/rcc.h
fredericotuchtenhag-spec: sorry, as i've mentioned, it probably conflicted with the other mods i made for the rcc.h file, which corrected some missing masks for the i2c registers

lib/stm32/h7/Makefile
fredericotuchtenhag-spec: sure

PR #1644, opened 2026-05-28
SPI Port for STM32H7xx family

fredericotuchtenhag-spec:
Creating this pull request as i've finished working on the basic functionality of the SPI peripheral on the STM32H7 family of MCUs, here's a briefing of what was already done:

- Complete remap of the associated registers and bit/bitmasks;
- Rewrote setter/getter/helper functions and structs to ease the config processs;
- Tested the write/read function using a NOR Flash IC, working as intended;
- **Chose to create new Header/Lib files for the H7 to avoid confusion, as the SPI registers on this family are drastically different from common_v1 and v2, could've gone with elifs instead but i think that just adds an unnecessary layer of complexity.**


fredericotuchtenhag-spec:
@dragonmux, i think i've covered all the changes you requested, please review when you can!

PR #1648, opened 2026-06-30
Add STM32G0 and STM32G4 meson files

trchen1033:


include/meson.build
```diff
@@ -35,6 +35,8 @@ target_paths = {
 	'stm32f1': 'stm32/f1',
 	'stm32f3': 'stm32/f3',
 	'stm32f4': 'stm32/f4',
+	'stm32g0': 'stm32/g0',
+	'stm32g4': 'stm32/g4',
```
dragonmux, line 40: These should really go after the F7 entry so they're kept in alphabetical order please.

lib/stm32/meson.build
```diff
@@ -52,6 +52,8 @@ subdirs = {
 	'stm32f1': 'f1',
 	'stm32f3': 'f3',
 	'stm32f4': 'f4',
+	'stm32g0': 'g0',
+	'stm32g4': 'g4',
```
dragonmux, line 57: Ditto with this.

lib/stm32/g4/meson.build
```diff
@@ -0,0 +1,105 @@
...
+)
+
+# Sources specific to the G4 series
+libstm32g4_sources = files(
+	'adc.c',
+	'fdcan.c',
+	'flash.c',
+	'pwr.c',
+	'rcc.c',
+	'vector_chipset.c',
```
dragonmux: This file is included in the build via another means and should be dropped from this files list. It gets included through the Cortex-M NVIC code in the lib/cm3 part of the tree.

dragonmux (changes requested):
Couple of review notes, but with them taken care of we'd be more than happy to merge this - it's looking good.

include/meson.build
tienc-beecorp: Done.

lib/stm32/meson.build
tienc-beecorp: Done.

lib/stm32/g4/meson.build
tienc-beecorp: Done.

dragonmux (approved):

PR #1651, opened 2026-07-16
Enable CRCv2 for STM32U5

ALTracer:
* STM32U585 contains a CRC peripheral (v2, with support for bit-reverse, 8/16/32 size, modifiable polynomial)
* Copy the boilerplate header (from STM32L4) to enable existing support in libopencm3 for that

Tested on https://github.com/WeActStudio/WeActStudio.STM32U585Cx_CoreBoard , like PR1612

### DIFF
--- a/include/libopencm3/stm32/crc.h
+++ b/include/libopencm3/stm32/crc.h
@@ -44,6 +44,8 @@
 #       include <libopencm3/stm32/g4/crc.h>
 #elif defined(STM32H7)
 #       include <libopencm3/stm32/h7/crc.h>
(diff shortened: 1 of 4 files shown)


dragonmux (changes requested):
Could you please include the CRC v2 peripheral's sources in the Makefile and meson.build so that if someone tries to use functions from that header it doesn't crash and burn? Otherwise this looks great. This should be a simple one-line change for both.

dragonmux (approved):
LGTM, merging. Thank you for the contribution and fixing this oversight!

PR #1652, opened 2026-07-19
Add ICACHE support

ALTracer:
* All modern STM32 SoC based on Cortex-M33 contain an ICACHE peripheral (Instruction Cache), disabled by default. To improve performance at high ratios of System-to-Flash clock, ICACHE should be enabled.
* This PR provides basic functions to enable, disable, and invalidate instruction cache, as well as macros for registers and bits.
* Currently enabled for U5 only, but written in such a way that other family chips (L5, H5, C5, U3) can leverage this support easily when added to libopencm3. Enabled in Makefiles and Meson.

DCACHE support is out of scope, because internal SRAM caching does not improve latency, and only bigger U5 & H5 (not H503) have FMC, OCTOSPI external memory where data load/store could be cached.

### DIFF
--- a/lib/stm32/u5/Makefile
+++ b/lib/stm32/u5/Makefile
@@ -42,6 +42,7 @@ OBJS += desig_common_v1.o
 OBJS += exti_common_all.o
 OBJS += flash_common_all.o flash.o
 OBJS += gpio_common_all.o gpio_common_f0234.o
+OBJS += icache_common_all.o
 OBJS += iwdg_common_all.o
 OBJS += i2c_common_v2.o
 OBJS += pwr.o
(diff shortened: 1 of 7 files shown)


lib/stm32/u5/Makefile
```diff
@@ -42,6 +42,7 @@ OBJS += desig_common_v1.o
 OBJS += exti_common_all.o
 OBJS += flash_common_all.o flash.o
 OBJS += gpio_common_all.o gpio_common_f0234.o
+OBJS += icache.o
```
dragonmux: Surely this would be `+= icache_common_all.o`? - this otherwise refers to a non-existent .c file

dragonmux (changes requested):
Only have the one note, which once addressed we're happy for this to be merged.

lib/stm32/u5/Makefile
ALTracer: Good catch, thanks.

dragonmux (approved):

